PR #25, opened 2018-11-21
fix file dialog and crash when mounting

ThomasWaldmann:


src/vorta/utils.py
```diff
@@ -59,6 +59,7 @@ def choose_folder_dialog(parent, title):
     options = QFileDialog.Options()
     options |= QFileDialog.ShowDirsOnly
     dialog = QFileDialog(parent, title, os.path.expanduser('~'), options=options)
+    dialog.setFileMode(QFileDialog.Directory)
```
ThomasWaldmann: borg accepts files AND dirs as source specification, so it would be best to be able to select either of them. but i could not find out how in the pyqt docs.

when having to decide, go for "directory".

PR #42, opened 2018-11-22
add flake8 checker, fix trivial issues

ThomasWaldmann:
@m3nu see the flake8 ignores in `setup.cfg`, they point to some problematic places in the code.


m3nu:
Awesome. Will look into the ignores.

PR #61, opened 2018-11-27
Extract cmd

### LINKED ISSUE #21: "Refreshing snapshots" run's forever for very large repos
I have... a LOT of snapshots (511 of them).

After adding my existing repository Vorta quickly shows size-information (Original Size: 44TB; Deduplicated Size: 162GB; Compressed Size: 88GB).

Also, the progress bar changes into the "flowing" orange and the text "Refreshing snapshots.." is shown.

It stays there. I just hit "cancel" after significantly more than one day. I have no idea if it actually did anything. The disks in my NAS weren't active and I'm seeing just about zero network traffic.

I have no idea what's happening or how to show what's currently being done.

m3nu:
- fixes #21 by reading stdout and stderr simultaneously.
- Adds a new dialog to read remote archives and extract selected files and folders.
screenshot 2018-11-27 at 09 35 23

Still needs some cleanup and tests. But will merge it fairly soon because the feature is isolated and doesn't touch much else.


src/vorta/borg/borg_thread.py
```diff
@@ -135,40 +139,74 @@ def run(self):
...
+        p = Popen(self.cmd, stdout=PIPE, stderr=PIPE, bufsize=1, universal_newlines=True,
+                  env=self.env, cwd=self.cwd, preexec_fn=os.setsid)
+        self.process = p
 
-        for line in iter(self.process.stderr.readline, ''):
+        # Prevent blocking. via https://stackoverflow.com/a/7730201/3983708
+
+        # Helper function to add the O_NONBLOCK flag to a file descriptor
+        def make_async(fd):
+            fcntl.fcntl(fd, fcntl.F_SETFL, fcntl.fcntl(fd, fcntl.F_GETFL) | os.O_NONBLOCK)
```
ThomasWaldmann: https://docs.python.org/3/library/os.html#os.set_blocking

m3nu: So this would be a shortcut instead of using this function?

ThomasWaldmann: yup.

m3nu: Good tip. This made it a lot [simpler](https://github.com/borgbase/vorta/pull/63/files#diff-74bac960beabf99e83633ac112450502). Thanks.

PR #86, opened 2018-12-09
v0.6.1 Release

m3nu:


src/vorta/views/repo_add_dialog.py
```diff
@@ -87,8 +93,8 @@ def init_ssh_key(self):
 
     def validate(self):
         """Pre-flight check for valid input and borg binary."""
-        if len(self.values['repo_url']) < 5:
-            self._set_status('Please enter a valid repo URL or path.')
+        if self.is_remote_repo and not re.match(r'.+\..+:.+', self.values['repo_url']):
```
ThomasWaldmann: user@host:/path does not match this.

m3nu: The idea was to only match FQDNs. Like `user@host.local:/path`. or `host.com:repo`

If someone is use e.g. `user@nas:/path`, then just `.+:.+` will be better.

m3nu: Changed to `.+:.+`.

PR #114, opened 2019-01-09
Add Background notifications for Linux

### LINKED ISSUE #84: Notifications for Linux desktop
Hi,

i really like vorta! works great. 
It's possible to add a features for a mail report for each backup? 

Thanks

bpereto:
* [X]  Add `notify2` for Linux notifications
* [X]  Have option to globally enable/disable notifications. "Display notifications when background tasks fail."
* [X]  Additional sub-option to notify about successful backups as well: "Also notify about successful background tasks."
* [X] Tests
* [X] requirements for notify2

fixes #84

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ addons:
     - libacl1-dev
     - libacl1
     - build-essential
-    - python3-dbus
+    - libdbus-1-dev
+    - libdbus-glib-1-dev
     - dbus
   homebrew:
     update: false
@@ -58,6 +59,7 @@ install:
     pyenv install -s $PYTHON36
     eval "$(pyenv init -)"
     pyenv shell $PYTHON36
+    pip install dbus-python
   elif [ $TRAVIS_OS_NAME = "osx" ]; then
     brew upgrade pyenv
     pyenv install -s $PYTHON37
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ addons:
     - libacl1-dev
     - libacl1
     - build-essential
+    - python3-dbus
     - dbus
   homebrew:
     update: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,6 @@ install:
     pyenv install -s $PYTHON36
     eval "$(pyenv init -)"
     pyenv shell $PYTHON36
-    pip install dbus-python
   elif [ $TRAVIS_OS_NAME = "osx" ]; then
     brew upgrade pyenv
     pyenv install -s $PYTHON37
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ addons:
     - libacl1-dev
     - libacl1
     - build-essential
-    - libdbus-1-dev
-    - libdbus-glib-1-dev
-    - dbus
   homebrew:
     update: false
     packages:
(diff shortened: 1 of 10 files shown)


bpereto:
thanks, seems to be a problem with dbus on travis..
notify2 relies on nativ installed dbus python bindings that are not available in the tox environments.

m3nu:
I already install a window manager on Travis to do the GUI tests. I can probably add a desktop with dbus as well. Let me look into it later.

m3nu:
Made some small adjustments in #115. Adding inline-comments to explain details.

src/vorta/notifications.py
```diff
@@ -20,25 +25,44 @@ def pick(cls):
...
-            from Foundation import NSUserNotificationCenter
 
-            notification = NSUserNotification.alloc().init()
-            notification.setTitle_(title)
-            notification.setInformativeText_(text)
-            center = NSUserNotificationCenter.defaultUserNotificationCenter()
-            if center is not None:  # Only works when run from app bundle.
-                center.deliverNotification_(notification)
+    def deliver(self, title, text, level='info'):
+        if not SettingsModel.get(key='enable_notifications').value:
```
m3nu: Some duplicate code here that decides on whether to send a notification or not. Moved this to a single function.

src/vorta/notifications.py
```diff
@@ -1,6 +1,11 @@
 import sys
 from vorta.models import SettingsModel
 
+if sys.platform == 'darwin':
```
m3nu: If dbus isn't available, the import fails. Since Linux is very heterogeneous, we have to assume it fails.

So I moved those imports to a try-except block.

m3nu:
Still looking into the testing stuff on Travis.

To add my commits to your PR:

```
git remote add m3nu https://github.com/m3nu/vorta
git pull m3nu bpereto-add-notifications
```

.travis.yml
```diff
@@ -10,6 +10,8 @@ addons:
     - libacl1-dev
     - libacl1
     - build-essential
+    - python3-dbus
```
m3nu: Good intention, but the test happens in pyenv-supplied Python, not the system Python (which is too old). I'm trying to install dbus there. If it doesn't work, we'll just mock it.

m3nu: Looks like Travis' dbus is too old anyways.

`Requested 'dbus-1 >= 1.8' but version of dbus is 1.6.18`

m3nu:
While sleeping I thought of 2 more things:

- When choosing the notification class, DBus isn't just for Linux. [BSDs](https://www.freshports.org/devel/dbus) use it as well. So it could be renamed to `DBusNotifications` and used whenever the bus-connection [here](https://github.com/borgbase/vorta/pull/114/files#diff-5d199731fc87f26ed8a194d36ea4ca17R18) succeeds.
- Use logging instead of pritnt statements [here](https://github.com/borgbase/vorta/pull/114/files#diff-5d199731fc87f26ed8a194d36ea4ca17R92) (3x). Would be

```
from vorta.log import logger
logger.warning(...)
```

bpereto:
> Use logging instead of print statements here (3x). Would be
common use of logger everywhere in vorta would great

> DBusNotifications
good idea

m3nu:
> common use of logger everywhere in vorta would great

There is a common logger – `vorta.log` – which logs to a file in a default location. It's already used in many places. Mostly to pass on Borg logs. We should use it for DBus errors as well.

PR #115, opened 2019-01-10
Bpereto add notifications

m3nu:
Small changes to #114:

- Assume no `dbus` is available on Linux and fall back to dummy notifier.
- Move duplicate code to decide on notification to parent class.


bpereto:
no problem :). glad it works now.
seems to work on my machine - so this PR seems done to me.

I updated my PR with your fixes.

PR #128, opened 2019-01-16
Fix umount

antlarr:
unmounting an archive didn't work for me because a directory like '/foo/' is compared to the items in the list of mounted directories, which contained '/foo' . This PR fixes this problem by normalizing the paths before comparison.

Also, use the wants_folder parameter when asking the user for a mount point or an extraction point


antlarr:
These changes are also included in #129 so I'm closing this.

PR #129, opened 2019-01-17
 Improvements to the archive tab

antlarr:
* Fix for unmounting archives by normalizing the paths

Unmounting an archive didn't work for me because a directory like '/foo/' is compared to the items in the list of mounted directories, which contained '/foo' . This PR fixes this problem by normalizing the paths before comparison.

* Use the wants_folder parameter when asking the user for a mount point or an extraction point

* Improvements to the mount/unmount feature

Make it possible to mount more than one archive at the same time. When the user selects an archive, the mount button is updated and changes between Mount/Unmount actions depending on the state of the selected archive.

* Use a mount icon for mounted archives and check premounted archives  …
Add a column to show an icon in mounted archives.

* Use psutil to check on startup (and refreshes) the mounted archives,
so Vorta doesn't get out-of-sync with what's really mounted on the system.

* Add a context menu to the archive table (and open folder action)

Add a context menu to the archive table and a way to quickly open a file browser on the archive's mount point.

* Also open the mount point folder when double clicking on the mounted icon

### DIFF
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -10,6 +10,7 @@ import getpass
 from collections import defaultdict
 from functools import reduce
 import operator
+import psutil
 
 from paramiko.rsakey import RSAKey
 from paramiko.ecdsakey import ECDSAKey
@@ -247,3 +248,32 @@ def format_archive_name(profile, archive_name_tpl):
         'user': getpass.getuser()
     }
     return archive_name_tpl.format(**available_vars)
+
+
+def open_folder(folder):
+    """Open a file browser on a folder."""
+    cmd = 'open' if sys.platform == 'darwin' else 'xdg-open'
+    subprocess.Popen([cmd, folder])
+
+
+def get_mount_points(repo_url):
+    mount_points = {}
+    for proc in psutil.process_iter():
+        if proc.name() != 'borg':
+            continue
+
+        if 'mount' not in proc.cmdline():
+            continue
+
+        for idx, parameter in enumerate(proc.cmdline()):
+            if parameter.startswith(repo_url + '::'):
+                archive_name = parameter[len(repo_url) + 2:]
+
+                # The borg mount command specifies that the mount_point
+                # parameter comes after the archive name
+                if len(proc.cmdline()) > idx + 1:
+                    mount_point = proc.cmdline()[idx + 1]
+                    mount_points[archive_name] = mount_point
+                break
+
+    return mount_points
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -259,21 +259,19 @@ def open_folder(folder):
 def get_mount_points(repo_url):
     mount_points = {}
     for proc in psutil.process_iter():
-        if proc.name() != 'borg':
-            continue
+        if proc.name() == 'borg' or proc.name().startswith('python'):
+            if 'mount' not in proc.cmdline():
+                continue
 
-        if 'mount' not in proc.cmdline():
-            continue
+            for idx, parameter in enumerate(proc.cmdline()):
+                if parameter.startswith(repo_url + '::'):
+                    archive_name = parameter[len(repo_url) + 2:]
 
-        for idx, parameter in enumerate(proc.cmdline()):
-            if parameter.startswith(repo_url + '::'):
-                archive_name = parameter[len(repo_url) + 2:]
-
-                # The borg mount command specifies that the mount_point
-                # parameter comes after the archive name
-                if len(proc.cmdline()) > idx + 1:
-                    mount_point = proc.cmdline()[idx + 1]
-                    mount_points[archive_name] = mount_point
-                break
+                    # The borg mount command specifies that the mount_point
+                    # parameter comes after the archive name
+                    if len(proc.cmdline()) > idx + 1:
+                        mount_point = proc.cmdline()[idx + 1]
+                        mount_points[archive_name] = mount_point
+                    break
 
     return mount_points
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -84,11 +84,12 @@ def get_private_keys():
     available_private_keys = []
     if os.path.isdir(ssh_folder):
         for key in os.listdir(ssh_folder):
+            key_file = os.path.join(ssh_folder, key)
+            if not os.path.isfile(key_file):
+                continue
             for key_format in key_formats:
                 try:
-                    parsed_key = key_format.from_private_key_file(
-                        os.path.join(ssh_folder, key)
-                    )
+                    parsed_key = key_format.from_private_key_file(key_file)
                     key_details = {
                         'filename': key,
                         'format': parsed_key.get_name(),
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -32,15 +32,13 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         self.toolBox.setCurrentIndex(0)
 
         self.folder_icon = QIcon(':/icons/folder-open.svg')
-        self.archiveTable.setHorizontalHeaderItem(3, QTableWidgetItem(self.folder_icon, ''))
         header = self.archiveTable.horizontalHeader()
         header.setVisible(True)
         header.setSectionResizeMode(0, QHeaderView.ResizeToContents)
         header.setSectionResizeMode(1, QHeaderView.ResizeToContents)
         header.setSectionResizeMode(2, QHeaderView.ResizeToContents)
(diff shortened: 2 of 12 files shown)


antlarr:
Also, I'd like to ask about some plans I have:

* Add some kind of mounted icon to the archive table when an archive is mounted, so the user can quickly see which are mounted.
* Since the dictionary of mount points per archive is stored in-process, it's lost when the user restarts Vorta. Which means that upon restart, all archives appear as unmounted and the user has to resort to the command line to unmount them, which is not nice. I would like to use psutil to parse the running commands on the system, search for "borg mount" commands and extract the archive names/mount points from there so the information used by Vorta is always correct (I would love to see any other way to get the archive associated with a mount point, but I'm afraid I couldn't find any other way to get that information but to parse command lines).

Is there interest in those changes?

m3nu:
> Is there interest in those changes?

Yes. Definitely interested. This was neglected so far. But there is some [existing code to check for mount points](https://github.com/borgbase/vorta/blob/master/src/vorta/borg/umount.py). It's currently used to validate the mount point before unmounting it.

I've opened an issue for this feature at #131.

m3nu:
How about merging PRs #128 and #129, since they are related? Just pulling the code. Excited to test it.

m3nu:
Sounds good.

src/vorta/utils.py
```diff
@@ -246,3 +246,13 @@ def format_archive_name(profile, archive_name_tpl):
...
     }
     return archive_name_tpl.format(**available_vars)
+
+
+def open_folder(folder):
+    """Open a file browser on a folder."""
+    if sys.platform == 'darwin':
+        cmd = 'open'
+    else:
+        cmd = 'xdg-open'
```
ThomasWaldmann: cmd = 'open' if sys.platform == 'darwin' else 'xdg-open'

src/vorta/views/archive_tab.py
```diff
@@ -181,64 +194,74 @@ def mount_action(self):
...
+            self.update_mount_button_text()
 
     def umount_action(self):
-        if self.mount_point is not None:
+        snapshot_name = self.selected_archive_name()
+
+        try:
+            mount_point = self.mount_points[snapshot_name]
+        except KeyError:
+            mount_point = None
```
ThomasWaldmann: mount_point = self.mount_points.get(snapshot_name)

src/vorta/views/archive_tab.py
```diff
@@ -297,8 +320,71 @@ def receive():
...
+
+    def update_mount_button_text(self):
+        snapshot_name = self.selected_archive_name()
+        if not snapshot_name:
+            return
+
+        if snapshot_name in self.mount_points:
+            self.set_mount_button_mode('Unmount')
+        else:
+            self.set_mount_button_mode('Mount')
```
ThomasWaldmann: mode = 'Unmount' if snapshot_name in self.mount_points else 'Mount'
self.set_mount_button_mode(mode)

src/vorta/views/archive_tab.py
```diff
@@ -297,8 +320,71 @@ def receive():
...
+
+    def open_folder_action(self):
+        snapshot_name = self.selected_archive_name()
+        if not snapshot_name:
+            return
+
+        try:
+            mount_point = self.mount_points[snapshot_name]
+        except KeyError:
+            mount_point = None
```
ThomasWaldmann: .get(), see above.

src/vorta/views/archive_tab.py
```diff
@@ -297,8 +320,71 @@ def receive():
...
+
+        if mount_point:
+            open_folder(mount_point)
```
ThomasWaldmann: above you used "is not None", maybe always use the same method.

if the empty string or any other falsy value except None can not happen, the shorter one is nicer.

src/vorta/utils.py
antlarr: done

src/vorta/views/archive_tab.py
antlarr: sure

src/vorta/views/archive_tab.py
antlarr: ok

src/vorta/views/archive_tab.py
antlarr: ok

src/vorta/views/archive_tab.py
antlarr: I tend to use the implicit check in my code but I noticed Vorta sources tend to prefer the "is not None" kind of check, so I tried to use it everywhere to keep the code style but this one seem to have slipped. I'll fix it. Thanks for the review!

src/vorta/views/archive_tab.py
```diff
@@ -181,11 +196,20 @@ def list_result(self, result):
     def selected_archive_name(self):
         row_selected = self.archiveTable.selectionModel().selectedRows()
         if row_selected:
-            snapshot_cell = self.archiveTable.item(row_selected[0].row(), 3)
+            snapshot_cell = self.archiveTable.item(row_selected[0].row(), 4)
```
ThomasWaldmann: iirc @m3nu just removed the word "snapshot" from the code in the i18n PR and replaced it with "archive".

m3nu: There will be some conflicts with #134. If we merge #134 before #129, it should be rebased when ready.

m3nu:
Resolved some conflicts with #134. Are you still planning changes, @antlarr ?

m3nu:
One test and flake8 failing.

src/vorta/utils.py
```diff
@@ -252,3 +253,22 @@ def open_folder(folder):
     """Open a file browser on a folder."""
     cmd = 'open' if sys.platform == 'darwin' else 'xdg-open'
     subprocess.Popen([cmd, folder])
+
+
+def get_mount_points(repo_url):
+    processes = [p for p in psutil.process_iter() if 'borg' in p.name() and 'mount' in p.cmdline()]
+
+    mount_points = {}
+    for proc in psutil.process_iter():
```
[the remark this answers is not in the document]
antlarr: Sorry, that was code from an early stage of development. I just removed it

antlarr:
I'll fix those comments by Thomas and the test, then I'm done. I'm just not sure when I'll have time (probably tomorrow, but I'll try to finish everything today). 

Btw, what flake8 issue do you mean?

src/vorta/views/archive_tab.py
antlarr: I renamed the existing snapshot_* variables to archive_*

src/vorta/views/archive_tab.py
```diff
@@ -366,17 +389,32 @@ def archiveTable_context_menu_event(self, event):
...
+        if column == 3:
+            self.open_folder_action()
+
+    def row_of_archive(self, archive_name):
+        items = self.archiveTable.findItems(archive_name, QtCore.Qt.MatchExactly)
+        rows = [item.row() for item in items if item.column() == 4]
+        if rows == []:
+            return None
+
+        return rows[0]
```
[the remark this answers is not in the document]
antlarr: ok

src/vorta/views/archive_tab.py
```diff
@@ -181,11 +196,20 @@ def list_result(self, result):
...
 
+    def set_mount_button_mode(self, mode):
+        self.mountButton.clicked.disconnect()
+        if mode == 'Mount':
+            self.mountButton.setText('Mount')
+            self.mountButton.clicked.connect(self.mount_action)
+        elif mode == 'Unmount':
+            self.mountButton.setText('Unmount')
+            self.mountButton.clicked.connect(self.umount_action)
+
```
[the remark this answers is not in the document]
antlarr: done

src/vorta/utils.py
```diff
@@ -252,3 +253,22 @@ def open_folder(folder):
...
+        if proc.name() != 'borg':
+            continue
+
+        if 'mount' not in proc.cmdline():
+            continue
+
+        if proc.cmdline()[-2].startswith(repo_url):
+            archive_name = proc.cmdline()[-2][len(repo_url) + 2:]
+            mount_point = proc.cmdline()[-1]
+            mount_points[archive_name] = mount_point
```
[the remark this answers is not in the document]
antlarr: I changed a bit the algorithm so the archive_name position in the command line is not hardcoded anymore

m3nu:
Just tested this. Good stuff really. Made 2 small changes:

- detecting mount points via the `borg` binary only works when using the fat binary. The process name could also be `pythonxxx`.
- I put the truncated mount point in the table. Why not show it, if it's known?

Now I just need to figure out how to merge this, since there are some extra commits from yesterday in here.

m3nu:
Moved a squashed version to #156 . Easier to see the changes there.

PR #130, opened 2019-01-17
Make the main window resizeable

antlarr:
Remove the maximum size restriction and assign a layout to the main window so it resizes gracefully (but keep the default size to the current one, just don't forbid the user from resizing the window if he/she wants to do it).


m3nu:
Thanks. Lets try it. With the window resizable more care needs to be taken to gracefully resize all the parent elements.

PR #134, opened 2019-01-18
i18n infrastructure and policy

### LINKED ISSUE #99: i18n: translation, localisation (overview)
recently showed borg at the local linux user group, it was well received, some people prefer gui to cli tools, so vorta is just what they waited for.

but, first question of one of the guys: "gibt's das auch auf Deutsch?" ("is this also available in German?") and i had to tell "no, not yet".

so, while i18n for borg is rather optional, because UNIX admins hacking stuff on the cli usually also speak English rather well, it's a different story for the target audience of vorta - we need i18n for them.

so, what needs to be done:
- (this ticket) code needs to prepared for i18n, qt style:

  - `vorta.i18n` package with some i18n infrastructure code
  - `self.tr('english phrase')` (scope comes from instance class name)
  - `app.translate('scope', 'english phrase')`
- #135 packaging
- #136 check original texts
- #137 infrastructure for translators, integration
- #138 i18n policy, see also #140
- #139 layout short vs. long texts / rtl languages
- #141 localisation
- UI needs to be as self-explaining as possible (because I guess we do not want to translate the docs also).
- I'ld advise to only add 1 language at first, get everything implemented / tested for that and only after that add more languages.

### LINKED ISSUE #138: i18n policy
there should be a quality policy for the translations, like e.g.:

- no google translate or other automated translation [of course]
- only human native or as-good-as-native speakers, should be users of vorta (one-time translations are not that helpful if there is noone updating them regularly as needed)
- must have >90% translated strings or language will be made unavailable in next release (this will help to motivate past or future translators). ideally there is also a notifcation service to send notifies to translators if their translation needs more work.

ThomasWaldmann:
fixes #99 
fixes #138


ThomasWaldmann:
Just tweeted to get some help with translating all the strings.

If somebody wants to help, push commits with comment "german translation" to my branch, the interesting file is `src/vorta/i18m/ts/vorta.de_DE.ts` (please do not change other files).

commit/push frequently to avoid conflicts.

src/vorta/scheduler.py
```diff
@@ -114,7 +114,7 @@ def create_backup(self, profile_id):
                 logger.error('Error during backup creation.')
         else:
             logger.error('Conditions for backup not met. Aborting.')
-            logger.error(msg['message'])
+            logger.error(msg['message'])  # XXX: in the log, we usually use english, but this might be translated.
```
ThomasWaldmann: shall we keep this in english, so we can use it for logging?

maybe we could use `translate('message', 'some message') to get it into some specific scope, no matter where the translate calls are.

at one place, there could be a dummy call, just to make it extractable.
then we could hopefully do a late call, like `translate(m)`.

ThomasWaldmann: decided to do it like that and use a trans_late function, to translate late.

ThomasWaldmann:
ok, anything missing (besides the missing de_DE strings)?

m3nu:
Maybe we should use a branch for that instead a PR? So I can make changes too.

ThomasWaldmann:
"Add more commits by pushing to the i18n branch on ThomasWaldmann/vorta."

requirements.d/conda.yml
```diff
@@ -18,6 +18,4 @@ dependencies:
   - pyobjc-core
   - pyobjc-framework-Cocoa
   - pyinstaller
-  - pip:
-      - sentry-sdk
```
ThomasWaldmann: hmm, belongs into your other PR?

Makefile
```diff
@@ -21,16 +24,32 @@ pypi-release:
...
 	twine upload dist/vorta-0.6.5.tar.gz
 
-bump-version:
+bump-version:  ## Add new version tag and push to upstream repo.
 	bumpversion patch
 	#bumpversion minor
 	git push upstream
 
-travis-debug:
+travis-debug:  ## Connect to Travis instance using SSH.
```
ThomasWaldmann: Prepare connecting ...

ThomasWaldmann:
the transifex api token is private, right? then @m3nu needs to test the to/from transifex related tooling and docs. :)

ThomasWaldmann:
@m3nu how will new/changed strings from .py/.ui get into existing translation (like "de")?

`translations-from-source` only updates the `en_US` .ts file.

ThomasWaldmann:
I'm translating now on transifex (german).

ThomasWaldmann:
I updated the Makefile a bit because the tx pull/push stuff got in my way, asking for API key...
I didn't want to invoke tx anyway, because I downloaded the .ts file with my browser, I just wanted the "compile qm" part.

ThomasWaldmann:
hmm, should transifex automatically pull new stuff from github? if so, it does not work. :)

i updated the en_US.ts with new strings, but transifex only shows the old strings for german.

ThomasWaldmann:
btw, for everbody using pycharm:
- Settings -> "File Types"
- remove "*.ts" from TypeScript (gets rid of lots of wrong alarms)
- add "*.ts" to XML (there is no special entry for Qt translation files)

m3nu:
> hmm, should transifex automatically pull new stuff from github? if so, it does not work. :)

It doesn't do that. You need to `tx push` with the Makefile `make translations-from-source` command. `pylupdate5 ...` (gets new strings from source files) and then `tx push -s` (pushes new translations).

ThomasWaldmann:
I'ld do some cleanups and a force push first, if you let me. :)

m3nu:
> I'ld do some cleanups and a force push first, if you let me. :)

Please.

ThomasWaldmann:
OK, guess I am finished with it.

It is all still there, just did merge the lots of fixup commits and killed all the .ts files.

After this is merged, please pull all the ts files from transifex and commit them in a separate PR.

I tested it with system locale (en) and also with de and my old qm file, worked.

PR #143, opened 2019-01-19
Show/Hide the main window when clicking on the tray icon

antlarr:
This makes it easier to open the Vorta main window just by clicking
on the tray icon (instead of having to right-click and then click on
Settings). Also, vorta now has the common behaviour of closing the
main window if it's already visible when the user clicks on the tray
icon.

### DIFF
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -82,6 +82,12 @@ class VortaApp(QApplication):
         self.main_window.show()
         self.main_window.raise_()
 
+    def toggle_main_window_visibility(self):
+        if self.main_window.isVisible():
+            self.main_window.hide()
+        else:
+            self.open_main_window_action()
+
     def backup_started_event_response(self):
         set_tray_icon(self.tray, active=True)
 
--- a/src/vorta/tray_menu.py
+++ b/src/vorta/tray_menu.py
@@ -1,4 +1,4 @@
-import sys
+import os
 from PyQt5.QtWidgets import QMenu, QSystemTrayIcon
 
 from .borg.borg_thread import BorgThread
@@ -18,14 +18,16 @@ class TrayMenu(QSystemTrayIcon):
 
         self.setContextMenu(menu)
 
-        if sys.platform != 'darwin':
-            self.activated.connect(self.on_activation)
+        self.activated.connect(self.on_activation)
         self.setVisible(True)
         self.show()
 
     def on_activation(self, reason):
         if reason == QSystemTrayIcon.Trigger:
-            self.app.toggle_main_window_visibility()
+            if os.environ.get('XDG_CURRENT_DESKTOP', '') == 'KDE':
+                self.app.toggle_main_window_visibility()
+            else:
+                self.on_user_click()
 
     def on_user_click(self):
         """Build system tray menu based on current state."""
--- a/src/vorta/tray_menu.py
+++ b/src/vorta/tray_menu.py
@@ -1,3 +1,4 @@
+import sys
 from PyQt5.QtWidgets import QMenu, QSystemTrayIcon
 
 from .borg.borg_thread import BorgThread
@@ -17,7 +18,8 @@ class TrayMenu(QSystemTrayIcon):
 
         self.setContextMenu(menu)
 
-        self.activated.connect(self.on_activation)
+        if sys.platform != 'darwin':
+            self.activated.connect(self.on_activation)
         self.setVisible(True)
         self.show()
 
--- a/src/vorta/tray_menu.py
+++ b/src/vorta/tray_menu.py
@@ -16,8 +16,14 @@ class TrayMenu(QSystemTrayIcon):
         menu.aboutToShow.connect(self.on_user_click)
 
         self.setContextMenu(menu)
+
+        self.activated.connect(self.on_activation)
         self.setVisible(True)
         self.show()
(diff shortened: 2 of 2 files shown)


m3nu:
What if I just want to run a backup using the tray icon without opening the window?

antlarr:
> What if I just want to run a backup using the tray icon without opening the window?

Well, everything else works as before. A right click still opens the context menu. The commit just adds a new behaviour. When left clicking on the icon (which previously did nothing) now opens the main window, which is standard behaviour among most tray icon applications.

m3nu:
Aahh. MacOS only has left click (mostly). If it's common behavior on Linux, it's fine to do it like that.

antlarr:
Ah, maybe I should do that only when platform is not darwin?

m3nu:
> A right click still opens the context menu.

I'm not seeing any code that's specific to right or left click. When I run it the window visibility toggles every time I click the icon. That's a bit weird.

m3nu:
I would add it, if the main use of Vorta would be to show the Settings window. Similar as a network- or battery monitor. But in this case, the settings window is only meant to be used to change settings now and then. Most of the time it should just be in the background. If you need this feature to look after pending operations, maybe the current status display in the menu can be extended.

That said, the feature is probably too exotic and doesn't add a lot of value. We are also trying to keep the `== darwin/linux` stuff to an absolute minimum.

I could be wrong of course. And I'm not a Linux desktop user. It would be good if some Linux desktop users could give their opinion as well.

antlarr:
Well, on Linux (at least on KDE) it's quite common that a left click on a system tray icon opens/close the main window (I just checked pidgin, kleopatra, kopete, qbittorrent... all those behave like that) and a right click opens a context menu. I searched if it's documented anywhere and it doesn't seem to be, though, it just seems to be a de-facto standard. The only thing I could find about that is [this guideline](https://www.redhat.com/archives/xdg-list/2003-March/msg00060.html) posted in 2003 in the xdg-list and open-hci mailing lists at freedesktop.org (which is the closest we have in Linux to a standards organization handling desktop behaviour). There it's said:

> Suggested:
> [...]
>	+ Single-clicking on the icon or pressing Enter/Space when the
>	  icon has focus (this of course requires that the icons are
>	  focusable) should present a dialog with further information
>	  on the status currently displayed. The dialog may optionally
>	  also the user to control the status.
> [...]
>	+ Right click or Shift-F10 should display a context
>      [...]

ThomasWaldmann:
I just did a left click on all tray icons I have:
- most just open the icon's tray menu
- vorta does nothing (now), would open a windows (after this changeset)

So, can we just open the menu?

m3nu:
+1 for just opening the menu

m3nu:
Ubuntu via VMWare: both left and right click open the tray menu.

ThomasWaldmann:
@antlarr using qubes os, has xfce by default.

antlarr:
Ok, in that case this seems to be a KDE behaviour mostly. I see two options:

1) When the user clicks on the tray icon, check the XDG_CURRENT_DESKTOP environment variable, if it's KDE, open the main window to keep a coherent desktop behaviour, if the environment variable is not set or has some other value, open the context menu.

2) When the user clicks on the tray icon, just open the context menu.

I'd prefer option 1. Opinions?

Hofer-Julian:
Nextcloud opens the context menu with one click on the tray icon, and the main window with a double click.

I find this to be very usable and it doesn't involve a right click as well.

How do you feel about this third option?

antlarr:
I just asked the KDE Visual Design Group about this and they said currently there's nothing defined (but they agree it should be) and also added _"the most common pattern is probably … - left click: open the panel or app … - right click: open a context menu … - double click: not used"_

About double-click opening the main window... I also think that's uncommon (at least, in KDE) and most users won't know about it, so I don't see much benefit in option 3 when compared with option 2.

About user stats, I think this can give some insight: https://www.linuxjournal.com/content/best-desktop-environment .

I would prioritize integration into the user desktop (so the application behaves like the other applications the user has) thus contributing to a more common user experience for users of all desktops/OS, even if that means using some platform-specific code. But you, @m3nu, have the final word to decide between options 1, 2 and 3.

antlarr:
Just in case you're interested, the  KDE Visual Design Group just started working [here](https://invent.kde.org/websites/hig-kde-org/merge_requests/8) on adding the behaviour of system tray icons to their [HIG](https://hig.kde.org/)

antlarr:
> * KDE has a common pattern, but no strong opinion on it, as mentioned.
> 
> So this change would only apply to KDE, which has a "common pattern" but no real rule or recommendation.

Actually, as I mentioned before, the KDE Visual Design Group started working on that a few days ago and the common pattern is already the official recommendation, as can be seen at [https://hig.kde.org/plattform/tray.html#behavior](https://hig.kde.org/plattform/tray.html#behavior).

m3nu:
Not what I meant, but fine.

The PR still needs a rebase though.

m3nu:
Rebased and merged.

For the future, let's avoid sparkling those conditions all over. Platform-specific functionality should be in one module with a platform-specific implementation for each. Like `vorta.notifications` or `vorta.updater`.

antlarr:
ok, I'll try to keep that in mind. Thanks for accepting this!

PR #146, opened 2019-01-20
Translation-related improvements and German translation

m3nu:


.gitignore
```diff
@@ -6,6 +6,7 @@ docs/
 *.autosave
 bin/macosx64/Sparkle.framework
 bin/macosx64/borg
```
ThomasWaldmann: oops. i recently killed that subdir because the vorta.spec works more easily in `bin/` (not `bin/macosx64`).
is it required by sparkle?

src/vorta/i18n/__init__.py
```diff
@@ -13,7 +13,7 @@ def init_translations(app):
     global application, translator, locale  # if we don't keep a reference on these, it stops working. pyqt bug?
     application = app
     translator = QTranslator()
-    locale = QLocale()
+    locale = QLocale('de')
```
ThomasWaldmann: used for debugging and forget to remove again?

ThomasWaldmann: i used this to debug:
```
LANG=de_DE vorta --foreground
```

.gitignore
m3nu: Not required. I just keep it there. This entry is redundant anyways due to the `bin/*` below.

src/vorta/i18n/__init__.py
m3nu: Yepp. Will use the env-var from now.

m3nu:
Left to do:

- [ ] review EN strings
- [ ] ..?

.gitignore
```diff
@@ -4,8 +4,7 @@ build/
 dist/
 docs/
 *.autosave
-bin/macosx64/Sparkle.framework
-bin/macosx64/borg
+bin/*
```
ThomasWaldmann: assuming that the dist/ and docs/ work, no need for the `*`?

ThomasWaldmann: is docs/ something you are still working on or why is it here?

ThomasWaldmann:
/me would push "review EN strings" to next PR.

.gitignore
m3nu: That's true. And they seem to work. 🎉

PR #147, opened 2019-01-20
get_private_keys: do not try to read sockets

sebcode:
I store my ControlMaster sockets in `~/.ssh/` due to `ControlPath ~/.ssh/%r@%h:%p` in my ssh config. While trying to read private key files, Vorta stumbled upon these sockets and crashed on startup:

```
  File "/Users/seb/.pyenv/versions/3.6.8/lib/python3.6/site-packages/paramiko/pkey.py", line 278, in _read_private_key_file
    with open(filename, "r") as f:
OSError: [Errno 102] Operation not supported on socket: '/Users/seb/.ssh/seb@asdf'
```

### DIFF
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -84,13 +84,12 @@ def get_private_keys():
     available_private_keys = []
     if os.path.isdir(ssh_folder):
         for key in os.listdir(ssh_folder):
-            if not os.path.isfile(os.path.join(ssh_folder, key)):
+            key_file = os.path.join(ssh_folder, key)
+            if not os.path.isfile(key_file):
                 continue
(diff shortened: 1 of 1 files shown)


src/vorta/utils.py
```diff
@@ -84,6 +84,8 @@ def get_private_keys():
     available_private_keys = []
     if os.path.isdir(ssh_folder):
         for key in os.listdir(ssh_folder):
+            if not os.path.isfile(os.path.join(ssh_folder, key)):
```
ThomasWaldmann: there is another join() some lines below, so it would be better to just compute it once.

m3nu:
Thanks for handling this case @sebcode!

PR #150, opened 2019-01-20
Programmatically detect theme color (for macOS)

sebcode:
When I first ran Vorta, I didn't see the macOS menu bar icon at all because I'm using dark mode. So I thought the app doesn't work. Until I realized that I have to configure the app explicitly to use a light icon. Could be confusing for end users.

This code sets `use_light_icon=True` if macOS dark mode is detected on the first run of the app.

### DIFF
--- a/src/vorta/models.py
+++ b/src/vorta/models.py
@@ -242,7 +242,7 @@ def init_db(con):
     for setting in get_misc_settings():
         s, created = SettingsModel.get_or_create(key=setting['key'], defaults=setting)
         if created and setting['key'] == "use_light_icon":
-            s.value = uses_dark_mode()
+            s.value = bool(uses_dark_mode())
         s.label = setting['label']
         s.save()
 
--- a/src/vorta/models.py
+++ b/src/vorta/models.py
@@ -11,7 +11,7 @@ import peewee as pw
 from playhouse.migrate import SqliteMigrator, migrate
 
 from vorta.i18n import trans_late
-from vorta.utils import slugify
+from vorta.utils import slugify, detect_use_light_icon
 
 SCHEMA_VERSION = 10
 
@@ -241,6 +241,8 @@ def init_db(con):
     # Create missing settings and update labels. Leave setting values untouched.
     for setting in get_misc_settings():
         s, created = SettingsModel.get_or_create(key=setting['key'], defaults=setting)
+        if created and setting['key'] == "use_light_icon":
+            s.value = detect_use_light_icon()
         s.label = setting['label']
         s.save()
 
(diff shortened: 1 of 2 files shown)


src/vorta/models.py
```diff
@@ -241,6 +241,8 @@ def init_db(con):
     # Create missing settings and update labels. Leave setting values untouched.
     for setting in get_misc_settings():
         s, created = SettingsModel.get_or_create(key=setting['key'], defaults=setting)
+        if created and setting['key'] == "use_light_icon":
```
m3nu: Instead of adding a condition, why not change the default setting directly [here](https://github.com/borgbase/vorta/blob/929248deee35d2ebd9a1dad1a07426fe749f44c3/src/vorta/models.py#L193)? E.g.

```
        {
            'key': 'use_light_icon',
            'value': detect_use_light_icon(),
...
```

src/vorta/utils.py
```diff
@@ -208,6 +208,18 @@ def set_tray_icon(tray, active=False):
     tray.setIcon(icon)
 
 
+def detect_use_light_icon():
```
m3nu: How about calling it `uses_dark_theme()` or something?

This accidentally makes sense for the setting `use_light_icon` as well. `uses_dark_theme() = use_light_icon = True`

m3nu:
We had the same issue on Linux. I don't know how to detect the theme color there, but your function can be extended for this case in the future. So thanks for that.

src/vorta/utils.py
sebcode: Sure, I renamed it to uses_dark_mode

src/vorta/models.py
m3nu: Because I put it in a stupid place in `src/vorta/views/misc_tab.py`.

Performance-wise you're right. Then keep it as you first had it.

sebcode:
Now that I think about it, it would actually be better to have a proper "template" menu bar icon for macOS, i.e. an image works in light and dark mode. That way the auto detection would be redundant (at least for macOS).

m3nu:
If Qt supports it. The icon is implemented as `QSystemTrayIcon`.

src/vorta/utils.py
```diff
@@ -208,6 +208,17 @@ def set_tray_icon(tray, active=False):
...
 
+def uses_dark_mode():
+    """
+    This function detects if we are running in dark mode (e.g. macOS dark mode).
+    """
+    if sys.platform == 'darwin':
+        from Foundation import NSUserDefaults
+        stdud = NSUserDefaults.standardUserDefaults()
+        return stdud.stringForKey_("AppleInterfaceStyle") == "Dark"
+    return False
```
ThomasWaldmann: could be:
True = dark
False = light
None = don't know

Plastix:
I just ran into this same issue. Thanks for addressing this in a PR @sebcode.

m3nu:
Good stuff. Thanks for the contribution. Will be even more useful together with #151 .

PR #152, opened 2019-01-20
Update .gitignore

Hofer-Julian:
Updated .gitignore according to [Github](https://github.com/github/gitignore/blob/master/Python.gitignore).

Also added entries for environments.

Only some entries are added, nothing should be deleted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,17 +1,46 @@
-.DS_Store
-.idea/
-bin/
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# Distribution / packaging
 build/
 dist/
+.eggs/
+lib/
+*.egg-info/
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Are you not using virtual environments?

Most of the rest is only reorganised.

If the  .spec file should be commited then of course this entry should be removed.

ThomasWaldmann:
I usually use virtual envs, but their directory is not inside the git workdir, so no problem with that. :)

m3nu:
Or [Conda](https://conda.io/miniconda.html) (popular with the data science crowd) or [Pyenv](https://github.com/pyenv/pyenv) with [pyenv-virtualenv](https://github.com/pyenv/pyenv-virtualenv).

I'll go ahead and close this for now, as it would be premature optimization.

Still thanks for the contribution, @Hofer-Julian . We have a lot of [good-first-issues](https://github.com/borgbase/vorta/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) at the moment. Maybe you find a nice one there.

PR #153, opened 2019-01-20
add title "Vorta for Borg Backup" to tray menu

ThomasWaldmann:
pun intended. :)


ThomasWaldmann:
Additional idea:
- remove the "Settings" menu entry
- connect showing the vorta window to "Vorta for Borg Backup" menu entry.

m3nu:
Here is how Nextcloud does it:

screenshot 2019-01-21 at 08 43 33

Keeping "Settings" is probably a good idea. Linking "Help" could make sense. There is already a tutorial. We also have the status display and "Quit"

The "Vorta for BorgBackup" item should open something useful if it's there. Borg website?

m3nu:
Here another one from "Coconut Battery"

screenshot 2019-01-21 at 10 57 15

m3nu:
This makes it more similar to the other apps.  Disabling the title looks alright. Setting the font bold didn't have an effect on macOS. Made it a bit smaller actually.

screen shot 2019-01-21 at 15 43 14

antlarr:
Hmm, I don't have any apps which show a title in the context menu, and in fact, all those examples above are actions, not titles. So I'm not sure this is ok. 

"Open CoconutBattery" does something as well as "Open Nextcloud in browser". Having an action that does nothing is... unexpected. Maybe we can assume that the main Vorta window is not just for settings anymore and it's the Vorta application (which can be used to interact with borg backups as in mounting/unmounting/extracting archives), so it can be moved up to the first position and renamed to "Open Vorta for BorgBackup"

m3nu:
I think the main purpose here is to add some Borg branding, which is OK and important.

Maybe the window title is a better place than the context menu? Like so:

screen shot 2019-01-21 at 18 49 03

Restructuring the context menu a bit, like above is still a good idea too.

ThomasWaldmann:
As I said in the ticket, the intention was to make it clear what this tray item is (and also have it look a bit less ugly than something starting with a disabled (low contrast) item that is hard to read).

If you just have "Settings" and "Next task" and some relatively generic icon, this is not really clear before you actually click on (now) "Settings".

Actually, "Settings" is wrong. If you click on it, a window opens where you can do all sorts of stuff (backup, restore, list, check) and also configure some settings.

So actually, what really happens is that the main application window opens (so saying "Open Vorta" would be already better IF the user knows what Vorta is). Not always is the person who currently looks through the tray items also the person who installed the stuff / who knows what is for what task.

My intention by using "Vorta for Borg Backup" was to have some pun (sorry, if it did not get over) and also to give some context "ah, this is a backup tool!".

ThomasWaldmann:
About adding "Help" (show Vorta builtin or online help): yes, can be done later within a separate PR.

OTOH, Vorta has some "more info" links already in the app window.

"Vorta for Borg Backup" as app window title: can be done, but keep that last blank. :)

ThomasWaldmann:
Collapsed the changesets. Anything within the scope of my changes still to do?

Shall I change the main window title to "Vorta for Borg Backup"?

Can do that for consistency, but guess it is not really required.

m3nu:
I'm good to go ahead. This kept me up at night, but I didn't come up with something smarter.

I've renamed "Exit" to "Quit" (to match the other apps) and changed the window title.

ThomasWaldmann:
Fixed the pep8 issue.

In my part of the changes:
```
self.tr('Vorta for Borg Backup')
```
@m3nu did not translate the window title.

Just noticed that the pun does not work if this gets translated.

So, we do not use tr() for this and always keep this in English?
Or translate at both places, let it to the translator?

ThomasWaldmann:
What I meant is that "Vorta for Borg Backup" in English (AFAIK) also has the meaning that the Vorta are aiding/backing the Borg (or shall do so).

That pun is gone if you translate only one word of it and if it is only the "for", because then the sentence is not read as English any more.

m3nu:
> the Vorta are aiding/backing the Borg (or shall do so).

I guessed so. But "for" has too many meanings. "X for president", "change X for Y", "get X for Y". And "Vorta" not being as well known as "borg" won't help either.

PR #156, opened 2019-01-21
Improve archive mount feature. Fixes #131. By @antlarr

### LINKED ISSUE #131: Display mounted Archives, look for existing mounts on startup
(copied from #129)

The following improvements to the mount-feature are suggested:

- Add some kind of mounted icon to the archive table when an archive is mounted, so the user can quickly see which are mounted.
- Since the dictionary of mount points per archive is stored in-process, it's lost when the user restarts Vorta. Which means that upon restart, all archives appear as unmounted and the user has to resort to the command line to unmount them, which is not nice. I would like to use psutil to parse the running commands on the system, search for "borg mount" commands and extract the archive names/mount points from there so the information used by Vorta is always correct (I would love to see any other way to get the archive associated with a mount point, but I'm afraid I couldn't find any other way to get that information but to parse command lines).

m3nu:
- Find existing mount points after restart.
- Show mount point in table.
- Add context menu for archive table.

### DIFF
--- a/src/vorta/assets/UI/archivetab.ui
+++ b/src/vorta/assets/UI/archivetab.ui
@@ -94,6 +94,11 @@ font-weight: bold;
          </column>
          <column>
           <property name="text">
+           <string>Mounted</string>
+          </property>
+         </column>
+         <column>
+          <property name="text">
            <string>Name</string>
           </property>
          </column>
--- a/src/vorta/borg/umount.py
+++ b/src/vorta/borg/umount.py
@@ -1,3 +1,4 @@
+import os.path
 import psutil
 from .borg_thread import BorgThread
 from ..i18n import trans_late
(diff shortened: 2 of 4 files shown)


antlarr:
I added the menu because I wanted to have an easy way to open the mount point with a file browser but I felt that adding another button would clutter the window. Later I added the icon in the tablewidget (which is now changed to the mount point text) so it's not really necessary anymore. I think it might be useful in the future to add more non-standard actions (i.e. 'Info...' option that opens a dialog showing details about that archive). In any case, it's really not needed right now anymore so feel free to remove that code if you prefer to have less code to maintain.

antlarr:
Also, I haven't tested the code yet, but I understand you now always show the last 5 characters of the mount point in the table, right? I'll wait for you to submit this to propose a change if you don't mind: Making that column resizeable and show an ellipsis and whatever characters fit in there. So if it's small, it would show '...older' but if it's large it shows '...p folder' and if it's large enough '/backup folder', would that be ok?

m3nu:
Quick response.. You're right about the "Open" functionality. If you're not too attached to the context menu, I would suggest this:

- remove the context menu. It's just too unexpected and undiscoverable.
- improve the "mounted" column to: a) show the path depending on column width and b) add an icon that shows that the cell can be used to open the folder. E.g. [this one](https://fontawesome.com/icons/external-link-square-alt?style=solid).

m3nu:
OK. Hyperlink sounds good. Will merge this after checks are done.

PR #160, opened 2019-01-22
implement TRANS_SCALE env var for i18n UI layout checking

ThomasWaldmann:
check how the UI behaves for longer / shorter strings, for RTL strings.

does currently only work IF a language is loaded from a .qm file, otherwise there won't be a translator object / it won't call its .translate method.

works:
```
LANG=de_DE TRANS_SCALE=200 vorta --foreground
```

### DIFF
--- a/src/vorta/i18n/__init__.py
+++ b/src/vorta/i18n/__init__.py
@@ -8,11 +8,45 @@ from PyQt5.QtCore import QTranslator, QLocale
 
 logger = logging.getLogger(__name__)
 
+# To check if the UI layout copes flexibly with translated strings, an env var
+# TRANS_SCALE can be used - it will result in transformed translated strings, e.g.:
+# 100 = scale 100%, normal
+# N>0 = scale to N%
+# N<0 = additionally, do a RTL simulation at scale -N%
+# usage: TRANS_SCALE=200 LANG=de_DE vorta --foreground
+trans_scale = int(os.environ.get('TRANS_SCALE', '100'))
+
+
+class VortaTranslator(QTranslator):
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
result of a quick test with this: quite a lot of the layout works well, but there are some places where it does not adapt, but e.g. truncate.

src/vorta/i18n/__init__.py
```diff
@@ -8,11 +8,45 @@
 
 logger = logging.getLogger(__name__)
 
+# To check if the UI layout copes flexibly with translated strings, an env var
+# TRANS_SCALE can be used - it will result in transformed translated strings, e.g.:
+# 100 = scale 100%, normal
+# N>0 = scale to N%
```
ThomasWaldmann: in fact N>=0, but 0 is not that useful.

m3nu:
Yeah. Mostly works. Misc needs line wrapping.

m3nu:
Cool idea. Thanks for that. I'll add a note about usage to the Translation section in CONTRIBUTING.md

I did notice that the `LANG` env var doesn't have an effect on macOS. I may explicitly add it to your class.

m3nu:
I know. Guess Qt implements it somewhere.

m3nu:
- add `LANG` env to debug on macOS
- added some instructions on how to use the new `LANG` and `TRANS_SCALE`

One thing that bit me today: when the line number in the `.ts` (and their corresponding `.qm`) are out of sync with the source code, translations don't show up. We don't commit those files, but mentioning it here and added it to CONTRIBUTION as well.

Issue #136, opened 2019-01-18
i18n: check original english texts carefully


ThomasWaldmann:
if we have to change original texts, all translators will have to update their translations, so better do this before it gets to the translators. agreeing on some terminology and checking for consistency everywhere is also useful (e.g. don't call it archive here and snapshot there if you mean the same thing).

for this:
- the english original texts should be exported (to a .ts file) and spell and grammar checked by a good native speaker. this includes stuff like upper / lowercase and punctuation.
- make a glossary for better consistency?
- there are some places using `doing foo..` and some `doing foo...` - use 3 dots consistently.

ThomasWaldmann:
- fix "Password" to "Password:" (init new backup repo dialogue)

m3nu:
That would be my mess.

ThomasWaldmann:
The hint about "FUSE for macOS" is too specific / not helpful on other OS.

ThomasWaldmann:
The longer text shown when adding a repo talks about using a thumb drive for full backups.
While it is only an example, I'ld like to change that (not that someone takes this as a recommendation or "the usual way").

I think we should advise using a 2.5" USB3 HDD for full backups:
- the usually quite big capacity allows for full backups without the risk of being too small
- it does not require a separate power supply (compared to 3.5" USB HDDs)
- it is not lost as easily as a thumbdrive
- it doesn't die as often as thumbdrives (from my experience, YMMV)

m3nu:
> The hint about "FUSE for macOS" is too specific / not helpful on other OS.

It's [only shown](https://github.com/borgbase/vorta/blob/4f6137207b7580dda754f24c1243484a592e31f9/src/vorta/views/archive_tab.py#L39) for macOS. Nobody else will see it.

m3nu:
> I think we should advise using a 2.5" USB3 HDD for full backups:

Makes sense. Change it to:

> Backup profiles allow for granular backups from different sources to different destinations. You could e.g. back up essential documents to a remote repository via Wifi, while doing a full backup onto a local **2.5" hard drive or thumb drive**.

ThomasWaldmann:
BTW, these original string changes can be done in a separate PR, after #134 is merged. It already starts to diverge (see wording "snapshot" vs. "archive") with other PR.

m3nu:
Ok. Stopping now. I also worry about conflicts with #134.

ThomasWaldmann:
[...]

PR #161, opened 2019-01-22
Check English Strings 

m3nu:
Issue #136

### DIFF
--- a/src/vorta/i18n/ts/vorta.de.ts
+++ b/src/vorta/i18n/ts/vorta.de.ts
@@ -1167,11 +1167,6 @@
         <translation>Auch über erfolgreiche Hintergrund-Aufgaben benachrichtigen</translation>
     </message>
     <message>
-        <location filename="../../models.py" line="213"/>
-        <source>Add Vorta to Login Items in Preferences &gt; Users and Groups &gt; Login Items</source>
-        <translation>Starte Vorta automatisch bei der Anmeldung (via Einstellungen &gt; Benutzer und Gruppen &gt; Autostart)</translation>
-    </message>
-    <message>
         <location filename="../../models.py" line="218"/>
         <source>Check for updates on startup</source>
         <translation>Prüfe beim Start auf Aktualisierungen</translation>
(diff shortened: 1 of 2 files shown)


src/vorta/i18n/ts/vorta.de.ts
```diff
@@ -1144,47 +1139,47 @@
...
+        <source>Also notify about successful background tasks</source>
+        <translation>Auch über erfolgreiche Hintergrund-Aufgaben benachrichtigen</translation>
     </message>
     <message>
-        <location filename="../../models.py" line="222"/>
-        <source>Add Vorta to Login Items in Preferences &gt; Users and Groups &gt; Login Items.</source>
-        <translation>Füge Vorta zu Login Items in Preferences &gt; Users and Groups &gt; Login Items hinzu.</translation>
+        <location filename="../../models.py" line="213"/>
+        <source>Add Vorta to Login Items in Preferences &gt; Users and Groups &gt; Login Items</source>
+        <translation>Starte Vorta automatisch bei der Anmeldung (via Einstellungen &gt; Benutzer und Gruppen &gt; Autostart)</translation>
```
ThomasWaldmann: maybe translate the german back to english and use that as original string?

src/vorta/i18n/ts/vorta.de.ts
```diff
@@ -449,121 +434,116 @@
     </message>
     <message>
         <location filename="../../assets/UI/archivetab.ui" line="97"/>
+        <source>Mounted</source>
+        <translation>Eingehängepunkt</translation>
```
ThomasWaldmann: that seems strange at least.

m3nu: Haha. Should be "Einhängepunkt"

m3nu: It shows the mount path. So I figured it's better than just "Eingehängt", which implies yes/no.

m3nu: Fixed in TX.

ThomasWaldmann: Well, if it is the wrong term, the source string should be changed or it might be strange in all other (more close) translations.

m3nu: True. Let's use "Mount Point". This changed only yesterday.

ThomasWaldmann: also, the question is if that is just a description of the path "oh, this is a (potential) mount point" or a description of the current status "this is currently mounted".

m3nu: Changed

src/vorta/i18n/ts/vorta.de.ts
m3nu: Good idea. I also dropped the last part.

Issue #119, opened 2019-01-12
delete specific archive(s)


ThomasWaldmann:
**Is your feature request related to a problem? Please describe.**
It would be good to have a UI to delete a specific archive.

Prune is there, but does not / not always handle the use case, e.g. if you accidentally backed up some crap, improve your exclude list excluding the crap and back up again afterwards. Then you would like to delete the first backup and maybe not want to wait for / rely on prune doing that.

There might be also the case where you just want to manually get rid of something for other reasons, without killing the whole repo.

**Describe the solution you'd like**
Select 1 (or multiple) archives from the list, click "delete" button.

m3nu:
We may get some per-archive action icons in #131. Those could be easily extended to run a new `borg.delete` command.

m3nu:
Closing, since this was already implemented and merged in #164.

PR #164, opened 2019-01-23
Implement feature to delete specific archive

Hofer-Julian:
My try on fixing https://github.com/borgbase/vorta/issues/119.

I noticed that the size in the archive table makes no sense anymore when you delete one archive.

For example:
- two archives of 500B and 6MB
- 6MB archive gets deleted
- there is only one archive with 500B left which holds data of 6MB

### DIFF
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -12,6 +12,7 @@ from vorta.borg.check import BorgCheckThread
 from vorta.borg.mount import BorgMountThread
 from vorta.borg.extract import BorgExtractThread
 from vorta.borg.umount import BorgUmountThread
+from vorta.borg.delete import BorgDeleteThread
 from vorta.views.extract_dialog import ExtractDialog
 from vorta.i18n import translate
 from vorta.utils import get_asset, pretty_bytes, choose_file_dialog, format_archive_name, get_mount_points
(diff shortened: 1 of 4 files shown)


m3nu:
> * there is only one archive with 500B left which holds data of 6MB

Good stuff. Will test it later today. But looks as expected at first glance.

I wouldn't worry about the sizes in the table. They don't need to add up necessarily. It's just a historic record of *how much data was added* when creating this archive.

m3nu:
Travis is failing due to an [update to Pip](https://github.com/pyinstaller/pyinstaller/issues/4003) that breaks PyInstaller. Not clear how this will be fixed. We can just stay at a lower Pip version for now.

src/vorta/views/archive_tab.py
```diff
@@ -365,3 +368,28 @@ def row_of_archive(self, archive_name):
...
+            thread.result.connect(self.delete_result)
+            self._toggle_all_buttons(False)
+            thread.start()
+        else:
+            self._set_status("No archive selected")
+
+    def delete_result(self, result):
+        if result['returncode'] == 0:
+            self._set_status(self.tr('Archive deleted.'))
+            self.list_action()
```
m3nu: In theory we could just remove the archive from the local DB is the delete-command was successful without listing/syncing all archives.

On the other hand it saves some code and syncing archives is always a good idea.

m3nu:
Travis is happy again with pip 18.1.

m3nu:
Merged. Thanks a lot for the contribution, @Hofer-Julian

PR #169, opened 2019-01-27
Use ellipsis the Qt way in the archive table

antlarr:
Showing the last 5 characters of the mount points doesn't give much
information, so we now allow the user to resize the mount_point column
and allow Qt to draw an ellipsis and the number of characters that
fit in the column (or the whole mount point if it fits).


m3nu:
Cool. Didn't know this mode existed. Thanks for adding!

Issue #168, opened 2019-01-27
Crash when cancelling "add new profile"


Hofer-Julian:
**Describe the bug**
The window which opens after "add new profile" is clicked, crashes the whole program when you close it.

**To Reproduce**
Steps to reproduce the behaviour:
1. Click "Add new profile"
2. Close window

**Desktop (please complete the following information):**
 - OS: Fedora 29

m3nu:
Yepp. When you "Cancel" adding the profile. Probably a small issue when processing the result.

Assigning to myself and will fix in the next update.

Thanks for reporting!

m3nu:
PS: If you want to work on this @Hofer-Julian, be my guest. It's likely a very small issue. Let me know here in the next 9 hours. Else, I'll tackle it tomorrow morning.

Hofer-Julian:
Yeah, I'll try it tomorrow. If I can't figure it out, I'll write it here :)

m3nu:
Fixed via #170

(Another pro-tip: If you write "... Fixes #1xx" in your commit message, it will automatically close the issue after merging it.)

PR #170, opened 2019-01-27
Fix Issue "Crash when cancelling "add new profile" #168"

Hofer-Julian:
Fix Issue https://github.com/borgbase/vorta/issues/168


m3nu:
Awesome.

You do have some extra commits in your PR. The recommended way is usually to pull/rebase from the current master branch and then make a new feature-branch in your personal repo. Then make a PR from the feature branch to the official branch, which will only have that one relevant commit. In this case we're lucky and Github filtered the old commits anyways.

Hofer-Julian:
Yeah, I found the procedure you mentioned in the internet after I saw this mess, but I didn't know how to fix it afterwards :)

PR #174, opened 2019-01-31
Fix crash due to zombie processes

lamyj:
In `vorta.utils.get_mount_points` the call to `process.name()` may raise an exception (e.g. zombie processes on macOS). In that case, Vorta will crash. This PR fixes this by catching exceptions in `process.name()`: if we can't get the process name, this is obviously not an active borg mount.


m3nu:
Good find. Actually happened to me the other day and I was about to look into it. Thanks!

PR #176, opened 2019-02-01
Ask for confirmation on delete (and small fixes)

antlarr:
The delete button might be dangerous for some users who might click it accidentally. Now the user is asked for confirmation before performing a destructive irreversible action.

Also, some fixes: remove unused code, translate a missing text which wasn't translated and remove a duplicated line (which was setting an untranslated status text just after the translated text was set).

### DIFF
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -2,7 +2,7 @@ import os.path
 import sys
 from datetime import timedelta
 from PyQt5 import uic, QtCore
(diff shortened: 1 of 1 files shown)


m3nu:
Great idea. And not even KDE-specific ;-)

antlarr:
> Great idea. And not even KDE-specific ;-)

LOL. You'll never forgive me for those KDE-specific lines ;-) (in my defense, vorta now behaves great on KDE :) )

m3nu:
Thanks. Nice commit.

PR #184, opened 2019-02-08
compressor selection: add zlib, add zstd,8, descriptions, fixes #145

### LINKED ISSUE #145: zlib selection UI, compression levels, descriptions
I just noticed that the UI for it is missing, it just offers all the others.

Also, I am not quite sure about the UI texts for the compression methods, my idea is rather:

- lz4 (modern, super fast low compression)
- zstd (modern, wide range from very fast to very high compression)
- lzma (legacy, slower higher compression)
- zlib (legacy, faster lower compression)

ThomasWaldmann:
see comments in src.

### DIFF
--- a/src/vorta/i18n/__init__.py
+++ b/src/vorta/i18n/__init__.py
@@ -18,6 +18,10 @@ trans_scale = int(os.environ.get('TRANS_SCALE', '100'))
 
 
 class VortaTranslator(QTranslator):
+    """
+    Extends QTranslator to increase the length of strings for testing. Fallback for untranslated
+    strings to English doesn't work currently. So only use for testing.
+    """
     def translate(self, context, text, disambiguation=None, n=-1):
         translated = super().translate(context, text, disambiguation=disambiguation, n=n)
         scale = trans_scale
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
i didn't research yet what maximum zstd level makes sense for borg (we typically have ~2MiB data to compress).

m3nu:
Added a few things:

- Database migration to map old values to new ones.
- Only use `VortaTranslator` when scaling option is used. Else it won't fall back to EN strings. At least on macOS.
- Pin PyQt5 version for tox tests to avoid [errors on Travis' old Ubuntu](https://bugreports.qt.io/browse/QTBUG-69100)

src/vorta/i18n/__init__.py
```diff
@@ -21,8 +21,6 @@ class VortaTranslator(QTranslator):
     def translate(self, context, text, disambiguation=None, n=-1):
         translated = super().translate(context, text, disambiguation=disambiguation, n=n)
         scale = trans_scale
-        if scale == 100:  # normal, production usage
-            return translated
```
ThomasWaldmann: so, what's wrong with that? that should implement base class behaviour.

src/vorta/i18n/__init__.py
```diff
@@ -46,7 +44,11 @@ def translate(self, context, text, disambiguation=None, n=-1):
 def init_translations(app):
     global application, translator, locale  # if we don't keep a reference on these, it stops working. pyqt bug?
     application = app
-    translator = VortaTranslator()
+    if trans_scale == 100:
+        translator = QTranslator()
+    else:
+        translator = VortaTranslator()  # Doesn't fall back to EN on empty strings.
+
```
ThomasWaldmann: would be interesting to find out why this changes observed behaviour.

m3nu: It would. Some classes are just not meant to be subclassed by mere mortals.

ThomasWaldmann: nitpick:
```
        translator = QTranslator() if trans_scale == 100 else VortaTranslator() 
```

m3nu:
Rebased and merged. I wonder why there is no "Rebase, squash and merge" option on Github.

Issue #96, opened 2018-12-17
Customize base borg command to add additional options


soakes:
Hi,

I have been trying to setup Vorta with rsync.net, however it does work but its using the old 0.x server with a lot of errors in the log and I just hate errors :)

```
2018-12-17 20:25:38,670 - WARNING - Remote:
2018-12-17 20:25:38,670 - WARNING - Please note:
2018-12-17 20:25:38,670 - WARNING - If you see a TypeError complaining about the number of positional arguments
2018-12-17 20:25:38,670 - WARNING - given to open(), you can ignore it if it comes from a borg version < 1.0.7.
2018-12-17 20:25:38,670 - WARNING - This TypeError is a cosmetic side effect of the compatibility code borg
2018-12-17 20:25:38,671 - WARNING - clients >= 1.0.7 have to support older borg servers.
2018-12-17 20:25:38,671 - WARNING - This problem will go away as soon as the server has been upgraded to 1.0.7+.
```
In the welcome email, they state that the following:

> If you need to use borg 1.x we recommend running your command as follows:
> 
> borg COMMAND --remote-path=borg1 ...
> 
> No special options are required to run borg 0.x

I have been trying to figure away on how to do this with Vorta.  From what I can see there is no option in the client and so would you consider in adding this in the future? meanwhile I can get away with a scripted solution but would be nice to have a UI.

Kind Regards,

Simon

ThomasWaldmann:
guess the whole (local) borg base command should be configurable and default to just `borg`.

soakes:
That would be perfect.

m3nu:
This would be done [here](https://github.com/borgbase/vorta/blob/master/src/vorta/borg/borg_thread.py#L119). Currently it looks for a global `borg` bin, then an embedded one (pyinstaller) and then fails.

soakes:
This is awesome thank you.  Will give it a whirl.

adamluk:
Would it be possible to have an explanation of how to utilise #109 with --remote-path=borg1 please?

m3nu:
#109 is unrelated to this. If you want to specify additional CLI options (like `--remote-path`) it would be covered in this issue. Hoping to implement it soon. Of course PRs are also welcome by others.

soakes:
Thank you guys.

m3nu:
Hey @soakes, @adamluk , this feature was now implemented by @ethernet-zero in #193 . Please have a look if it works well for you. Then we can merge it.

[...]

PR #193, opened 2019-02-15
Add borg extra arguments (#96)

ethernet-zero:
Added a field to the repo information so additional arguments can be
added to the borg command line. The arguments specified there are
appended to every command that's run on that repo.

### DIFF
--- a/src/vorta/borg/borg_thread.py
+++ b/src/vorta/borg/borg_thread.py
@@ -43,7 +43,8 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
         self.app.backup_cancelled_event.connect(self.cancel)
 
         extra_args = shlex.split(params.get('extra_borg_arguments', ''))
-        cmd = [self.prepare_bin()] + extra_args + cmd[1:]
+        cmd[0] = self.prepare_bin()
+        cmd = cmd[:2] + extra_args + cmd[2:]
 
         env = os.environ.copy()
         env['BORG_HOSTNAME_IS_UNIQUE'] = '1'
--- a/src/vorta/borg/borg_thread.py
+++ b/src/vorta/borg/borg_thread.py
@@ -2,6 +2,7 @@ import json
 import os
 import sys
 import shutil
+import shlex
 import signal
 import select
 import time
@@ -110,6 +111,7 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
         ret['ssh_key'] = profile.ssh_key
         ret['repo_id'] = profile.repo.id
         ret['repo_url'] = profile.repo.url
+        ret['extra_borg_arguments'] = profile.repo.extra_borg_arguments
         ret['profile_name'] = profile.name
 
         ret['ok'] = True
@@ -142,8 +144,10 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
         log_entry.save()
         logger.info('Running command %s', ' '.join(self.cmd))
 
-        p = Popen(self.cmd, stdout=PIPE, stderr=PIPE, bufsize=1, universal_newlines=True,
+        extra_args = shlex.split(self.params.get('extra_borg_arguments', ''))
+        p = Popen(self.cmd + extra_args, stdout=PIPE, stderr=PIPE, bufsize=1, universal_newlines=True,
                   env=self.env, cwd=self.cwd, start_new_session=True)
+
         self.process = p
 
         # Prevent blocking of stdout/err. Via https://stackoverflow.com/a/7730201/3983708
--- a/src/vorta/borg/borg_thread.py
+++ b/src/vorta/borg/borg_thread.py
@@ -42,7 +42,8 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
         self.app = QApplication.instance()
         self.app.backup_cancelled_event.connect(self.cancel)
 
-        cmd[0] = self.prepare_bin()
+        extra_args = shlex.split(params.get('extra_borg_arguments', ''))
+        cmd = [self.prepare_bin()] + extra_args + cmd[1:]
 
         env = os.environ.copy()
         env['BORG_HOSTNAME_IS_UNIQUE'] = '1'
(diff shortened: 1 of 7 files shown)


src/vorta/views/repo_tab.py
```diff
@@ -129,7 +129,7 @@ def ssh_copy_to_clipboard_action(self):
 
     def compression_select_action(self, index):
         profile = self.profile()
-        profile.compression = self.repoCompression.currentData()
+        profile.compression = self.repoCompression.currentData() or ''
```
ThomasWaldmann: this seems to be unrelated to the topic of your PR.

ethernet-zero: Oops, you're right. I had to add that because after applying migrations to the settings database the program ended with an error on that line because of the NOT NULL constraint on that field.

Should I remove that file from the PR?

m3nu: We recently changed the available compression options. They should be migrated when you run a version later than commit 3efecd71f3b0fc69261284efcd8d628bb2b89d59.

src/vorta/borg/borg_thread.py
```diff
@@ -142,7 +144,8 @@ def run(self):
         log_entry.save()
         logger.info('Running command %s', ' '.join(self.cmd))
 
-        p = Popen(self.cmd, stdout=PIPE, stderr=PIPE, bufsize=1, universal_newlines=True,
+        extra_args = shlex.split(self.params.get('extra_borg_arguments', ''))
+        p = Popen(self.cmd + extra_args, stdout=PIPE, stderr=PIPE, bufsize=1, universal_newlines=True,
```
m3nu: This appends extra args at the very end. Like `borg create --debug $REPO::$SNAPSHO --rsh=xxx`.

Another option would be to append them in `vorta.borg.borg_thread.prepare_bin()`. Because extra args apply to *all* different `borg` commands, right?

ethernet-zero: They do.

Sure, it'd work either way. I added them at the end to not change the current interface of `prepare_bin()`.

ThomasWaldmann: please check "order matters" in the borg docs.

m3nu: Why not insert the extra args around [Line 44 in this file](https://github.com/borgbase/vorta/blob/master/src/vorta/borg/borg_thread.py#L44). This is where global options are added. Python also has a helper function to insert elements into a list, like `self.cmd`

```
# (existing) change path to Borg binary, if needed
cmd[0] = self.prepare_bin()

# insert extra args
cmd.insert(1, params.['extra_borg_arguments'])  # TODO: skip if no extra args, split args
```

ethernet-zero: I agree. @ThomasWaldmann you're right about the order of the arguments, I fixed it.

@m3nu I moved constructing the command line to where you suggested. 👍

m3nu:
We're getting there. When testing I noticed:

- The "Add/Edit Repo" repo window/sheet is fairly overloaded with the new field. Most users don't need this field at all and may be confused by it. So I suggest to hide it under "Advanced" or something. Need to see what options Qt offers here. I'm sure there area a few.
- The field label says **Extra Borg arguments:**, but CONTRIBUTING.md says: *Field labels (same or next line) end with a colon and are titleized. "Allowed Networks:"*. So should be **Extra Borg Arguments:**

m3nu:
Tests are passing now. I wonder if there should be a test to see if the Borg command was correctly contructed given a repo with some custom commands.

ethernet-zero:
>     * The "Add/Edit Repo" repo window/sheet is fairly overloaded with the new field. Most users don't need this field at all and may be confused by it. So I suggest to hide it under "Advanced" or something together with the SSH dropdown. Need to see what options Qt offers here. I'm sure there area a few.
If we moved the field to the Advanced tab, its contents would have to change with the selected repo, right? It'd make sense to put it there along with the pre- and post-backup commands, but since it'd be tied to the repo and its contents would change, it could be a good idea to clarify that in a small help text label or something like that.

>     * The field label says **Extra Borg arguments:**, but CONTRIBUTING.md says: _Field labels (same or next line) end with a colon and are titleized. "Allowed Networks:"_. So should be **Extra Borg Arguments:**
Right, fixing it.

m3nu:
> If we moved the field to the Advanced tab, its contents would have to change with the selected repo, right? It'd make sense to put it there along with the pre- and post-backup commands, but since it'd be tied to the repo and its contents would change, it could be a good idea to clarify that in a small help text label or something like that.

I didn't mean moving it to some tab. I meant adding a small label "Advanced" in the **same** window and only showing the SSH key- and extra args fields when the user clicks it.

m3nu:
screen shot 2019-02-19 at 18 47 07

A *Basic*/*Advanced* tab could work. (I'm not saying that you need to code everything yourself. We just need a good solution. Implementation is usually the easy part)

ethernet-zero:
Ah, alright, got it. :smile:

m3nu:
Maybe the "Encryption" selection is advanced too. So only URL and Password are required. That could work and we can add more advanced options later.

Coding-wise the layout changes don't affect the code at all, as long as the field names stay the same.

ethernet-zero:
Could this work?




I left the SSH key selection in the Basic tab to not leave one tab with three fields and another tab with just one, but it could be moved to Advanced just as easily.

ethernet-zero:
And this would be when moving the SSH key field to the Advanced tab:




Which one do you prefer?

m3nu:
Having the URL separate on top seems unusual and makes it harder to understand. I'd think about it in terms of Required + optional parameters. When moving the URL up, you are creating a third group.

How about:

- Tab *General*/*Required*/*Basic*/...: URL, password
- Tab *Advanced*: SSH, extra params, encryption

I looked into some other apps I use to see how they solved it. Also to point out that this is not about *my* preference. It needs to make sense for whoever is using it to enable them to quickly do the task they want to do.

screen shot 2019-02-20 at 13 07 57
screen shot 2019-02-20 at 13 09 26
screen shot 2019-02-20 at 13 11 43
screen shot 2019-02-20 at 13 12 32

ethernet-zero:
There you go:

ethernet-zero:
I fixed the lines that build the borg command line so the extra arguments are placed right after the subcommand, because putting them after the borg binary messed up the activity log.

m3nu:
Rebased and tried to clean up the repo-add window a bit.

m3nu:
While working on another issue, I had Vorta hanging before every call to Borg. Seems to happen when `shlex.split` is called on `None`. Same thing happens in iPython.

```
In [1]: import shlex
In [2]: shlex.split(None)  # will hang here
```

Since the DB field was set to allow null, `None` was getting passed. I changed this now to only have the default value of an empty string.

I didn't see this hanger before, but I'll include this change in another PR just in case.

ethernet-zero:
It seems that passing `None` to `shlex.split()` makes it read from the standard input, that's why it hangs. Good catch. :smile:

PR #196, opened 2019-02-17
Remove keyring dependency - continues #195 by @Hofer-Julian

m3nu:
Clean rebase.


m3nu:
(I was too excited to test this, so rebased it myself. Sorry.. You should still try to try out the `cherry-pick` method I mentioned to clean up the other PR anyways.)

Regarding this commit: 

- I don't think we need to copy the code from the `keyring` package (files `keyring_[backend,core,errors].py`). Copying code can bring licensing and copyright issues. Literally the only thing we reuse is 2 method names. The errors aren't used anywhere.
- About the Abstract base class: This is a good way to define a class interface and make sure all methods are defined in all subclasses. Since we only have 2 methods and 3 classes it may be overkill.
- Next, it would be nice to move the code for class selection OUT of `vorta.utils` into the a new base class, making it a [factory pattern](https://python-3-patterns-idioms-test.readthedocs.io/en/latest/Factory.html) ([Gamma, 1994, ch3](https://www.amazon.com/Design-Patterns-Elements-Reusable-Object-Oriented/dp/0201633612)).

Here a stub of how it could look like:

```python
class VortaKeyring:
    @classmethod
    def get_keyring(cls):  # or choose_keyring(), factory(), etc
        # code to choose correct subclass
        # if sys.platform == 'darwin':
        #   ...
        #   return DarwinKeyring()

    def set_password(self):
        # to be overriden
        raise NotImplementedError
        
    def get_password(self):
        raise NotImplementedError

class DarwinKeyring(VortaKeyring):
    def set_password(self):
        # actual implementation
```

Then when starting the app in `vorta.utils` or `vorta.application` we just do:

`keyring = VortaKeyring.get_keyring()  # variable keyring will be an instance of the correct subclass`

Benefits:
- All Keyring-related stuff is in one place, including selecting it.
- No need to copy stuff from this other package
- Each Keyring subclass can have additional implementation details that are hidden.
- We can select and instantiate the correct subclass ONCE when the app starts. This also solves the dbus-issue you described.
- Nothing from your new `keyring` package needs to be imported except the base class – `VortaKeyring`

I hope this is not total BS and I'd love for @ThomasWaldmann to weigh in. Looks like we stumbled into a full-fledged CS lesson here. Really lucky.

Hofer-Julian:
>     * I don't think we need to copy the code from the `keyring` package (files `keyring_[backend,core,errors].py`). Copying code can bring licensing and copyright issues. Literally the only thing we reuse is 2 method names. The errors aren't used anywhere.
Yeah, I already wondered about that, but I thought because of it's MIT licence it should be fine.
But I prefer your solution, which doesn't involve copying.
> 
>     * About the Abstract base class: This is a good way to define a class interface and make sure all methods are defined in all subclasses. Since we only have 2 methods and 3 classes it may be overkill.
That might be true.
> 
>     * Next, it would be nice to move the code for class selection OUT of `vorta.utils` into the a new base class, making it a [factory pattern](https://python-3-patterns-idioms-test.readthedocs.io/en/latest/Factory.html) ([Gamma, 1994, ch3](https://www.amazon.com/Design-Patterns-Elements-Reusable-Object-Oriented/dp/0201633612)).
Seems, like I have to reread the [realpython-tutorial](https://realpython.com/factory-method-python/).

>     * We can select and instantiate the correct subclass ONCE when the app starts. This also solves the dbus-issue you described.
I am glad, that this should solve the issue. Currently I have no idea, why it occurs in the first place.

m3nu:
Closed the old PR. I saw the Dbus issue first for notifications. As long as it's initialized during startup, it's fine.

m3nu:
Still wrapping this up, @Hofer-Julian ? Else I'll just do the class revamp. You already did the main part anyways. The rest is just some refactoring.

Hofer-Julian:
I haven't started yet, wasn't sure if we want to wait for feedback from @ThomasWaldmann.
I can start now.

src/vorta/keyring/keyring_errors.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class PasswordSetError(KeyringError):
+    """Raised when the password can't be set.
+    """
+
+
+class PasswordDeleteError(KeyringError):
+    """Raised when the password can't be deleted.
+    """
```
ThomasWaldmann: it looks like you don't have a delete method any more, so guess this is unused?

src/vorta/keyring/keyring_errors.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class InitError(KeyringError):
+    """Raised when the keyring could not be initialised
+    """
+
+
+class KeyringLocked(KeyringError):
+    """Raised when the keyring could not be initialised
+    """
```
ThomasWaldmann: c&p issue?

src/vorta/keyring/keyring_secretstorage.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def get_password(self, service, repo_url):
+        connection = secretstorage.dbus_init()
+        collection = secretstorage.get_default_collection(connection)
+        attributes = {'application': service, 'repo_url': repo_url}
+        try:
+            items = collection.search_items(attributes)
+            return next(items).get_secret().decode()
+        except Exception:
+            return None
```
ThomasWaldmann: diaper pattern detected :)

src/vorta/utils.py
```diff
@@ -20,9 +20,9 @@
 from PyQt5.QtGui import QIcon
 from PyQt5 import QtCore
 import subprocess
-import keyring
-from vorta.keyring_db import VortaDBKeyring
-
+import vorta.keyring.keyring_core as keyring
+from vorta.keyring.keyring_db import VortaDBKeyring
+import secretstorage
```
ThomasWaldmann: if it is imported here, even darwin will require it - is this intended?

src/vorta/keyring/keyring_secretstorage.py
Hofer-Julian: I don't know what you mean with that^^

I included it, because @m3nu did something similar in keyring_db.py and I hoped it was the reason my implementation didn't work (now we know it is a issue with dbus).

src/vorta/utils.py
Hofer-Julian: Good point, I will change that.

src/vorta/keyring/keyring_secretstorage.py
m3nu: He means the `except Exception`. It should be more specific. "It's [called](https://mike.pirnat.com/2009/05/09/the-diaper-pattern-stinks/) a Diaper because it catches all the shit."

src/vorta/keyring/keyring_secretstorage.py
```diff
@@ -0,0 +1,22 @@
...
+    """A wrapper for the secretstorage package to support the custom keyring backend"""
+
+    def set_password(self, service, repo_url, password):
+        connection = secretstorage.dbus_init()
+        collection = secretstorage.get_default_collection(connection)
+        attributes = {'application': service, 'repo_url': repo_url}
+        collection.create_item(f"vorta-{repo_url}", attributes, password.encode(), replace=True)
+
+    def get_password(self, service, repo_url):
+        connection = secretstorage.dbus_init()
```
m3nu: Didn't you want to call `dbus_init()` only once at the beginning due to some issue and to test if DBUS is actually there? In that case it would be in `VortaSecretStorageKeyring .__init__()`. If it fails, it falls back to the database keystore, like we do already in `vorta.utils`

m3nu: According to the [docs](https://secretstorage.readthedocs.io/en/latest/) it's called only once and then passed around. So better to set it in `__init__()` and then use it for password requests.

> `secretstorage.dbus_init() ` [...] Returns a new connection to the session bus, instance of jeepney.DBusConnection instance. This connection can then be passed to various SecretStorage functions, such as get_default_collection().

Hofer-Julian:
I implemented the factory pattern.
Unfortunately the d-bus problem persists.

m3nu:
Looks good. I'll fix the macOS issue when testing it.

Tests on Linux are also passing. Do you still see the DBus-issue in connection with QThread that you mentioned? We only run `dbus_init()` once in the main thread now.

> The secretservice throws an exception when started from BorgThread. This can be tested by creating a new repository on a Linux machine. The processresult method in BorgInitThread will start set_password. This will start secretstorage.dbus_init() in VortaSecretStorageKeyring, which throws an unhandled exception. This does not happen when secretstorage.dbus_init() is started at any other part of the program.

Hofer-Julian:
Interesting. On my linux machine tests are failing and the DBus-issue persists

m3nu:
Rebased and fixed a small import-issue on macOS.

Will do a final review tomorrow and merge then.

m3nu:
Ended up making more changes than planned, so I'd appreciate some more testing. Changes I made at a glance:

- Renamed the modules in package `vorta.keyring`. Since the package is already called `keyring`, we don't need to call the module in it `keyring_xxx.py`.
- Workaround for an [issue](https://gitlab.com/takluyver/jeepney/issues/6) in the (young) DBus library, when calling it from a thread.
- DBus should be available on BSD desktops as well. So try to use it there.
- Remove `keyring` dependency and add `secretstorage` dependency.
- If password for a repo isn't found in the main Keystore: Fall back to the database, if it was used before. (I found a bug in loading the Keystore for Linux and users may have fallen back to storing it in the DB. If this is the case, there is a warning about this in the logs. New passwords will be saved to SecretStore in the future.)
- Remove some diaper exceptions I added earlier. In particular the exception regarding the macOS keychain shouldn't be needed any more.

I tested this on Ubuntu 18 and macOS. Since this is an essential feature and I want to avoid breaking anything.

Hofer-Julian:
Great! I will test it tomorrow :)

m3nu:
Note on Travis: I got DBus to work, but there is probably a Gnome prompt blocking me for something. Don't know how to get rid of it for now.

Hofer-Julian:
Everything works fine on Fedora 29.
Great work!

PR #202, opened 2019-02-20
prevent KeyError on macOS

ipwog:
On my Hackintosh there is no `KnownNetworks` in com.apple.airport.preferences.plist


m3nu:
Thanks. Merged.

PR #203, opened 2019-02-22
Expand user path in python before setting BORG_RSH

jonas-hagen:
We experienced some problems on MacOS 10.11.6, Borg 1.1.9 (installed with brew cask) where borg could not find the ssh key when there was a tilde in the path to the key file. Setting the `BORG_RSH` environment variable to expanded paths works. This patch fixes this for Vorta, we tested it on said setup.


m3nu:
Makes sense. Thanks.

m3nu:
I'll update Homebrew once this is merged. But should also get new version with the built-in updater.

jonas-hagen:
Thanks!

PR #208, opened 2019-02-26
Reduce memory usage by removing main window from memory when closed. …

### LINKED ISSUE #207: Reduce memory usage by keeping main window out of memory
I'm reporting this from Linux  (openSUSE Tumbleweed to be precise, if it matters).

On my system, Vorta uses about 215~220 MB of RAM when running, even when not performing any action and merely residing idle on the system tray.

This seems to be a bit high.

For comparison, KeepassXC uses about 25~30 kB; it is a different software, with a different purpose, but it behaves in a similar way: it resides on the tray when idle, and like Vorta it uses Qt.

m3nu:
…Fixes #207

### DIFF
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -50,9 +50,8 @@ class VortaApp(QApplication):
         self.setQuitOnLastWindowClosed(False)
         self.scheduler = VortaScheduler(self)
 
-        # Prepare tray and main window
+        # Prepare system tray icon
         self.tray = TrayMenu(self)
-        self.main_window = MainWindow(self)
 
         # Apply dark stylesheet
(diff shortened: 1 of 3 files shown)


ymage:
I have an issue with this fix. **main_window** attribute is not created when launched in background.
So when clicked in tray icon, it crashes with

`  File "/usr/lib/python3.7/site-packages/vorta/tray_menu.py", line 28, in on_activation
    self.app.toggle_main_window_visibility()
  File "/usr/lib/python3.7/site-packages/vorta/application.py", line 86, in toggle_main_window_visibility
    if self.main_window.isVisible():
AttributeError: 'VortaApp' object has no attribute 'main_window'`

Issue #126, opened 2019-01-15
offer fat binaries for all supported platforms


ThomasWaldmann:
It would be nice to have some (semi-)automated way to offer vorta (and borg) as pyinstaller fat binary for vorta releases:
- macOS (see also #79)
- Linux
- later: FreeBSD (see also #125)

Could use similar automation as seen in borg's Vagrantfile.

If easy enough, it could be done for every release or else just now and then.

ThomasWaldmann:
btw, recently wanted to install vorta for an ubuntu user.

after preparing virtual env etc. wondered why I can't `pip install vorta` and then noticed I have only python 3.5 on ubuntu xenial (similar problem on debian stretch btw).

so, a fat linux binary with everything would really help for such cases.

m3nu:
It's on my list. Really. :-)

Hofer-Julian:
I recently read this tutorial: https://www.mfitzp.com/article/packaging-pyqt5-apps-with-fbs/

Might this be interesting for vorta?

m3nu:
We already use Pyinstaller, which this project wraps. They just solve a few other problems on top of it. Some are already solved in Vorta (finding asset files, making installers), some aren't that relevant (shipping different files for each OS). Maybe these guys solve them better, but they also impose their folder structure on the project, which could be a bad thing.

Still good to know it exists. May be useful for a green field project that needs installers for many OSs.

Hofer-Julian:
I see :)

m3nu:
Added the [first Linux binary](https://github.com/borgbase/vorta/releases/tag/v0.6.10). It got fairly large. 2x of macOS. UPX or zipping didn't make much difference. I guess there are more libraries included.

Would be helpful is someone could test it on different Linux platforms, like Arch or Debian.

ThomasWaldmann:
I've seen a similar size difference for borg itself, so seems to be normal.

ThomasWaldmann:
debian stretch (whonix, to be precise):
```
$ ./vorta-linux64 
[2469] Error loading Python lib '/tmp/_MEIMBkOkG/libpython3.6m.so.1.0': dlopen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.25' not found (required by /tmp/_MEIMBkOkG/libpython3.6m.so.1.0)
```

[...]

PR #213, opened 2019-02-27
Fix autostart

Hofer-Julian:
As discussed in https://github.com/borgbase/vorta/issues/126


m3nu:
Love PRs that remove code.

PR #224, opened 2019-03-04
Allow mounting a whole repo again.

m3nu:
I didn't realize that #156 by @antlarr removed the option to mount a whole repo. I think this is useful functionality and I had someone ask for it today. So I have re-enabled it in this PR.

The UI could be improved a bit. Currently it works by pressing the "Mount" button, while no archive is selected. And it's not possible to unmount the whole from the UI. The latter could be added. I'm just not sure on how to deselect all archives in a sensible way.

### DIFF
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -207,7 +207,7 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
     def set_mount_button_mode(self, mode):
         self.mountButton.clicked.disconnect()
         mount = (mode == 'Mount')
-        self.mountButton.setText('Mount' if mount else 'Unmount')
+        self.mountButton.setText(self.tr('Mount') if mount else self.tr('Unmount'))
         self.mountButton.clicked.connect(self.mount_action if mount else self.umount_action)
 
     def mount_action(self):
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Maybe we could add a button (or something else) for mounting the whole repo.
At least for me it would not be obvious that the way to mount the whole repo would be to have no archive selected and then press the mount button.

m3nu:
Currently we have per-archive ops on the left side and whole-repo ops on the right. Sadly we're running out of space to add more buttons.

My suggestions would be to use smaller toolbox buttons, disable the left side when no archive is selected and possibly add some labels, like "Apply to selected Archive" and "Apply to whole repo".

Hofer-Julian:
> Maybe we could add a button (or something else) for mounting the whole repo.
> At least for me it would not be obvious that the way to mount the whole repo would be to have no archive selected and then press the mount button.

I changed my opinion. Maybe it is the best way to just mount the whole repo when no archive is selected.


> I'm just not sure on how to deselect all archives in a sensible way.

You can just select one archive at the same time right?
For me the most intuitive way to deselect it would be to click at it again.

m3nu:
Deselecting would be useful. Currently I close and reopen the window. I'll try to add this to the current PR.

Issue #233, opened 2019-03-20
Run automatically when pen-drive is plugged in


qazip:
**Is your feature request related to a problem? Please describe.**
I would like vorta to support running a backup whenever a pen-drive is plugged in.

**Describe the solution you'd like**
- Plug in pen-drive;
- Vorta backs up to pen-drive;
- Unplug pendrive;

m3nu:
I have the same use case and fully agree that this would be nice.

I guess the solution would be different for each OS.

It may be able to use PR #231 to solve this. It can basically detect if Vorta already runs and then trigger some action. In the PR a window is opened, but it could as well trigger a backup.

Then the user could use the Distro's way to handle mount hooks to run e.g. `vorta backup /mnt/xxx` or so.

qazip:
As long as plugging an usb-device (pen-drive) triggers the backup to the pen-drive, I am happy. It would be nice to have confirmation of backup at the end.

MatejKovacic:
For Linux you can create udev rule or run script when a specific UUID of a disk is detected. This latter would be even better.

m3nu:
Then it should be possible to have a checkbox in Vorta and write some templates into those folders if they exist. Any link to this mechanism? Is it widely supported?

MatejKovacic:
For UDEV rules:
1. Create a file `/etc/udev/rules.d/90-vorta.rules`

2. Content of this file:
`ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="1c9e", ATTR{idProduct}=="f000", RUN+="runvortabackup.service"`

3. Create file `/etc/systemd/system/runvortabackup.service`

4. Content of the file
```
[Unit]
Description=Run Vorta backup

[Service]
Type=oneshot
RemainAfterExit=no
ExecStart=/usr/local/bin/.... (enter here)

[Install]
WantedBy=multi-user.target

```


5. Reload UDEV rules:
`udevadm control --reload-rules`

I am not absolutely sure it this is 100% correct (I am writing this from my mind, however, this is one possibility. Another, **much better** is to detect a disk device with specific UUID. I will describe this later.

m3nu:
Cool. Thanks for sharing. Since this modifies system folders, Vorta can't automatically change this when it's running as user application.

But I could see this added to a kind of "Cookbook" at first, so users can do it as one-time setup manually. Then once there is some experience with it it could be integrated and maybe ask for root-permissions using some mechanism that Gnome or other desktops probably have.

[...]

Issue #132, opened 2019-01-17
Open main window when started for the first time


m3nu:
Not everyone is happy with the current behavior of going straight to the tray icon after starting. We actively decided NOT to show a window on startup, so that Vorta can just silently do it's job without wasting time for users who know what they are doing. For development purposes there is the `--foreground` option which will open a window on startup.

I still prefer to just minimize the window on startup, but would suggest the following changes:

- [ ] When no settings exist (i.e. the first time Vorta is started), show the settings window.
- [x] When Vorta is started a second time, also open the settings window. (Currently we just display an error message.

ThomasWaldmann:
Can there be conflicts if 2 vorta access the settings DB?

m3nu:
Technically it's [not an issue](https://www.sqlite.org/faq.html#q5) as long as they don't write at the same time. But logically it would be confusing if they both change different settings.

Hofer-Julian:
Actually I also think, that on startup, it would be preferable to have it hidden.

But if I would start it manually, regardless if it is already running or not, I would expect to see a GUI in the foreground.

ansemjo:
Current [Gnome Desktops](https://www.gnome.org/) simply do not have a tray area anymore - which left me quite confused after I first started Vorta because nothing visibly happened. Running `vorta` in a terminal then said it was running already. Killing the running instance and starting it again with `-f` pops up the main window.

I think showing the main window would be desirable. Just don't make it kill the background process when you exit the window and reopen said window when you start Vorta from the applications list again?

Hofer-Julian:
> Not everyone is happy with the current behavior of going straight to the tray icon after starting. We actively decided NOT to show a window on startup, so that Vorta can just silently do it's job without wasting time for users who know what they are doing. For development purposes there is the `--foreground` option which will open a window on startup.
> 
> I still prefer to just minimize the window on startup, but would suggest the following changes:
> 
>     * When no settings exist (i.e. the first time Vorta is started), show the settings window.
> 
>     * When Vorta is started a second time, also open the settings window. (Currently we just display an error message.

To solve the first, it would be good to have a "--background" instead of a "--foreground" option, right? We'd only need to change the behaviour of the autostart method for macOS if I'm not mistaken.



> While doing the translation work, I restarted vorta quite often. I noticed that some commandline option would be useful that starts it in foreground, GUI window visible and terminates when window is closed.

The next step would probably be to implement this. I would also find this function convenient.

m3nu:
[...]

PR #243, opened 2019-03-26
Implement QtSingleApplication

Hofer-Julian:
This implements similar functionalites as [QtSingleApplication](https://github.com/qtproject/qt-solutions/tree/master/qtsingleapplication) (adapted from https://stackoverflow.com/a/12712362/11038610) which unfortunately is not available for Python.

It solves part of https://github.com/borgbase/vorta/issues/132 and enables work on https://github.com/borgbase/vorta/issues/233.

In theory it should work on all Linux distributions, macOS and Windows.


Hofer-Julian:
@m3nu if this works on macOS it should be okay to merge this pull request :)

m3nu:
Fully agree. I just take some more time testing new PRs after the flood of hotfixes I had to do recently.

Hofer-Julian:
Completely understandable. Can I do more to help, apart from have it running on my machine?

(I am also currently looking into writing tests for the other functions I implemented)

m3nu:
I can confirm that your change works as promised on macOS and Ubuntu 18.04.

Also thanks for the offer. Since you're already maintainer, you can just jump on anything you see. Like better tests. Some issues are just dependent on the environment and hard to test.

Hofer-Julian:
> Some issues are just dependent on the environment and hard to test.

Some issues seemed to be KDE specific. I also noticed that vorta sometimes responds differently under Wayland than it does under X. Is it be possible to set up two Linux environments under Travis? Like Kubuntu 18.04 with KDE and X and Fedora 29 with GNOME and Wayland?

PR #250, opened 2019-04-08
Add helper class to manage Borg version and supported features.

### LINKED ISSUE #205: Only offer repokey-blake2 support to >=1.1.4 borg versions
**Describe the bug**

I had borg 1.0.x installed on my laptop. I tried to initialize a repo, and vorta defaults to `repokey-blake2`.
Initialising failed without any meaningful feedback (see #108 and #162). Looking at the console I found that `borg init` complained about `repokey-blake2` as encryption mode.
After upgrading borg to 1.1.x this was fixed.

 - OS: Ubuntu 18.04
- vorta 0.6.10, Python 3.6.7

m3nu:
Fixes #205

### DIFF
--- a/src/vorta/borg/_compatibility.py
+++ b/src/vorta/borg/_compatibility.py
@@ -1,7 +1,8 @@
 from pkg_resources import parse_version
 
 MIN_BORG_FOR_FEATURE = {
-    'REPOKEY_BLAKE2': parse_version('1.1.4'),
+    'BLAKE2': parse_version('1.1.4'),
+    'ZSTD': parse_version('1.1.4'),
     # add new version-checks here.
 }
 
--- a/src/vorta/borg/_compatibility.py
+++ b/src/vorta/borg/_compatibility.py
@@ -1,0 +1,24 @@
+from pkg_resources import parse_version
+
+MIN_BORG_FOR_FEATURE = {
+    'REPOKEY_BLAKE2': parse_version('1.1.4'),
+    # add new version-checks here.
+}
+
+
+class BorgCompatibility:
+    """
+    An internal class that keeps details of the Borg version
+    in use and allows checking for specific features. Could be used
+    to customize Borg commands by version in the future.
+    """
+
+    version = '0.0'
+    path = ''
+
+    def set_version(self, version, path):
+        self.version = version
+        self.path = path
+
+    def check(self, feature_name):
+        return parse_version(self.version) >= MIN_BORG_FOR_FEATURE[feature_name]
--- a/src/vorta/views/repo_add_dialog.py
+++ b/src/vorta/views/repo_add_dialog.py
@@ -1,5 +1,6 @@
 import re
 from PyQt5 import uic
+from PyQt5.QtWidgets import QApplication
 
 from ..utils import get_private_keys, get_asset, choose_file_dialog
 from vorta.borg.init import BorgInitThread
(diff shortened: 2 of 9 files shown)


src/vorta/borg/_compatibility.py
```diff
@@ -0,0 +1,24 @@
+from pkg_resources import parse_version
+
+MIN_BORG_FOR_FEATURE = {
```
ThomasWaldmann: zstd compression comes to mind. present since 1.1.4.

src/vorta/borg/_compatibility.py
```diff
@@ -0,0 +1,24 @@
...
+    'REPOKEY_BLAKE2': parse_version('1.1.4'),
```
ThomasWaldmann: blake2 since 1.1.0.

src/vorta/borg/_compatibility.py
```diff
@@ -0,0 +1,24 @@
...
+    # add new version-checks here.
```
ThomasWaldmann: from our docs/changes.rst:
- pre-1.1.4 potential data corruption issue
- pre-1.0.9 manifest spoofing vulnerability (CVE-2016-10099)
- pre-1.0.9 potential data loss

vorta could warn for < 1.1.4 or < 1.0.9 if you like.

src/vorta/borg/_compatibility.py
ThomasWaldmann: guess the key should be rather named 'BLAKE2' (it applies to repokey and also to keyfile modes).

src/vorta/views/repo_add_dialog.py
```diff
@@ -85,12 +86,19 @@ def run_result(self, result):
...
+            self.encryptionComboBox.addItem(self.tr('Repokey-Blake2 (Recommended, key stored in repository)'),
+                                            'repokey-blake2')
+
         self.encryptionComboBox.addItem(self.tr('Repokey'),
                                         'repokey')
-        self.encryptionComboBox.addItem(self.tr('Keyfile-Blake2 (Key stored in home directory)'),
-                                        'keyfile-blake2')
+
+        if app.borg_compat.check('REPOKEY_BLAKE2'):
+            self.encryptionComboBox.addItem(self.tr('Keyfile-Blake2 (Key stored in home directory)'),
```
ThomasWaldmann: with the key naming fix, (see above) this will read less strange.

m3nu:
Thanks for the input. Made some more changes:

- Check for ZStd feature: This was tricky, as I didn't want to change existing user settings. So I ended up disabling them in the list and giving an error when a backup is created with an old version.
- Renamed the `BLAKE2` feature and went to just disabling the options instead of removing them altogether. And refactored into a list to have less repetitive code.
- Moved the `borg_compat` class instance to `utils`. It's easier accessible there and we already have the keyring instance there, which is similar in its usage.

src/vorta/borg/_compatibility.py
```diff
@@ -1,7 +1,8 @@
 from pkg_resources import parse_version
 
 MIN_BORG_FOR_FEATURE = {
-    'REPOKEY_BLAKE2': parse_version('1.1.4'),
+    'BLAKE2': parse_version('1.1.4'),
```
ThomasWaldmann: 1.1.0

src/vorta/views/repo_tab.py
```diff
@@ -88,6 +92,11 @@ def init_ssh(self):
         for key in keys:
             self.sshComboBox.addItem(f'{key["filename"]} ({key["format"]})', key['filename'])
 
+    def toggle_available_compression(self):
+        use_zstd = borg_compat.check('ZSTD')
+        self.repoCompression.model().item(1).setEnabled(use_zstd)
+        self.repoCompression.model().item(2).setEnabled(use_zstd)
```
ThomasWaldmann: hardcoded 1 and 2 here is a potential future bug.

maybe determine offset somehow or add a comment above the place where the compression menu items are added to update this place in case of changes?

m3nu: I actually considered this, but finding the offset needs an extra statement. 

Since you noticed this as well, I added selecting the offset by value, rather than index.

src/vorta/borg/_compatibility.py
m3nu: Didn't you say there were some issues with blake2 before 1.1.4? So I figured I use the version where it was rock-solid.

src/vorta/views/repo_tab.py
```diff
@@ -94,8 +94,9 @@ def init_ssh(self):
 
     def toggle_available_compression(self):
         use_zstd = borg_compat.check('ZSTD')
-        self.repoCompression.model().item(1).setEnabled(use_zstd)
-        self.repoCompression.model().item(2).setEnabled(use_zstd)
+        for algo in ['zstd,3', 'zstd,8']:
+            ix = self.repoCompression.findData(algo)
+            self.repoCompression.model().item(ix).setEnabled(use_zstd)
```
ThomasWaldmann: nice and solid! :)

PR #254, opened 2019-04-15
Don't delete item from sources if item==None

aquarial:
There's a minor bug in the the sources panel.

Clicking 'Remove' when no folders/files are selected causes vorta to crash because item is `None`


m3nu:
Thanks. Good catch. Merged.

PR #256, opened 2019-04-16
Add option for starting Vorta in foreground.

bsvh:
Just adds a setting to start the UI in the foreground and not in the tray.


m3nu:
Not sure this is needed. What problem does it solve exactly?

Hofer-Julian:
I am not completely opposed to an additional foreground setting.

We already discussed Vorta's behaviour and I suggested that it should always starts in foreground when opened manually and in the background on startup.

You, @m3nu, were not convinced of that idea, but what we could do is:
- always start in the background on startup
- give the option to start in foreground when started manually (which would be default-ON on DEs without system tray)
What do you say about that?

wechsler42:
Hi there,
as "new" user I was a bit puzzled not seeing any sign of vorta alive except in the system tray. Therefore I would second @Hofer-Julian and kindly ask for implementing this feature.

m3nu:
Didn't we want to start it in the foreground if no settings are present, i.e. it's started the first time?

Hofer-Julian:
@m3nu does the one thing exclude the other?
It could start in foreground for people with
- no settings
- foreground option to ON
and we could set the foreground option to ON for people without a system tray.

Hofer-Julian:
I couldn't find an elegant and reliable way to check the availability of the system tray, before the application is executed.

I will therefore stop working on this pull request.

src/vorta/application.py
```diff
@@ -52,7 +52,8 @@ def __init__(self, args_raw, single_app=False):
             self.setStyleSheet(qdarkstyle.load_stylesheet_pyqt5())
 
         args = parse_args()
-        if hasattr(args, 'foreground') and args.foreground:
+        fg_setting = SettingsModel.get(key='foreground').value
+        if hasattr(args, 'foreground') and args.foreground or fg_setting:
```
m3nu: Did you fully test this? There may be an issue with precedence of those logical conditions and it could need some parenthesis.

Hofer-Julian: This is outdated now.

PR #288, opened 2019-05-24
If BORG_PASSCOMMAND is found, unset BORG_PASSPHRASE. Fixes #272

### LINKED ISSUE #272: Respect `BORG_PASSCOMMAND` environment variable
I would like to keep the repository password out of my Keyring and instead enter the password everytime it is required for reading/writing the repository.

Would it be possible to keep `ENV['BORG_PASSWORD']` unset if `ENV['BORG_PASSCOMMAND']` is defined (see: https://github.com/borgbase/vorta/blob/master/src/vorta/borg/borg_thread.py#L58-L61)?

This would offer the opportunity to implement a custom method to fetch the repository password.

m3nu:
I wonder if there should be a mention of this in the UI somewhere.

### DIFF
--- a/src/vorta/borg/borg_thread.py
+++ b/src/vorta/borg/borg_thread.py
@@ -60,6 +60,9 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
         else:
             env['BORG_PASSPHRASE'] = '9999999'  # Set dummy password to avoid prompt.
 
+        if env.get('BORG_PASSCOMMAND', False):
+            env.pop('BORG_PASSPHRASE', None)  # Unset passphrase
+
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
How is does the use case currently work?
Do I have to set the BORG_PASSCOMMAND every time and then launch Vorta from the same session?

m3nu:
> How is does the use case currently work?
> Do I have to set the BORG_PASSCOMMAND every time and then launch Vorta from the same session?

Yeah. Should already exist in the env. There is no setting for it in Vorta.

Hofer-Julian:
Then I don't think it has to be mentioned in the GUI, but at some point in the documentation.

Issue #273, opened 2019-05-02
Build macOS Binaries on oldstable instead of stable to cover more versions


Urheberrecht:
**Describe the bug**
Vorta starts with the icon in the upper right corner of the screen. If trying to open from the icon, the icon disappears and the app seems to crash. Starting Vorta again fails.

**To Reproduce**
See above.

**Desktop (please complete the following information):**
 - OS: MacOS 10.13 SERVER (with all security updates) 

**Additional context**
From "vorta.log":
2019-05-02 14:30:45,390 - vorta.i18n - DEBUG - Loading translation failed for ['C'].
2019-05-02 14:30:45,545 - vorta.borg.borg_thread - INFO - Running command /var/folders/q6/j51_d8t162525yvwq1cy41h80000gn/T/_MEIcw9r1j/bin/borg --version

From "system.log":
May  2 14:30:46 Server-intern com.apple.xpc.launchd[1] (com.apple.xpc.launchd.domain.pid.vorta-darwin.61309): Could not resolve origin of domain. XPC services in this domain's bundle will not be bootstrapped: error = 107: Malformed bundle, taint = missing executable

From "vorta-darwin_2019-05-02-143052_Server-intern.crash":
Process:               vorta-darwin [61309]
Path:                  /Applications/Vorta.app/Contents/MacOS/vorta-darwin
Responsible:           vorta-darwin [61309]
       0x100c7f000 -        0x100c85fff +com.borgbase.client.macos (0.6.19 - 0.6.19) <7B85D4FD-A341-348E-8083-7DC82766F257> /Applications/Vorta.app/Contents/MacOS/vorta-darwin
       0x10aa3f000 -        0x10aa8fff7 +org.sparkle-project.Sparkle (1.21.2 12-ga5b23cdbe - 1.21.2) <7C201192-55FC-350C-BA71-8FC0F15B80D5> /Applications/Vorta.app/Contents/Frameworks/Sparkle.framework/Sparkle

m3nu:
What do you see when you run the binary from the Terminal? Like

```
$ ./Vorta.app/Contents/MacOS/vorta-darwin --foreground
```

I'm packaging it on 10.14, but tested on 10.13. The goal is to support the latest and one version before of macOS.

sbrosnihan:
Crashes for me too in sierra and Mojave. I get this when I run from terminal: 
prompt:  /Applications/Vorta.app/Contents/MacOS/vorta-darwin --foreground

[...]

PR #289, opened 2019-05-24
Ignore Python processes owned by other users when finding mount point

m3nu:
See #273

Tested by running a Python process as root. `$ sudo python`

### DIFF
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -242,8 +242,9 @@ def get_mount_points(repo_url):
                             mount_point = proc.cmdline()[idx + 1]
                             mount_points[archive_name] = mount_point
                         break
-        except psutil.ZombieProcess:
+        except (psutil.ZombieProcess, psutil.AccessDenied):
             # Getting process details may fail (e.g. zombie process on macOS)
+            # or because the process is owned by another user.
             # Also see https://github.com/giampaolo/psutil/issues/783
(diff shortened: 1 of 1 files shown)


Urheberrecht:
I am leaving a comment here, as I suppose it is related to the unknown python process.

While v0.6.20 is now running just fine on a standard macOS 10.13 (as per #273), I still see it crashing reproducibly if macOS Server 5.4 is running in parallel.

So I assume there is still something going on with one or more python processes started by macOS Server.

Not sure if this is worth investigating? But just in case: what info would be useful? 

Also, I am not able to test this against the most recent (and stripped down) version of macOS Server on 10.14.

Again, please bear with me in replying with a couple of days of delay.

Many thanks!

Urheberrecht:
Please ignore my previous post for the time being. I might have mixed up  v0.6.20 and  v0.6.20macos10.13.

I will retest tomorrow and report back.

Best regards.

Urheberrecht:
Tested this morning with the following result:

* vorta-0.6.20 crashes with Server 5.6.1 running on 10.13.16 (as expected)
* vorta-0.6.20-macos1013 still crashes with Server 5.6.1 running on 10.13.16 (probably not as expected)
* vorta-0.6.21pre does not crash anymore! :-)

Thank you very much!

Best regards.

PR #292, opened 2019-05-27
Update appdata and add it to bumpversion.cfg and rename flatpak-name to com.borgbase.Vorta

Hofer-Julian:
Update Appdata.
Additionally we would need to update the "bump-system", so that it updates releases as well.
@m3nu do you think you could implement that?


m3nu:
Maybe add the XML file in https://github.com/borgbase/vorta/blob/master/.bumpversion.cfg, so it will automatically update the version number?

Can the date be ignored? I just want to touch as little as possible by hand on every release. Big source for errors.

Hofer-Julian:
> Can the date be ignored? I just want to touch as little as possible by hand on every release. Big source for errors.

> A release tag can have the properties version, date and timestamp. The date property can have any time in ISO 8601 format as its value and should be present for every release. The timestamp tag contains the release time in the form of a UNIX epoch. This tag should not be used in metainfo files in newly written metadata, but will still be parsed in case it is present. The timestamp property is mainly used in generated distro-metadata. In case both release-time tags are present, the timestamp tag will take precedence over date. 
This sounds to me like it is at least strongly suggested
> 
> Maybe add the XML file in https://github.com/borgbase/vorta/blob/master/.bumpversion.cfg, so it will automatically update the version number?
Is it not possible to include the current date with bumpversion?

m3nu:
> Is it not possible to change the date with bumpversion?

No. Just deals with the date. But I can probably add it to the Makefile. One more command there doesn't matter much.

Hofer-Julian:
> 
> 
> > Is it not possible to change the date with bumpversion?
> 
> No. Just deals with the date. But I can probably add it to the Makefile. One more command there doesn't matter much.

That would be great!
I think we will also need that for the debian package (if you are still pursuing that :) )

Hofer-Julian:
Also: Could you update also the description of every release with the Makefile? If it is too much work, then I simply remove the description completely.

m3nu:
Can you just put a link to the release on Github using the version number. Like `For details see: https://github.com/borgbase/vorta/releases/tag/v0.6.20`

Hofer-Julian:
> 
> 
> Can you just put a link to the release on Github using the version number. Like `For details see: https://github.com/borgbase/vorta/releases/tag/v0.6.20`

Good idea! That way we can use bumpversion

Hofer-Julian:
> Can you just put a link to the release on Github using the version number. Like `For details see: https://github.com/borgbase/vorta/releases/tag/v0.6.20`

Unfortunately it seems like it's not that way:
`style-invalid         : <release> description should be prose and not contain hyperlinks`

So either we implement something to update also the description of every release with the Makefile or we remove the description.

I currently tend towards removing the description, because it also forces additional styleguides (like minimum linelength, ...)

Hofer-Julian:
> Can you just put a link to the release on Github using the version number. Like `For details see: https://github.com/borgbase/vorta/releases/tag/v0.6.20`

Unfortunately it seems like it's not that way:
`style-invalid         : <release> description should be prose and not contain hyperlinks`

So either we implement something to update also the description of every release with the Makefile or we remove the description.

I currently tend towards removing the description, because it also forces additional styleguides (like minimum linelength, ...).

@m3nu If you agree, I would suggest that we squash and merge this pull request, you add the appdata to bumpversion and to the Makefile for the date change and then we bump to a new release, which will hopefully be the first one to be published on flathub.

nedrichards:
I'd remove the description if I were you.

Hofer-Julian:
> I'd remove the description if I were you.

Then I'll do that.

Hofer-Julian:
@m3nu From my side, unless you want to push changes to the Makefile (for the date change) to this pull request, we can squash and merge it now.

m3nu:
Good job. I'll look over it tonight and then it can be merged and published? Is a new version needed?

Hofer-Julian:
> Good job. I'll look over it tonight and then it can be merged and published? Is a new version needed?

Yes, a new version will be needed, because flathub only accepts [stable releases](https://github.com/flathub/flathub/wiki/App-Requirements#stable-releases-reproducible-builds) as sources.

m3nu:
I see some settings should be changed on the Flatpak thread. Is this just minor stuff or any blockers?

Hofer-Julian:
I am not sure yet, but it probably will not have to be changed here already - only in the flathub repository.

m3nu:
OK. Just started merging everything for the next release. Still need to work on some graphical issues on macOS. So let's make sure Flatpak is also ready for this version.

Hofer-Julian:
I do not expect flathub's CI to complain about our next release, but at the same time I also cannot really test it beforehand...

nedrichards:
You could test a git snapshot when this is all merged?

Hofer-Julian:
@nedrichards I didn't know this was possible. Thank you!

Hofer-Julian:
@m3nu as soon as this pull request is merged, I will test it and give you the okay to release it :blush:

PR #307, opened 2019-06-15
Force fusion style on Linux

### LINKED ISSUE #298: Problem with QToolbox
Today I tried to find the root of the problem with Vorta's Toolbox (no Text is displayed).
QDesigner gave me an idea, but in the end I couldn't fix it.
Vorta-Adwaita-Problem

Hofer-Julian:
Fixes #298.

I mean the optimal case would be if the Adwaita-like style would work, but it is better than nothing.

### DIFF
--- a/src/vorta/__main__.py
+++ b/src/vorta/__main__.py
@@ -34,16 +34,15 @@ def main():
     # Init database
     sqlite_db = peewee.SqliteDatabase(os.path.join(SETTINGS_DIR, 'settings.db'))
     init_db(sqlite_db)
-
-    # Force fusion style on Linux
-    sys_argv = sys.argv
-    if sys.platform.startswith('linux'):
-        sys_argv += ['--style=fusion']
-
+    
     # Init app after database is available
     from vorta.application import VortaApp
-    app = VortaApp(sys_argv, single_app=True)
+    app = VortaApp(sys.argv, single_app=True)
     app.updater = get_updater()
+
+    # Force fusion style on Linux
+    if sys.platform.startswith('linux'):
+        app.setStyle('Fusion')
 
     sys.exit(app.exec_())
 
--- a/src/vorta/__main__.py
+++ b/src/vorta/__main__.py
@@ -34,7 +34,7 @@ def main():
     # Init database
     sqlite_db = peewee.SqliteDatabase(os.path.join(SETTINGS_DIR, 'settings.db'))
     init_db(sqlite_db)
-    
+
     # Init app after database is available
     from vorta.application import VortaApp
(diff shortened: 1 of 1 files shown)


m3nu:
Would this do the same?

```
app = QApplication([])  # wherever we init the app
app.setStyle('Fusion')
```

Hofer-Julian:
> Would this do the same?
> 
> ```
> app = QApplication([])  # wherever we init the app
> app.setStyle('Fusion')
> ```

Yes,  I changed it to your suggestion.

geckolinux:
Hi, would it  be possible to make this forced Fusion theme optional? I use the last Plasma / KDE libs on openSUSE Tumbleweed, and Qt5 apps always integrate just fine with my Qt and Plasma configuration. Thanks!

Hofer-Julian:
What exactly is your problem with the fusion-style?

geckolinux:
Just the fact that it doesn't fit in with my configured widget scheme. Was there a specific reason for hard-coding Fusion into Vorta? It's true that KDE and Qt apps have a lot of variables and most people have trouble getting a consistent look. But it _can_ be done with a bit of effort, and usually those who have gone through the trouble of doing so don't really like it when apps don't respect their system configuration.

Here's an example of a Qt5 app that respects my KDE and Qt themes, alongside a native KDE app:
Screenshot_20200118_231811

m3nu:
We used the fusion style because it avoided some UI issues. If those issues are gone in newer Qt versions, then a PR to change that will be welcome.

geckolinux:
I tried to force the theme using the `QT_QPA_PLATFORMTHEME` variable, but it doesn't fully change it. Is there any other way to change it at run time to test?

m3nu:
The best way to test is probably to comment out that part in the [code](https://github.com/borgbase/vorta/blob/329f096fb4bae8552728ee076b514eb8c43258eb/src/vorta/qt_single_application.py#L21) and then run it from the Git repo following our [dev-instructions](https://vorta.borgbase.com/contributing/).

(Or quick and dirty: edit the file in `site-packages` of the Python installation)

geckolinux:
> (Or quick and dirty: edit the file in site-packages of the Python installation)

Ah yes, good idea. Seems to work just fine with Breeze widgets and my custom color scheme:

Screenshot_20200119_122655

Hofer-Julian:
We currently force fusion style, because there are problems on GNOME based distributions.

geckolinux:
I'll try to give it a shot on another distro. Would it be best to test on Gnome proper, or Cinnamon too?

Hofer-Julian:
GNOME should be enough, thanks

Issue #277, opened 2019-05-06
System Tray not working correctly with Xfce (Xubuntu 18.04)


wechsler42:
Hi there,
first of all thanks for offering a GUI for BorgBackup :-)

**Describe the bug**
The system tray is dysfunctional, i.e. nothing happens on left click and on right click on the system tray only  the "Status Notifier Plugin" pull down menu (see attachment) is visible.
![menu_small](https://user-images.githubusercontent.com/19371424/57248029-a41c1600-7041-11e9-913d-c967262f024d.png)

**To Reproduce**
1. Go to system tray
2. Left Click on system tray --> Nothing happens
3. Right click on system tray --> Status Notifier Plugin show up
4. No error message appears in the terminal

**Desktop (please complete the following information):**
 - OS:  Xubuntu 18.04.2 LTS (Bionic)
 - Vorta : 0.6.19 installed via "pip3 install vorta" 

**Additional context**
Log file:
2019-05-06 20:33:49,455 - vorta.i18n - DEBUG - Loading translation failed for ['en', 'en-US', 'en-Latn-US'].
2019-05-06 20:33:49,523 - apscheduler.scheduler - INFO - Scheduler started
2019-05-06 20:33:50,727 - vorta.borg.borg_thread - INFO - Running command /usr/bin/borg --version
2019-05-06 20:38:18,035 - vorta.i18n - DEBUG - Loading translation failed for ['en', 'en-US', 'en-Latn-US'].
2019-05-06 20:38:18,037 - apscheduler.scheduler - INFO - Scheduler started
2019-05-06 20:38:18,585 - vorta.borg.borg_thread - INFO - Running command /usr/bin/borg --version

m3nu:
Hi, I don't think this is well-tested on Xfce. Mostly we have macOS and Gnome users.

If you can submit a PR to improve the behavior on Xfce, that would be great. Someone already added a workaround for KDE. Something similar could be needed for Xfce.

cbruno1:
I had the same problem, I solved it by disabling the "The menu is a main action" option in the Status Notifier Plugin settings.

m3nu:
Thanks for the tip, @cbruno1 . We should add this to the docs somehow.

PR #310, opened 2019-06-21
Update README to change installation recommendations.

Hofer-Julian:
Highlight macOS app bundles and Flatpak.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,7 +12,7 @@ $ brew cask install vorta
 
 <a href='https://flathub.org/apps/details/com.borgbase.Vorta'><img width='240' alt='Download on Flathub' src='https://flathub.org/assets/badges/flathub-badge-en.png'/></a>
 
-Settings can be transfered by copying ~/.local/share/Vorta/settings.db to ~/.var/app/com.borgbase.Vorta/data/Vorta/.
+Settings can be transferred by copying ~/.local/share/Vorta/settings.db to ~/.var/app/com.borgbase.Vorta/data/Vorta/.
 
 ### Distribution Packages
 If you added Vorta to a Linux distribution, open an issue or PR and it will be added here.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,6 +12,8 @@ $ brew cask install vorta
 
 <a href='https://flathub.org/apps/details/com.borgbase.Vorta'><img width='240' alt='Download on Flathub' src='https://flathub.org/assets/badges/flathub-badge-en.png'/></a>
 
+Settings can be transfered by copying ~/.local/share/Vorta/settings.db to ~/.var/app/com.borgbase.Vorta/data/Vorta/.
+
 ### Distribution Packages
 If you added Vorta to a Linux distribution, open an issue or PR and it will be added here.
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -28,7 +28,7 @@ Unless Vorta is started with the `--foreground` option, it will minimize to the 
 
 
 ## Install from Source
-The generic way is to install it as Python package using [PIP](https://pip.readthedocs.io/en/stable/installing/). First [install](https://borgbackup.readthedocs.io/en/stable/installation.html) Borg using the package of your distribution or via PyPI. The latter needs some additional [source packages](https://borgbackup.readthedocs.io/en/stable/installation.html#dependencies). Then install Vorta from PyPI. Your local Python version should be >= 3.6.
+The generic way is to install it as Python package using [PIP](https://pip.readthedocs.io/en/stable/installing/). First [install](https://borgbackup.readthedocs.io/en/stable/installation.html) Borg using the package of your distribution or via PyPI. The latter needs some additional [source packages](https://borgbackup.readthedocs.io/en/stable/installation.html#dependencies). Then install Vorta from PyPI. Your local Python version must be >= 3.6.
 ```
 $ pip3 install vorta
 ```
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -30,7 +30,7 @@ Unless Vorta is started with the `--foreground` option, it will minimize to the 
 
 
 ## Install from Source
-The generic way is to install it as Python package using [PIP](https://pip.readthedocs.io/en/stable/installing/). First [install](https://borgbackup.readthedocs.io/en/stable/installation.html) Borg using the package of your distribution or via PyPI. The latter needs some additional [source packages](https://borgbackup.readthedocs.io/en/stable/installation.html#dependencies). Then install Vorta from PyPI. Your local Python version must be >= 3.6.
+The generic way is to install it as Python package using [pip](https://pip.readthedocs.io/en/stable/installing/). First [install](https://borgbackup.readthedocs.io/en/stable/installation.html) Borg using the package of your distribution or via pip. The latter needs some additional [source packages](https://borgbackup.readthedocs.io/en/stable/installation.html#dependencies). Then install Vorta via pip. Your local Python version must be >= 3.6.
 ```
 $ pip3 install vorta
 ```
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,5 +1,31 @@
 # Installing
 
+## Apple macOS
+Install via [Homebrew Cask](https://brew.sh/) or download a pre-built app bundle for macOS (10.13+) from [Releases](https://github.com/borgbase/vorta/releases).
+```
+$ brew cask install vorta
+```
+
+## Linux
+
+### Flatpak
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,9 +14,16 @@ Vorta is a backup client for macOS and Linux desktops. It integrates the mighty
...
 
-For available packages and instructions, see [INSTALL.md](https://github.com/borgbase/vorta/blob/master/INSTALL.md)
+Pre-built app bundle for macOS (10.13+) can be downloaded from [Releases](https://github.com/borgbase/vorta/releases).
+
+On Linux we recommmend to install Vorta via Flathub
+
+<a href='https://flathub.org/apps/details/com.borgbase.Vorta'><img width='240' alt='Download on Flathub' src='https://flathub.org/assets/badges/flathub-badge-en.png'/></a>
+
+
+Alternative packages and further instructions, see [INSTALL.md](https://github.com/borgbase/vorta/blob/master/INSTALL.md)
```
ThomasWaldmann: there are either some `.` or `:` missing. also try to write complete sentences.b

README.md
```diff
@@ -14,9 +14,16 @@ Vorta is a backup client for macOS and Linux desktops. It integrates the mighty
 
 
 ## Installation
-Vorta should work on all platforms that support Qt and Borg. This includes macOS, Ubuntu, Debian, Fedora, Arch Linux and many others. Windows is currently not supported by Borg, but this may change in the future.
+Vorta should work on macOS and all Linux distributions that [support Flatpak](https://flatpak.org/setup/). Windows is currently not supported by Borg, but this may change in the future.
```
m3nu: Flatpak is not available on all Linux distros out of the box. On Ubuntu I had to install some extra KDE Flatpaks first. And Flatpak itself of course.

The lightest and most universal install option is still Pip. So "should work on [...] all Linux distributions that support Flatpak" is misleading, because it's not strictly required at all.

There is also no reason to split the install instructions up like this. They already have their own file.

Hofer-Julian: I am quite surprised by your response. So let me bring up some counter arguments.

> Flatpak is not available on all Linux distros out of the box. 

So is pip, but for flatpak we provide clear installation instructions compared to pip where we leave the user alone to find out they have to execute e.g. sudo apt install python3-pip beforehand.

`On Ubuntu I had to install some extra KDE Flatpaks first`

Yes, but that should happen automatically.

`The lightest and most universal install option is still Pip.`

Lightest yes, most universal? Also true, but is that so important when [this](https://flatpak.org/setup/) includes pretty much every commonly used Linux distribution?

> So "should work on [...] all Linux distributions that support Flatpak" is misleading, because it's not strictly required at all.

I agree with that one, haven't thought of that.

> 
There is also no reason to split the install instructions up like this. They already have their own file.


Yeah... But I would still argue that we should recommend certain kind of installation methods. One way or the other. I just want to note hear that after two years of using python I still had to search the internet just to install Vorta.
On Ubuntu you have to:

- find out how to install pip of python 3
- use pip3
- live with a version that doesn't update itself and has no icon

On Fedora:

- you have pip but get a mysterious error message when you try to install
- okay the internet says I have to install python3-devel
- now people might get an error again
- they have to think about using pip3 not pip and don't use sudo, but use the --user flag instead

And just think how many pip-related issues we had by now.

README.md
Hofer-Julian: Thanks for the tips :P
I will change it in the next commit.

README.md
Hofer-Julian: I now pushed an alternative to this pull request,

m3nu: I like the new version, but let's still keep it open for a while and see if all makes sense.

One reason why I didn't want to emphasize Flatpak too much is because it's a new feature. If some people find it, great. If there are any problems and too many find it, bad. Better keep it low for a while and emphasize it more over time.

I'm aware that Flatpak solves many common install issues (possibly the new #312?) and that it can reach a wider audience of non-tech people than `pip`. It could become our main distribution channel for Linux. We just need to be aware of the complexity involved and there will be improvements over time. E.g. I didn't get menu icons in after installing it on XFCE.

Once we trust it sufficiently, we can emphasize it more on the Github frontpage. Flathub will also help to discover new users to find Vorta. Especially the large group that doesn't install stuff from Github directly. (I actually made a landing page + static docs for Vorta some months ago, but never finished it. Mostly because I wasn't sure if we have the ease-of-use level required for a less technical audience.)

Hofer-Julian: > One reason why I didn't want to emphasize Flatpak too much is because it's a new feature. If some people find it, great. If there are any problems and too many find it, bad. Better keep it low for a while and emphasize it more over time.

That is a very good point. I agree, let's keep it low for a while.

> E.g. I didn't get menu icons in after installing it on XFCE.

Is it the same problem as #277? Because that one was not flatpak-specific.

INSTALL.md
```diff
@@ -28,7 +28,7 @@ Unless Vorta is started with the `--foreground` option, it will minimize to the
 
 
 ## Install from Source
-The generic way is to install it as Python package using [PIP](https://pip.readthedocs.io/en/stable/installing/). First [install](https://borgbackup.readthedocs.io/en/stable/installation.html) Borg using the package of your distribution or via PyPI. The latter needs some additional [source packages](https://borgbackup.readthedocs.io/en/stable/installation.html#dependencies). Then install Vorta from PyPI. Your local Python version should be >= 3.6.
+The generic way is to install it as Python package using [PIP](https://pip.readthedocs.io/en/stable/installing/). First [install](https://borgbackup.readthedocs.io/en/stable/installation.html) Borg using the package of your distribution or via PyPI. The latter needs some additional [source packages](https://borgbackup.readthedocs.io/en/stable/installation.html#dependencies). Then install Vorta from PyPI. Your local Python version must be >= 3.6.
```
ThomasWaldmann: The first sentence talk about "using PIP" (the command is lowercase, so "pip" might be less confusing).

The second sentence talks about "via PyPI" and means the same.

Guess both should say either "via pip" or "via PyPI" if they mean the same method.

Hofer-Julian: Good point.
Please just change the file, "allow edits from maintainers" is enabled ;-)

INSTALL.md
```diff
@@ -1,25 +1,37 @@
...
 
-## Install from Source
-The generic way is to install it as Python package using [PIP](https://pip.readthedocs.io/en/stable/installing/). First [install](https://borgbackup.readthedocs.io/en/stable/installation.html) Borg using the package of your distribution or via PyPI. The latter needs some additional [source packages](https://borgbackup.readthedocs.io/en/stable/installation.html#dependencies). Then install Vorta from PyPI. Your local Python version should be >= 3.6.
+## Linux
+
+### Flatpak
+
+<a href='https://flathub.org/apps/details/com.borgbase.Vorta'><img width='240' alt='Download on Flathub' src='https://flathub.org/assets/badges/flathub-badge-en.png'/></a>
+
+Settings can be transfered by copying ~/.local/share/Vorta/settings.db to ~/.var/app/com.borgbase.Vorta/data/Vorta/.
```
ThomasWaldmann: transferred

ThomasWaldmann: Also, the standard way should be the distribution package and the flatpak should only be the second best option.

For dist packages, you automatically get security updates and also they are usually cryptographically signed. Also they are a natural part of your linux installation, like all the other packages, not something extra / different.

flatpak could be recommended if a newer version is desired than provided by dist packages.

INSTALL.md
```diff
@@ -1,25 +1,37 @@
...
+
+### Distribution Packages
+If you added Vorta to a Linux distribution, open an issue or PR and it will be added here.
```
ThomasWaldmann: this should not be the first sentence here, as it addresses only a very small part of the readers (who happen to be vorta package maintainers).

see above for some more suitable content.

INSTALL.md
Hofer-Julian: > Also, the standard way should be the distribution package and the flatpak should only be the second best option.
> 
> For dist packages, you automatically get security updates and also they are usually cryptographically signed. Also they are a natural part of your linux installation, like all the other packages, not something extra / different.
> 
> flatpak could be recommended if a newer version is desired than provided by dist packages.

While I don't agree that Flatpak should come after distribution packages, I think we can postpone this to the time the first distribution package is published :)

Hofer-Julian: > transferred

Changed :)

INSTALL.md
Hofer-Julian: Done

INSTALL.md
Hofer-Julian: What do you mean with "see above"?

ThomasWaldmann: https://github.com/borgbase/vorta/pull/310#discussion_r302552864

m3nu:
This information is now here: https://github.com/borgbase/vorta.borgbase.com

PR #313, opened 2019-06-22
Implement borg diff

Hofer-Julian:
Ongoing work - do not merge yet.


Hofer-Julian:
It now works for me.
But further testing, maybe even write a test would probably be good.

It can take quite a long time for repositories > 1 TB (>10min), maybe we could provide some kind of feedback so users don't think it freezes.

Hofer-Julian:
When you cancel the diff-process, all the buttons stay disabled.
That needs to be fixes.

Hofer-Julian:
As people, not excluding myself, get anxious when nothing happens for longer than 30 sec, I would like to put the newest output of borg diff here:

Screenshot-4-Archives

That might also be reasonable for other functions like extract.

@m3nu do you think it would be possible (and make sense) to modify borg thread in a way that it shows this behaviour?

m3nu:
I'm actually looking to remove the status text there and only show status updates at the bottom, where most updates appear. Or you think there should be 2 status texts?

Hofer-Julian:
> 
> 
> I'm actually looking to remove the status text there and only show status updates at the bottom, where most updates appear. Or you think there should be 2 status texts?

One status label would be fine as well, but where exactly do you mean?
Right of the cancel button?

Hofer-Julian:
Problems with the cancel button are not diff-specific and should probably be handled separately.
Redesigning the status text should probably also be implemented separately.

I tested it now for quite some time on Fedora 30 with GNOME.
If someone wants to test it as well, great!
Otherwise I would like it merge it soon.

PR #325, opened 2019-07-08
Adding Matrix

Hofer-Julian:
I would like to propose to add a Vorta-specific chat rooms.
Benefits should be clear: Discussing stuff which does not belong in a chat room, support, attract beta testers,...

Why Matrix?

It seems like the most popular chat protocol for open source projects is IRC, but it is not beginner-friendly and you have to be online to receive messages.
Matrix is as easy to use as modern messengers (similar to Slack), supports all operating systems (including iOS and Android), it offers a web client and if there are too many people who miss IRC: Matrix also offers an IRC Bridge. 

I already created a room for this purpose.

I noticed that especially GNOME projects are moving to Matrix (https://wiki.gnome.org/Initiatives/Matrix), I am not sure about other projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ For available packages and instructions, see [INSTALL.md](https://github.com/bor
 
 ## Get in touch
 - If you have questions or simply want to talk about Vorta hit us up at [Matrix](https://matrix.to/#/#vorta:matrix.org)
-- If your questions are Borg-specific it might be advisable to join the #borgbackup IRC channel on chat.freenode.net instead (Note: borg currently requires registered nicknames). Matrix is very suitable to be used as an always-on IRC-client as explained in this [tutorial](https://gist.github.com/fstab/ce805d3001600ac147b79d413668770d).
+- If your questions are Borg-specific it might be advisable to join the #borgbackup IRC channel on chat.freenode.net instead. Matrix is very suitable to be used as an always-on IRC-client, simply [register/identify to NickServ](https://github.com/matrix-org/matrix-appservice-irc/wiki/End-user-FAQ#how-do-i-registeridentify-to-nickserv) and then join the room `#freenode_#borgbackup:matrix.org`.
 
 ## License and Credits
 - Thank you to all the people who already contributed to Vorta: [code](https://github.com/borgbase/vorta/graphs/contributors), [translations](https://github.com/borgbase/vorta/issues/159)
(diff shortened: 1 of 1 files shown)


m3nu:
I see more people and projects use Telegram. But if this gets used, it's fine too. Suggest to evaluate for 2 months and remove it, if it's not used much?

Hofer-Julian:
Never thought about removing it if it wouldn't get used enough.
Why would we do that?

Hofer-Julian:
> 
> 
> I see more people and projects use Telegram. But if this gets used, it's fine too. Suggest to evaluate for 2 months and remove it, if it's not used much?

I also thought about Telegram, but I personally only no one project which uses it.
Not 100% sure which option is better.

ThomasWaldmann:
IMHO: no telegram, its crypto seems to be problematic.

Hofer-Julian:
> IMHO: no telegram, its crypto seems to be problematic.

You are the second person noting this.
I am curious. Is it because you don't want to motivate people to use the (potentially insecure) Telegram messenger for other things as well?
Because for an open discussion channel, encryption doesn't matter right?
That is why I even planned to leave the Matrix channel unencrypted.

ThomasWaldmann:
Well, if you are happy using a product that advertises potentially insecure crypto, then you can happily use it and advertise it to other people also, but I rather won't. But I guess this is a potential footgun if people are unaware and might rely on its crypto for some purpose.

Hofer-Julian:
> Well, if you are happy using a product that advertises potentially insecure crypto, then you can happily use it and advertise it to other people also, but I rather won't. I guess this is a potential footgun if users are unaware and might rely on its crypto for some purpose.

Yeah, I see your point.
It is one of the reasons why I prefer Matrix + the fact that is decentralised and has bridges for practically everything.

Would you, @m3nu and @ThomasWaldmann participate in the Matrix channel?
And would that participation depend whether the IRC bridge is activated or not?

m3nu:
> Would you, @m3nu and @ThomasWaldmann participate in the Matrix channel?
> And would that participation depend whether the IRC bridge is activated or not?

There is already an IRC [channel](https://www.borgbackup.org/support/free.html) for Borg. Vorta alone could be too small for its own channel and many questions will be Borg-related anyways.

Depending on the "trend" in messengers, the best way could be to link to the Borg IRC or replace that with Matrix (+Riot as client?). I'm good to join either place, as long as there is no needless fragmentation.

Hofer-Julian:
The only thing I don't like about that idea is that I currently mute the borg IRC channel. Not enough participation is not a bad thing in my opinion because then I have the nerves to actually read every message.

However, as you mentioned, Riot is an amazing always-on IRC client. We could additionally link a tutorial how to join the borg IRC channel with Vorta.

Hofer-Julian:
I now mentioned Borg's IRC channel.

PR #356, opened 2019-09-01
Bugfix for handling 0 byte modification in diff view

philroche:
I encountered an exception being thrown when viewing diff of two archives.

The exception was
'''
vorta/views/diff_result.py", line 52, in parse_line
    size = int(size)
ValueError: invalid literal for int() with base 10: ''
'''

which was caused by diff line:

'''
0 B   -2.6 kB home/philroche/.config/google-chrome/Profile 1/QuotaManager-journal
'''

This commit fixes the issue by only stripping +/- from string if they exist.

### DIFF
--- a/src/vorta/views/diff_result.py
+++ b/src/vorta/views/diff_result.py
@@ -44,7 +44,7 @@ class DiffResult(DiffResultBase, DiffResultUI):
                 unit = line_splitted[1]
                 # If present remove '+' or '-' sign at the front
                 if '+' in size or '-' in size:
-                    size = size[1:]   
+                    size = size[1:]
 
             if line_splitted[0].startswith("["):
                 size = 0
--- a/src/vorta/views/diff_result.py
+++ b/src/vorta/views/diff_result.py
@@ -40,9 +40,11 @@ class DiffResult(DiffResultBase, DiffResultUI):
                 unit = line_splitted[2]
             else:
                 change_type = "modified"
-                # Remove '+' or '-' sign at the front
-                size = line_splitted[0][1:]
+                size = line_splitted[0]
                 unit = line_splitted[1]
+                # If present remove '+' or '-' sign at the front
+                if '+' in size or '-' in size:
+                    size = size[1:]   
 
             if line_splitted[0].startswith("["):
                 size = 0


Hofer-Julian:
Looks good to me.

Will merge it after I tested it.

Thank you for your contribution!

src/vorta/views/diff_result.py
```diff
@@ -40,9 +40,11 @@ def parse_line(line):
                 unit = line_splitted[2]
             else:
                 change_type = "modified"
-                # Remove '+' or '-' sign at the front
-                size = line_splitted[0][1:]
+                size = line_splitted[0]
                 unit = line_splitted[1]
+                # If present remove '+' or '-' sign at the front
+                if '+' in size or '-' in size:
+                    size = size[1:]
```
ThomasWaldmann: well, guess "it works", but what you really meant is:
```
if size[0] in ('+', '-'):
    size = size[1:]
```

src/vorta/views/diff_result.py
```diff
@@ -40,9 +40,11 @@ def parse_line(line):
                 unit = line_splitted[2]
             else:
                 change_type = "modified"
-                # Remove '+' or '-' sign at the front
-                size = line_splitted[0][1:]
+                size = line_splitted[0]
```
ThomasWaldmann: also, it is `split`, not `splitted`.

philroche:
@ThomasWaldmann Thanks for your review but this has been merged so I suspect if you would like these personal preference/wording changes a new PR will be required.

ThomasWaldmann:
sure.

m3nu:
I do feel that we have been merging too quickly in the past and the feedback by @ThomasWaldmann is quite appropriate.

To improve this, I've enabled "required reviews" for new PRs.

philroche:
I'm hesitant to respond as I'm a new contributor but are these PR comments really worthy of debate?

* ` if '+' in size or '-' in size:` comment: Both methods do the same thing and are equally as readable and understandable. As such it's personal preference and personal preference should never block a PR - unless it's is in your style guide. I did check https://vorta.borgbase.com/contributing/ to see if you had a style guide before I submitted my PR.

* line_splitted variable comment: This is a variable used already in this code. To block this PR for using a variable that already exists seems bizarre.

ThomasWaldmann:
```
>>> a = '-1.0'
>>> b = 'foo-bar'

>>> '-' in a or '+' in a
True
>>> '-' in b or '+' in b
True

>>> a[0] in ('+', '-')
True
>>> b[0] in ('+', '-')
False

>>> a[1:]
'1.0'
>>> b[1:]
'oo-bar'
```
So you see, this is not about personal preference, but about correctness for the general case.

As I already said, you were lucky that you had data that always conform to a special case, so "it worked". But it is still not the correct way to do that.

`splitted`: ok, if it already was wrong before, then it was out of scope of that commit to fix that.

We can still fix it, correct English is always nice to read.

m3nu:
Related: https://blog.codinghorror.com/the-broken-window-theory/

philroche:
@ThomasWaldmann OK, I concede the point about `if '+' in size or '-' in size`. Well made and worthy of changing but from what I have seen the output of `borg diff` will never contain any string like that. That is not to say that the output might change in the future. 

PR submitted to fix this https://github.com/borgbase/vorta/pull/359

Issue #349, opened 2019-08-18
Allow sorting Archives by any of the columns


archont00:
**Is your feature request related to a problem? Please describe.**
I keep a single repository (external HDD) for all computer archives. Hence, each archive name looks like `computer_name-date_stamp`.
Vorta seems to sort Archives only by date. Hence reviewing number of backups by computer name is difficult with a long history of backups.

**Describe the solution you'd like**
It would be nice to allow user to click on a heading of the columns (Date, Size, ..., Name) and have the list sorted up/down.

m3nu:
Merged and will be in the next release.

PR #361, opened 2019-10-02
Enable sorting of archive table columns (#349)

Plastix:
This enables column sorting for the table on the archive list page.
However, all columns are sorted by lexical string order. For example,
this means that archives with "52 MB" are ranked greater than "332 MB".

I wasn't able to figure out how to implement custom sorting behaviors. I did a fair amount of research and it doesn't look like PyQt supports this feature.

Completes: https://github.com/borgbase/vorta/issues/349


m3nu:
Cool. Very simple.

PR #363, opened 2019-10-07
Always mount with current user and read-only perms.

### LINKED ISSUE #362: Override permissions for `borg mount`
**Problem**
I sometimes restore backups of remote servers. Most files there would belong to other users, like root. Using default mount options, `borg mount` will use the original permissions, thus denying access to normal users.

**Solution**
Since Vorta targets desktops, I suggest to always run `borg mount` with the following options to give the current user read access:

`borg mount -o umask=0277,uid=$UID xxx@xxx.repo.borgbase.com:repo::2019-04-30-1556602978 mnt`

Not aware of alternatives. The above solution only partly solves the problem. Backups that use ACL (e.g. from Synology NAS) still need to be mounted as root and won't work as normal user on macOS, even with permissions set as above. It works fine on Linux though.

Any opinion on this or issues you experienced, @Hofer-Julian @ThomasWaldmann

m3nu:
Fixes #362

### DIFF
--- a/src/vorta/borg/mount.py
+++ b/src/vorta/borg/mount.py
@@ -1,3 +1,4 @@
+import os
 from .borg_thread import BorgThread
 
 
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Mounting works as expected!

m3nu:
Thanks for checking. Good time for a release?

Issue #316, opened 2019-06-23
Include ssh-askpass in Flatpak build


jans23:
When using Vorta installed with pip, ssh-askpass prompts for the SSH password. When using Vorta being installed from Flatpak, it doesn't ask for SSH password but instead remote access fails.

System: Ubuntu Linux 18.04, Vorta 0.6.21.

Hofer-Julian:
https://github.com/KDE/ksshaskpass and https://github.com/lxqt/lxqt-openssh-askpass might be alternatives

jans23:
How about https://gnupg.org/related_software/pinentry/ ?

prefixtry:
@Hofer-Julian Is there a solution you favor yet?
OpenSSH has https://man.openbsd.org/ssh-add
How GNOME's Keyring does it https://wiki.gnome.org/Projects/GnomeKeyring/Ssh

prefixtry:
@Hofer-Julian What requirements do you see for solving this issue?

Hofer-Julian:
A program similar to ssh-askpass, but using wayland instead of x11, but in the ideal case is able to use x11 as a fallback.

Hofer-Julian:
But I guess it is not a strict requirement...

prefixtry:
you mean a fallback mode?

jans23:
This issue is still relevant for me. Any idea how to make Vorta ask for SSH passphrase?

Issue #318, opened 2019-06-25
Tray icon missing on Ubuntu Budgie 19.04


drohm:
When installing the flatpak version, after starting Vorta, it shows up in the tray but the icon is missing:

![vorta-missing-tray-icon](https://user-images.githubusercontent.com/184959/60107281-a262fa80-9734-11e9-85b0-12a790dd740e.png)

**To Reproduce**
1. Install flatpak and launch vorta.

**Desktop (please complete the following information):**
 - OS: Ubuntu Budgie 19.04

Hofer-Julian:
Could please install Vorta via `pip3 install --user vorta` (you might have to execute `sudo apt install python3-pip` first) and find out if the icon is there missing too?

drohm:
@Hofer-Julian On my previous operating install (I had to reinstall yesterday) I used the pip3 installation and the tray icon showed up perfectly.  On this OS install I tried the flatpak and the icon is missing.

Hofer-Julian:
Okay!

Please run `flatpak run com.borgbase.Vorta --foreground` and post the output of the terminal here.

drohm:
Here's the output:
```
Gtk-Message: 11:47:48.582: Failed to load module "canberra-gtk-module"
Gtk-Message: 11:47:48.583: Failed to load module "canberra-gtk-module"
2019-06-25 11:47:48,603 - vorta.i18n - DEBUG - Loading translation failed for ['en', 'en-US', 'en-Latn-US'].
2019-06-25 11:47:48,606 - apscheduler.scheduler - INFO - Scheduler started
2019-06-25 11:47:48,608 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2019-06-25 11:47:48,608 - vorta.scheduler - DEBUG - New job for profile EzyRhinox was added.
libGL error: No matching fbConfigs or visuals found
libGL error: failed to load driver: swrast
2019-06-25 11:47:49,236 - vorta.borg.borg_thread - INFO - Running command /app/bin/borg --version
```

Hofer-Julian:
I didn't see the error message I expected, but please still try to run:
`flatpak run --own-name='org.kde.*' com.borgbase.Vorta`

drohm:
The only output from that was:
```
Gtk-Message: 12:42:12.224: Failed to load module "canberra-gtk-module"
Gtk-Message: 12:42:12.224: Failed to load module "canberra-gtk-module"
```

[...]

Issue #315, opened 2019-06-23
Flatpak should be added to INSTALL.md


jans23:
- The option to install Vorta via Flatpak should be added to INSTALL.md
- When installing Vorta via pip, it doesn't install ssh-askpass automatically. Either this dependency should be installed automatically or it should be mentioned in INSTALL.md.

Hofer-Julian:
See https://github.com/borgbase/vorta/pull/310 for point 1

PR #370, opened 2019-10-19
Improve flatpak

Hofer-Julian:
This PR updates the flatpak dependencies and some refactoring.
It will allows to build Vorta with GNOME Builder with only one click.
The next step will be to adjust the contribution docs.


Hofer-Julian:
From my side, this PR is ready to be merged.

prefixtry:
Thank you for working on an improvement! Does this improvement even include a solution for the following Flatpak related issues https://github.com/borgbase/vorta/issues/316 and https://github.com/borgbase/vorta/issues/315 or https://github.com/borgbase/vorta/issues/318? Or could you imagine integrating solutions into your PR?

Hofer-Julian:
I need a Wayland program in order to close https://github.com/borgbase/vorta/issues/316.
You are very welcome to suggest one in https://github.com/borgbase/vorta/issues/316.

https://github.com/borgbase/vorta/issues/315 is done.

For https://github.com/borgbase/vorta/issues/318 you are very welcome to try (e.g, on a virtual machine) if the problem is still persists and report it to https://github.com/borgbase/vorta/issues/318)

m3nu:
I'll need more time on this. If something crashes right when starting it, there could be more issues. So not merging it.

m3nu:
And do you have someone who can sign off on the Flatpak changes? I can't review those.

m3nu:
`os.fork()` crashes macOS due to https://bugs.python.org/issue33725

```
  File "/Users/manu/.pyenv/versions/3.6.7/envs/vorta/lib/python3.6/site-packages/peewee.py", line 3005, in execute_sql
    cursor.execute(sql, params or ())
sqlite3.OperationalError: disk I/O error
```

m3nu:
So here is the issue with your changes:

- You added a new setting in the DB `foreground` and use it, without setting up the database. This will fail in all cases. I don't know how this ended up in there.
- If you init the DB *before* forking, the new process won't be able to use the DB. This is how it is on macOS. I didn't test Linux.

Options to solve this:

1. Remove the new setting `foreground`.
2. Init the database 2x: First for the original process, then for the fork. This works, but not a fan.
3. Remove the background feature and always run in foreground.

Hofer-Julian:
@m3nu i can think of some people. I would have to check out if they are interested in investing time in a non-GNOME project.

Hofer-Julian:
@m3nu regarding the flatpak. All I did was update the dependencies and move from yaml to json. It is also not used by the users. This happens at a separate flathub repository. So if there are problems the effect would be really small.

Hofer-Julian:
@m3nu Thank you. I clearly missed these problems. If you mean with background feature the os.fork part, I would prefer to remove that. It would remove some logic, would make the code a little bit more readable and we wouldn't need to explain new contributors that they need to add the foreground option in order to debug Vorta.

Otherwise 2, because the foreground setting was implemented in order to use Vorta intuitively on systems without system tray and I would rather not remove it.

m3nu:
All for removing the option. Most people seem to need it in foreground anyways (macOS, Flatpak, Systemd?). So just having a `--background` option for special cases seems sensible.

Let's make a new PR for it. Can you just keep the Flatpak stuff in this PR and I do another one for the foreground stuff.

Hofer-Julian:
Yes, I will do some cherry-picking and force push to this branch.

Hofer-Julian:
I reduced this PR to the flatpak-specific commits and the flatpak still builds without problems.

m3nu:
I can't test it, but if you're confident to go ahead, let's merge it.

Hofer-Julian:
Yes, let's merge it.

m3nu:
Not sure about the macOS test issue. This used to happen with parallel tests, which are disabled already. Will try to debug this.

Hofer-Julian:
@m3nu thanks!

m3nu:
The error occurs only when it's forked (getting renderer errors). Suggest to merge this and then remove the default forking as discussed.

PR #373, opened 2019-10-22
Rewrite of borg diff core

Hofer-Julian:
It now heavily relies on regex.
This was done in order to make the code more stable
and easier to maintain.

Initial testing would be much appreciated as the logic is not trivial and string parsing is error prone. 
I tested it on various combinations with changed permissions, added links, directories and other files, but it is still possible that I missed certain combinations.

### DIFF
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -76,6 +76,8 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
 
         self.populate_from_profile()
 
+        self.selected_archives = None
+
     def _set_status(self, text):
         self.mountErrors.setText(text)
         self.mountErrors.repaint()
@@ -429,9 +431,10 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         window.show()
 
         if window.exec_():
-            selected_archives = window.selected_archives
-            archive_cell_newer = self.archiveTable.item(selected_archives[0], 4)
-            archive_cell_older = self.archiveTable.item(selected_archives[1], 4)
+            if window.selected_archives:
+                self.selected_archives = window.selected_archives
+            archive_cell_newer = self.archiveTable.item(self.selected_archives[0], 4)
+            archive_cell_older = self.archiveTable.item(self.selected_archives[1], 4)
             if archive_cell_older and archive_cell_newer:
                 archive_name_newer = archive_cell_newer.text()
                 archive_name_older = archive_cell_older.text()
@@ -457,5 +460,5 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
             window = DiffResult(result['data'], archive_newer, archive_older)
             self._toggle_all_buttons(True)
             window.setParent(self, QtCore.Qt.Sheet)
-            self._window = window  # for testing
+            self._resultwindow = window  # for testing
             window.show()
--- a/src/vorta/views/diff_dialog.py
+++ b/src/vorta/views/diff_dialog.py
@@ -3,12 +3,11 @@ from PyQt5.QtWidgets import QHeaderView, QTableView, QTableWidgetItem
 
 from vorta.utils import get_asset
 
-uifile = get_asset('UI/diffdialog.ui')
+uifile = get_asset("UI/diffdialog.ui")
 DiffDialogUI, DiffDialogBase = uic.loadUiType(uifile)
 
 
 class DiffDialog(DiffDialogBase, DiffDialogUI):
-
     def __init__(self, archiveTable):
         super().__init__()
         self.setupUi(self)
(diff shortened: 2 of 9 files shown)


Hofer-Julian:
You might remember that I tried to do that, but failed because of the global variables. If you see the way to do it, please push to this branch, I activated this option :)

Hofer-Julian:
Btw, I thought subclassing and inheriting is the same thing... 🤔

m3nu:
> Btw, I thought subclassing and inheriting is the same thing... 🤔

Never mind. Just realized I compared the wrong files. I really wanted to compare `src/vorta/views/diff_result.py` and `src/vorta/views/extract_dialog.py`. I wrote "and" up there because I was thinking of 2 options to avoid all this duplicate code:

1. Have a common `TreeModel` class and `extract` and `diff` both use it. (best option)
2. Your `DiffResult` inherits from the existing `ExtractDialog`. (you say this failed due to global (module-level) variables)

The module-level variables are there because the module combines three smaller classes. But shouldn't be impossible to change that.

But then I'm not sure it's worth the effort, since I hardly ever use extract and never needed diff.

Hofer-Julian:
> But then I'm not sure it's worth the effort, since I hardly ever use extract and never needed diff.

I don't think it would be worth it. Especially since the mid-term goal is to replace it with the go version.

m3nu:
Issues:

- no tests for this feature
- duplicate implementation of the `pretty_bytes()` function from `vorta.utils` module
- `TreeModel` and related code is duplicate between `vorta/views/extract_dialog.py` and `vorta/views/diff_dialog.py`

Hofer-Julian:
> Issues:
> 
>     * no tests for this feature
> 
>     * duplicate implementation of the `pretty_bytes()` function from `vorta.utils` module
> 
>     * `TreeModel` and related code is duplicate between `vorta/views/extract_dialog.py` and `vorta/views/diff_dialog.py`

Yes, but all of these points are true for the existing diff implementation as well, and I really see no benefit into investing time to write tests, when Vorta-Python is currently not actively developed on + will be replaced soon™ with Vorta-Go.

m3nu:
I made a mistake by merging/allowing this feature in the first place without any tests or DRY code. That's on me.

So either let's fix it now or do another PR to remove the feature again until it's ready.

m3nu:
> Yes, but all of these points are true for the existing diff implementation as well, and I really see no benefit into investing time to write tests, when Vorta-Python is currently not actively developed on + will be replaced soon™ with Vorta-Go.

Progress with Go isn't great for now. I'm not happy with event handling and didn't get the translations to work.

At the same time I was able to fix the signing issues of Vorta-Py on macOS. Happy to have something that works.

Hofer-Julian:
> 
> 
> > Yes, but all of these points are true for the existing diff implementation as well, and I really see no benefit into investing time to write tests, when Vorta-Python is currently not actively developed on + will be replaced soon™ with Vorta-Go.
> 
> Progress with Go isn't great for now. I'm not happy with event handling and didn't get the translations to work.
> 
> At the same time I was able to fix the signing issues of Vorta-Py on macOS. Happy to have something that works.

Too bad that Go makes problems.
If the Python-version is here to stay for some time, then it might be actually worth to invest some effort to make it maintainable, which of course involves unit tests. I can't promise when I will have time, but I will look into, when I have it.

Hofer-Julian:
I would however not (at least for now) refactor your code in order to deduplicate some of mine. Is that acceptable for you?

m3nu:
The filetree code is mostly copied from official Qt examples. No problem to refactor it a bit.

I do want to avoid having too many obscure features that need updates and maintenance in the future. Would rather focus on getting a few core features right, even if they are boring.

Hofer-Julian:
All requirements should be fulfilled now (test + DRY)
I haven't done much testing yet, but I've noticed one bug with extract, which occurs at the master branch as well, so it probably should be dealt with in a separate PR.

Testing and code review would be appreciated.

m3nu:
Thanks. Hope you learnt a lot, like I did when first adding tests.

I know there is one infrequent race condition, where the test ends, while a borg thread is still in the background. Planning to fix this the next time I see it.

Hofer-Julian:
Yes, in the end I was quite glad I did it.

This one is not a race condition, but it shows two trees in the export dialog. If I can reproduce it, I'll open an issue for that.

m3nu:
Looks good so far. Let me take a few more days to use it locally and then merge it.

Hofer-Julian:
I will do the same.
Please also test if you see a different behavior for export for repositories in the home directory and other places.

m3nu:
The feature worked well locally. Just had one idea on how to simplify it:

Instead of having another window to select 2 archives, why not select them on the first screen. And if a user clicks "Diff" without selecting 2 archives, it says "First select two archives using Ctrl or Cmd."

Hofer-Julian:
That's actually where I started. I just found out that it can be quite annoying to manually reduce the number of archives to 1 (or 2) when you (accidentally) selected  multiple of them. I tried to implement ways to only select two but they were buggy. I was therefore cautious to add a feature which makes other features less pleasant to use and went with the intermediate dialog.

m3nu:
Understood. Then we're done with this PR?

Hofer-Julian:
From my side, yes.

Issue #263, opened 2019-04-19
Schedule backup at next opportunity


MatejKovacic:
I am using Vorta on Ubuntu and Linux Mint and set that it should run every time at 10:00.

Now, the problem is, that sometimes computer is turned off at 10:00. Su the user comes at 11:00, starts the computer, do some work and turn the computer off. And so the next day and the day after. Since computer is switched off several days at 10:00, scheduled backup does not run. So I am missing backups for a three days.

The solution would be, that backup will be run at 10:00 **or** at the next possible opportunity. So, Vorta should have an option to enable scheduled backup at a precise time, **or** at a precise time and if that was not successful, at the next possible opportunity.

So in latter case, user turns the computer on at 11:00, Vorta detects that backup at 10:00 was not run, so it runs it right away.

m3nu:
Should be simple to do by [adjusting](https://apscheduler.readthedocs.io/en/latest/userguide.html#missed-job-executions) APScheduler's `misfire_grace_time`.

daniel-k:
I was researching the same thing, studied the code a bit and then found this issue. After taking a look into the docs of APScheduler, I think that just increasing `misfire_grace_time` wouldn't work in this case because (I assume) when the computer is turned off and started again, Vorta would start freshly too. Since it currently reconstructs the scheduled jobs at every program start from it's internal database and not by using a persisted job store, the job would not be "missed" because from APScheduler's point of view, it is still in the future.

I see two solutions to this:
1. use a persisted job store (but they bring some constraints)
2. keep track of the last execution and check if there's a missed "slot" since then

My personal motivation looking into schedulers is that I want an option to run the backup at a certain interval, e.g. "every 24h". There's no rule at which time of the day I'd use my laptop. So for me it makes no sense to have rule attached to a certain hour of the day. But still, I'd like to have daily backups.

@m3nu What's your take on this issue? I'd try to implement this and provide a PR, but I think it makes sense to agree on the implementation beforehand. Any guidance is well appreciated!

m3nu:
> @m3nu What's your take on this issue? I'd try to implement this and provide a PR, but I think it makes sense to agree on the implementation beforehand. Any guidance is well appreciated!

All improvements are welcome as long as they don't add a lot of code (to maintain in the future) for minor gains.

Before you implement anything, you should give an outline of the planned changes here, other maintainers a chance to give input.

daniel-k:
So I toyed around with the scheduling during the last days and came up with a working prototype.

[...]

PR #377, opened 2019-10-27
New trigger for everyday backup irrespective of actual time

daniel-k:
This PR implements a new backup trigger method for relative backup intervals. The current methods require Vorta to be running at a specific time of the day which may be not an optimal schedule for non-server systems. I figured that the most useful and simple interval would be "everyday", but I'm totally open for a different scheme. It actually was the most simple solution for the UI so there's not much lost if we decide to change it. 

In order to implement this feature, I had to change how the apscheduler instance is built. In the current implementation up until now, a new instance had to be be populated with jobs assembled from the settings database at every start of Vorta. This way, apscheduler wasn't able to detect if a backup would have been triggered in the past. Now, the job schedule is persisted into a separate table in the settings database using apscheduler's persistent jobstore.  

While testing the code, I often ran into the situation that a job was triggered when Vorta started but immediately failed. This was due to a borg version check being run at the same time, blocking the backup job. Therefore, I also implemented a job retrying mechanism (see 36d6f3b). Although I didn't test it, this should also fix failing backups with multiple profiles being scheduled at the same or overlapping times.

This PR also paves the way for fixing #263 by implementing a persisted job store. The only thing left to do should be to increase `misfire_grace_time` when creating jobs.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,6 +40,7 @@ install_requires =
   psutil
   qdarkstyle
   setuptools
+  sqlalchemy
   secretstorage; sys_platform != 'darwin'
   pyobjc-core; sys_platform == 'darwin'
   pyobjc-framework-Cocoa; sys_platform == 'darwin'
--- a/src/vorta/assets/UI/scheduletab.ui
+++ b/src/vorta/assets/UI/scheduletab.ui
@@ -31,15 +31,15 @@ font-weight: bold;
 }</string>
      </property>
      <property name="currentIndex">
-      <number>3</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="schedule">
       <property name="geometry">
        <rect>
         <x>0</x>
         <y>0</y>
-        <width>663</width>
-        <height>354</height>
+        <width>669</width>
+        <height>361</height>
        </rect>
       </property>
       <property name="font">
@@ -82,6 +82,30 @@ font-weight: bold;
             <string>Backup manually</string>
            </property>
           </widget>
+         </item>
+        </layout>
+       </item>
+       <item>
+        <layout class="QHBoxLayout" name="horizontalLayout_4">
+         <item>
+          <widget class="QRadioButton" name="scheduleEverydayRadio">
+           <property name="text">
+            <string>Backup everyday</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <spacer name="horizontalSpacer_7">
+           <property name="orientation">
+            <enum>Qt::Horizontal</enum>
+           </property>
+           <property name="sizeHint" stdset="0">
+            <size>
+             <width>40</width>
+             <height>20</height>
+            </size>
+           </property>
+          </spacer>
          </item>
         </layout>
        </item>
@@ -364,8 +388,8 @@ font-weight: bold;
        <rect>
         <x>0</x>
         <y>0</y>
-        <width>663</width>
-        <height>354</height>
+        <width>669</width>
+        <height>361</height>
        </rect>
       </property>
       <attribute name="icon">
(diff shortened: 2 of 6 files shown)


src/vorta/assets/UI/scheduletab.ui
```diff
@@ -31,15 +31,15 @@ font-weight: bold;
...
      </property>
      <widget class="QWidget" name="schedule">
       <property name="geometry">
        <rect>
         <x>0</x>
         <y>0</y>
-        <width>663</width>
-        <height>354</height>
+        <width>669</width>
+        <height>361</height>
```
daniel-k: Not sure if these changes are just noise

Hofer-Julian:
Looks good to me. Thank you for your contribution!

We should probably test the new everyday option on our machines for some days before we merge it.

Hofer-Julian:
Also we need to find out why Travis complains.
@daniel-k did you run the tests on you personal machine?

daniel-k:
Oh, I didn't see them failing. I have added `sqlalchemy` now into 45bf5be force-pushed. Hope Travis is happy now :)

daniel-k:
I just realized that with the current value for `misfire_grace_time`, the new mechanism using the persisten job store was practically useless because a backup would only be executed if it was missed by at most 3 minutes. Therefore I just disabled the timeout altogether in 6700522. 

Furthermore, ecb9014 includes for retry scheduled jobs for next schedule display in the GUI.

setup.cfg
```diff
@@ -40,6 +40,7 @@ install_requires =
   psutil
   qdarkstyle
   setuptools
+  sqlalchemy
```
m3nu: That's a heavy dependency and it does the same thing as Peewee does. I understand that it's the quickest way to get a permanent job store for APScheduler, but not too happy.

src/vorta/scheduler.py
```diff
@@ -15,15 +18,95 @@
...
+
 class VortaScheduler(QtScheduler):
+    failed_profiles = set()
+
     def __init__(self, parent):
         super().__init__()
         self.app = parent
+
+        # persist jobs to database to continue schedule of relative interval jobs
+        self.configure(jobstores={'default': SQLAlchemyJobStore(url=f'sqlite:///{DB_PATH}')})
```
m3nu: It's not a problem if 2 ORMs use the same database file?

daniel-k: They are using different tables:
```
sqlite> .tables
apscheduler_jobs    eventlogmodel       schemaversion       sourcedirmodel    
archivemodel        repomodel           settingsmodel       wifisettingmodel  
backupprofilemodel  repopassword        snapshotmodel     
```

However, it would also be fine to have put the jobs into a separate database. Right now, I don't see a strong reason either for or against it. So, your call ;)

setup.cfg
daniel-k: Sure, that's true. Looking at the [implementation of the SQLAlchemy jobstore](https://github.com/agronholm/apscheduler/blob/master/apscheduler/jobstores/sqlalchemy.py), it seems doable to port it to either peewee or plain sqlite. However, I'm not too keen right now to be honest.

sudwhiwdh:
What does it take from your side to solve this open issue?

m3nu:
We won't merge this for now, as the new dependency is too heavy for the functionality.

Still open for another way to achieve a persistent job store. Maybe one could just use the existing event log for it?

pascalziegler:
I wrote an jobstore with sqlite3 (https://github.com/agronholm/apscheduler/pull/426). Would you merge this request, if I would adopt it to use that?

m3nu:
Of course. Can be done, when it's in a public release of APScheduler.

pascalziegler:
Unfortunately the maintainer of APScheduler won't merge my PR. This means the sqlite3 jobstore won't become a part of the public release

PR #404, opened 2020-02-17
Add CONTRIBUTORS file and copyright note. Fixes #403

### LINKED ISSUE #403: Please document at least one copyright holder so that the terms of the license will be valid
Hi!

Thank you for this software.  As a longtime user of Borg, I'm confident that Vorta is exactly what people need/want :-) so I've begun packaging Vorta for Debian (and thus derivatives such as Ubuntu, Mint, Pop! OS, Elementary, et al), and this issue will block inclusion into the Debian Archive.

**Describe the bug**
A user does not possess the rights conferred by a license unless a copyright holder is named.  Briefly, a license is a contract between the creator of a work (the copyright holder) and the recipient or user of that software, and without a named copyright holder this contract does not occur, thus the license is invalid.  This occurs because copyright occurs automatically with "all rights reserved" (eg: no rights to redistribute) in countries that have adopted the Berne Convention.  As a result of this, users of Vorta technically don't have a GPL-3 license to use and/or redistribute the software.  This applies Pypi and Flathub too, by the way.

**Solution**
The easiest thing to do would be to attribute everything to @m3nu.  The next easiest solution is to assign copyright for each file to whoever added the file to the git repo.  The next easiest solution is to use git to get statistical breakdowns per-file, and to document the copyright of everyone who contributed more than 10 lines.  Most projects do not maintain a separate CONTRIBUTORS.translations file, and the translation become the property of the named copyright holders.  That said, I'd like to second @ThomasWaldmann 's proposal (https://github.com/borgbase/vorta/issues/159#issuecomment-456253325) to provide such a CONTRIBUTORS[.translations] file because I prefer to attribute the work people did to the people who did it.

See §"How to Apply These Terms to Your New Programs" of the GPL-3 for more info.  I will of course be happy to help do the work of git spelunking, if necessary, and I'm required to do a full copyright review and analysis before uploading my package to the Debian Archive.  An alternative method to adding a copyright header to each file is to maintain a CONTRIBUTORS file.

Here is another link that may be helpful: https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

Oh, and if there is more than 10 lines of BorgBackup code in Vorta then https://github.com/borgbackup/borg/blob/master/LICENSE and their AUTHORS file will need to be included for compliance with their license, probably in the `src/vorta/borg` subdir.

Best,
Nicholas

m3nu:
This is to add a copyright note and list of authors, as suggested in #403.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (C) 2018-2020 Manuel Riel and Vorta contributors (see CONTRIBUTORS.txt)
+Copyright (C) 2018-2020 Manuel Riel and Vorta contributors (see CONTRIBUTORS.md)
 
                     GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,3 +1,5 @@
+Copyright (C) 2018-2020 Manuel Riel and Vorta contributors (see CONTRIBUTORS.txt)
+
                     GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ See our website for [download links and and install instructions](https://vorta.
 - If you have questions or simply want to talk about Vorta, find us on [Matrix](https://matrix.to/#/#vorta:matrix.org). For Borg-specific topics, join the `#borgbackup` IRC channel on `chat.freenode.net` instead.
 
 ## License and Credits
-- Thank you to all the people who already contributed to Vorta: [code](https://github.com/borgbase/vorta/graphs/contributors), [translations](https://github.com/borgbase/vorta/issues/159)
-- Licensed under GPLv3. See [LICENSE.txt](LICENSE.txt) for details.
+- See [CONTRIBUTORS.txt](CONTRIBUTORS.txt) to see who programmed and translated Vorta.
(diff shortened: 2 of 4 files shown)


Hofer-Julian:
LGTM.

Regarding #403, I wonder if now is a good time to merge #373 (and fix eventual problems with it if there are any) and then make a release. We will not be able to get updates quickly to Debian/Ubuntu/... users afterwards.

LICENSE.txt
```diff
@@ -1,3 +1,5 @@
+Copyright (C) 2018-2020 Manuel Riel and Vorta contributors (see CONTRIBUTORS.txt)
```
ThomasWaldmann: it's .md currently.

README.md
```diff
@@ -25,7 +25,7 @@ See our website for [download links and and install instructions](https://vorta.
 - If you have questions or simply want to talk about Vorta, find us on [Matrix](https://matrix.to/#/#vorta:matrix.org). For Borg-specific topics, join the `#borgbackup` IRC channel on `chat.freenode.net` instead.
 
 ## License and Credits
-- Thank you to all the people who already contributed to Vorta: [code](https://github.com/borgbase/vorta/graphs/contributors), [translations](https://github.com/borgbase/vorta/issues/159)
-- Licensed under GPLv3. See [LICENSE.txt](LICENSE.txt) for details.
+- See [CONTRIBUTORS.txt](CONTRIBUTORS.txt) to see who programmed and translated Vorta.
```
ThomasWaldmann: it's .md currently.

m3nu: Right. Good find.

Issue #317, opened 2019-06-23
Nice dashboard tab


jans23:
Currently Vorta appears to be very technical-focused piece of software. One reason is that all windows (actually all tabs in the only window) are about settings. What I'm used to from other backup software instead is a start screen which gives an overview of the state and offers performing regular tasks instead. Such dashboard tab could cover aspects such as:

- Profile name: default (heading for the subsequent information)
- Last successful backup: yesterday 12:34
- Next scheduled backup: today 13:45
- Buttons: backup now, verify backup, settings

I think such dashboard-like tab would improve Vorta's appearance and usability.

m3nu:
Could be nice. There are already some stats on the first tab. How about extending it and moving e.g. the compression settings (that are less important) somewhere else?

The repository and SSH are quite important, so those should stay in a central place.

jans23:
I would clearly separate between dashboard and settings. Settings are supposed to be configured only once and stay unmodified for relatively long period. In contrast, a dashboard should contain status information and major buttons for every day use.

Let me elaborate my suggestion:

### Backup to NAS [This is the profile]

- Last successful backup: yesterday 12:34
- Next scheduled backup: today 13:45
- Buttons: start backup, restore backup, configure

### Backup to USB [This is another profile]

- Last successful backup: 1.6.2019 12:34
- Next scheduled backup: 30.6.2019 13:45
- Buttons: start backup, restore backup, configure

ThomasWaldmann:
Maybe rather use ISO 8601 date/time formatting.

prefixtry:
A kind of appealing welcome screen would be great, yes! After all, this is the first impression of new users with Vorta. As a relevant screenshot I would link https://elementary.io/docs/human-interface-guidelines#welcome-screen.

prefixtry:
A kind of appealing welcome screen would be great, yes! After all, this is the first impression new users get of Vorta. As a relevant screenshot I would like to link https://elementary.io/docs/human-interface-guidelines#welcome-screen here.

m3nu:
Right. That would be nice. PRs are always welcome.

samuel-w:
My plan is for the main window to be 2 tabs: Overview and Global Settings (current Misc tab)
The Overview tab would have all the profiles, with each profile formatted like this (just a mockup):
![mockup](https://user-images.githubusercontent.com/67625312/92217766-c43e2e00-ee5d-11ea-87b0-eaf3d586a59b.png)

[...]

PR #412, opened 2020-03-09
More reliable cancel (WIP)

### LINKED ISSUE #332: UI hanging when cancelling a running backup
Start a backup manually, then cancel it. The UI freezes, and the borg process is defunct (zombie, not waited by the parent process)

This occurs on archlinux, using 0.6.21

Please tell me if you need logs.

Regards

m3nu:
Maybe fixes #332


m3nu:
The feature worked well locally. Just had one idea on how to simplify it:

Instead of having another window to select 2 archives, why not select them on the first screen. And if a user clicks "Diff" without selecting 2 archives, it says "First select two archives using Ctrl or Cmd."

sudwhiwdh:
What does it take from your side to solve this open request?

And is there something like a roadmap to Vorta 0.7?

m3nu:
> What does it take from your side to solve this open request?
> 
> And is there something like a roadmap to Vorta 0.7?

We follow semantic versioning, so 0.7 would need some radical new features. I don't see those yet and will focus on incremental fixes.

For this PR, I can't reproduce the error for now.

sudwhiwdh:
> We follow semantic versioning, so 0.7 would need some radical new features. I don't see those yet and will focus on incremental fixes.

What do you mean "incremental fixes"?
Would https://github.com/borgbase/vorta/issues/317 be such a major change?

> For this PR, I can't reproduce the error for now.

Does that mean this PR should be discarded?
Or does it still bring improvements?

PR #419, opened 2020-03-18
Make Vorta icon rectangular

Hofer-Julian:
I tried to update the Vorta on flathub today, but noticed that with v0.6.24 the icon is stretched.
Please don't ask me why it worked until now, but by adding making the icon rectangular everything works again.

Because of that I will wait until v0.6.25 is released, before I update Vorta on flathub.


Hofer-Julian:
@m3nu if anything macOS related depends on this icon (which I think is not the case), then please briefly check if everything still works.

m3nu:
> @m3nu if anything macOS related depends on this icon (which I think is not the case), then please briefly check if everything still works.

Not used for macOS.

Issue #416, opened 2020-03-12
Issue formatting bytes


damiengole:
**Describe the bug**
Vorta does not launch (Mac OS 10.13.6), and no backups are running

**To Reproduce**
1. Go to Applications
2. Click on Vorta.app
3. See error:

vorta.i18n - DEBUG - Loading translation failed for ['C']. (found in Library/logs/Vorta)

admin$ /Applications/Vorta.app/Contents/MacOS/vorta-darwin --foreground
2020-03-12 11:50:56,877 - vorta.i18n - DEBUG - Loading translation succeeded for ['fr', 'fr-FR', 'fr-Latn-FR'].
2020-03-12 11:50:56,949 - apscheduler.scheduler - INFO - Scheduler started
2020-03-12 11:50:56,952 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-03-12 11:50:56,952 - vorta.scheduler - DEBUG - New job for profile Default was added.
Traceback (most recent call last):
  File "vorta/__main__.py", line 47, in <module>
  File "vorta/__main__.py", line 40, in main
  File "vorta/application.py", line 59, in __init__
  File "vorta/application.py", line 80, in open_main_window_action
  File "vorta/views/main_window.py", line 37, in __init__
  File "vorta/views/repo_tab.py", line 59, in __init__
  File "vorta/views/repo_tab.py", line 78, in init_repo_stats
  File "vorta/utils.py", line 105, in pretty_bytes
KeyError: 5
[2017] Failed to execute script __main__

**Desktop (please complete the following information):**
 - OS: Mac OS 10.13.6

**Additional context**
Vorta does open its window with other accounts, but without any configuration saved.

damiengole:
Apparently if I delete the file settings.db, the app launches, but I loose all the settings.

m3nu:
The error about the failed translations are unrelated. Just tried by setting the language to EN, if none is provided by macOS. Didn't make a difference.

I think you had some weird data in your settings DB and formatting something failed.

The fix here is to make the formatting fail gracefully, even if weird data is passed.

PR #433, opened 2020-04-03
Display "NaN" if size of borgrepository is out of bounds

TheReal1604:
Fixes https://github.com/borgbase/vorta/issues/416

It just displays NaN now, instead of crashing the application completely:

grafik


m3nu:
Thanks! Merged now.

PR #437, opened 2020-04-07
Set StartupWMClass= in com.borgbase.Vorta XDG desktop file

yegortimoshenko:
Before:

Screenshot from 2020-04-08 01-56-00

After:

Screenshot from 2020-04-08 02-04-33

### DIFF
--- a/src/vorta/assets/metadata/com.borgbase.Vorta.desktop
+++ b/src/vorta/assets/metadata/com.borgbase.Vorta.desktop
@@ -5,3 +5,4 @@ Exec=vorta
 Type=Application
 Icon=com.borgbase.Vorta
 Categories=Utility
+StartupWMClass=vorta


Hofer-Julian:
Hi @yegortimoshenko, thank you for your PR!
Do I see this correctly that the only differences between the screenshots are the additional icon + Vorta instead of vorta?
If yes, then the "after" state is already present on my Fedora workstation without your changes.
Looking through your pinned repos, i guess you use NixOS?

yegortimoshenko:
@Hofer-Julian Hi!

> Do I see this correctly that the only differences between the screenshots are the additional icon + Vorta instead of vorta?

Yes. Essentially, setting StartupWMClass helps desktop environment to map window by its `WM_CLASS` (you can see that by using `xprop` tool and clicking on a window) onto the application that owns that window. It has no effect other than that.

>  If yes, then the "after" state is already present on my Fedora workstation without your changes.

Are you using Flatpak? If so, that would explain the discrepancy. Don't think it will work outside of Flatpak while desktop entry name (currently `com.borgbase.Vorta`) doesn't match `WM_CLASS` (currently `vorta`), other than by setting `StartupWMClass`.

Hofer-Julian:
> @Hofer-Julian Hi!
> 
> > Do I see this correctly that the only differences between the screenshots are the additional icon + Vorta instead of vorta?
> 
> Yes. Essentially, setting StartupWMClass helps desktop environment to map window by its `WM_CLASS` (you can see that value by launching `xprop` tool and clicking on a window) onto the application that owns that window. It has no effect other than that.
> 
> > If yes, then the "after" state is already present on my Fedora workstation without your changes.
> 
> Are you using Flatpak build? If so, that would explain the discrepancy. Don't think it will work outside of Flatpak while desktop entry name (currently `com.borgbase.Vorta`) doesn't match `WM_CLASS` (currently `vorta`), other than by setting `StartupWMClass`.

Understand, thanks for explanation.

PR #460, opened 2020-05-19
Correct window size on hidpi screen

SamuelBenard:
based on this : [https://leomoon.com/journal/python/high-dpi-scaling-in-pyqt5/](https://leomoon.com/journal/python/high-dpi-scaling-in-pyqt5/)


m3nu:
Merged. Thanks for the contribution!

PR #475, opened 2020-05-28
Start removing qtdarkstyle and custom fusion style.

### LINKED ISSUE #471: support running without qdarkstyle for better Linux Desktop integration
This request is related to #331

After working on packaging Vorta and QDarkStyle, I realised that defaulting to a non-system Qt style is not what most users want.  Namely, for a package installed with the system package manager on Debian (and derivatives such as Ubuntu, Kubuntu, Neon, etc) the expectation is that all applications follow the style of the desktop as a whole.  This is from a Qt/KDE Plasma perspective; 

Pyqt5 provides access to the default style, whatever it may be.  I'm not sure what the ideal solution is, because QDarkStyle is so different from the usual way of doing things on the Linux desktop, but I'm guessing it will be something like boolean somewhere in Vorta's source.  When toggled (with a patch or with sed, when the deb is created), Vorta would not attempt to import QDarkStyle, would disable in-application theme switching (GUI), and would skip all code paths involving QDarkStyle, instead using the pyqt5 defaults.

m3nu:
Remove custom style sheets, as dark mode is now supported [natively](https://www.learnpyqt.com/blog/macos-mojave-dark-mode-support-pyqt5122/) in Qt.

Changes:

- Remove `qtdarkstyle` dependency
- Remove custom `fusion` style on Linux
- Remove workaround for different icon colors. SVGs are now colored when loaded. Resource files aren't needed any more.
- Listen for theme changes and adjust icons (window and tray) in real time (tested on macOS)

Fixes #471


m3nu:
Native Qt dark style that can be changed while Vorta is open. (One element, `QToolbox` need reopening the window for now. Minor Qt bug.)

Screen Shot 2020-05-28 at 17 54 48
Screen Shot 2020-05-28 at 17 55 02

src/vorta/assets/UI/archivetab.ui
```diff
@@ -28,11 +28,6 @@
        <bold>false</bold>
       </font>
      </property>
-     <property name="styleSheet">
-      <string notr="true">QToolBox::tab {
```
Hofer-Julian, line 32: Could elaborate on the style issue?
I've also noticed that quite a lot of bold texts on my Qt apps look broken since I've updated my OS.
I've originally thought that might be a flatpak problem.
Is this a known Qt bug?

src/vorta/views/utils.py
```diff
@@ -1,16 +1,13 @@
...
+    Return SVG icon in the correct color.
     """
-    if SettingsModel._meta.database.obj is None:
-        return 'vorta.views.dark'
-    else:
-        use_light_icon = SettingsModel.get(key='use_dark_theme').value
-        return 'vorta.views.light' if use_light_icon else 'vorta.views.dark'
+    svg_str = open(get_asset(f"icons/{icon_name}.svg"), 'rb').read()
+    if uses_dark_mode():
+        svg_str = svg_str.replace(b'#00000', b'#ffffff')
```
Hofer-Julian, line 11: This is pretty cool!

src/vorta/qt_single_application.py
```diff
@@ -18,9 +17,6 @@ def __init__(self, id, *argv):
         super().__init__(*argv)
         self._id = id
 
-        if sys.platform.startswith('linux'):
```
Hofer-Julian, line 21: It does not seem to be necessary anymore, but I don't understand why.
Could please explain that?

m3nu: No idea. 🤷‍♂️

src/vorta/assets/UI/archivetab.ui
m3nu: I think a Qt bug. I just made it bold, but it totally destroyed the style of that widget.

m3nu: So better be careful with any styles.

Hofer-Julian: Good to know!

src/vorta/application.py
```diff
@@ -64,6 +56,17 @@ def __init__(self, args_raw, single_app=False):
...
+        self.installEventFilter(self)
+
+    def eventFilter(self, source, event):
+        if event.type() == QtCore.QEvent.ApplicationPaletteChange and type(source) == MainWindow:
+            self.main_window.set_icons()
+            self.main_window.repoTab.set_icons()
+            self.main_window.archiveTab.set_icons()
+            self.main_window.scheduleTab.set_icons()
+        if event.type() == QtCore.QEvent.ApplicationPaletteChange and source == self.tray.contextMenu():
+            self.tray.set_tray_icon()
```
real-yfprojects, line 68: Is this switch over the source widget necessary? I suppose that wouldn't hurt to update the icons on every palette change.

PR #477, opened 2020-05-29
Prevent the same repo from being added twice

### LINKED ISSUE #473: UI Crash when adding same repo twice
**The bug**
Re-adding the same repo and then deleting it creates an inconsistency between the app and the database which crashes the UI.

**To Reproduce**
1. Have one existing repository connection
2. Click on Add Existing Repository
3. Put in the same path for the existing connection. This seems to update the UI, but point to the same database entry. If you close the UI and reopen it it will just show the one repo, but in the existing UI session it remains as a second entry.
4. Delete the new entry.
5.  Select the other remaining entry. The UI will crash and cannot be recovered without fixing or recreating the database file.

**More Info**
Version: v0.6.23
The error is as follows when you try to open the UI window after deleting the bug:
```
  File "/app/lib/python3.7/site-packages/peewee.py", line 6665, in get
    return clone.execute(database)[0]
  File "/app/lib/python3.7/site-packages/peewee.py", line 4121, in __getitem__
    return self.row_cache[item]
IndexError: list index out of range

During handling of the above exception, another exception occurred:

  File "/app/lib/python3.7/site-packages/vorta/application.py", line 80, in open_main_window_action
  File "/app/lib/python3.7/site-packages/vorta/views/main_window.py", line 37, in __init__
    self.repoTab = RepoTab(self.repoTabSlot)
  File "/app/lib/python3.7/site-packages/vorta/views/repo_tab.py", line 59, in __init__
    self.init_repo_stats()
  File "/app/lib/python3.7/site-packages/vorta/views/repo_tab.py", line 74, in init_repo_stats
    repo = self.profile().repo
  File "/app/lib/python3.7/site-packages/peewee.py", line 4269, in __get__
    return self.get_rel_instance(instance)
  File "/app/lib/python3.7/site-packages/peewee.py", line 4260, in get_rel_instance
    obj = self.rel_model.get(self.field.rel_field == value)
  File "/app/lib/python3.7/site-packages/peewee.py", line 6242, in get
    return sq.get()
  File "/app/lib/python3.7/site-packages/peewee.py", line 6670, in get
    (clone.model, sql, params))
vorta.models.RepoModelDoesNotExist: <Model: RepoModel> instance matching query does not exist:
SQL: SELECT "t1"."id", "t1"."url", "t1"."added_at", "t1"."encryption", "t1"."unique_size", "t1"."unique_csize", "t1"."total_size", "t1"."total_unique_chunks", "t1"."extra_borg_arguments" FROM "repomodel" AS "t1" WHERE ("t1"."id" = ?) LIMIT ? OFFSET ?
Params: [1, 1, 0]
```

m3nu:
The change to repo_add_dialog.py adds a check in the database
to see if the repo already exists, and blocks the ui with an
error message if so.

The change to repo_tab.py refactors the selector population logic
so that the selector stays in sync with the database and doesn't
leave the entry in the UI that triggers the crash.

Fixes #473

### DIFF
--- a/src/vorta/views/repo_tab.py
+++ b/src/vorta/views/repo_tab.py
@@ -25,9 +25,7 @@ class RepoTab(RepoBase, RepoUI, BackupProfileMixin):
         self.repoSelector.addItem(self.tr('+ Initialize New Repository'), 'new')
         self.repoSelector.addItem(self.tr('+ Add Existing Repository'), 'existing')
         self.repoSelector.insertSeparator(3)
-        for repo in RepoModel.select():
-            self.repoSelector.addItem(repo.url, repo.id)
-
+        self.set_repos()
         self.repoSelector.currentIndexChanged.connect(self.repo_select_action)
         self.repoRemoveToolbutton.clicked.connect(self.repo_unlink_action)
 
@@ -58,6 +56,13 @@ class RepoTab(RepoBase, RepoUI, BackupProfileMixin):
 
         self.init_repo_stats()
         self.populate_from_profile()
+
+    def set_repos(self):
+        count = self.repoSelector.count()
+        for x in range(4, count):
+            self.repoSelector.removeItem(4)
+        for repo in RepoModel.select():
+            self.repoSelector.addItem(repo.url, repo.id)
 
     def populate_from_profile(self):
         profile = self.profile()
@@ -170,7 +175,7 @@ class RepoTab(RepoBase, RepoUI, BackupProfileMixin):
             profile.repo = new_repo.id
             profile.save()
 
-            self.repoSelector.addItem(new_repo.url, new_repo.id)
+            self.set_repos()
             self.repoSelector.setCurrentIndex(self.repoSelector.count() - 1)
             self.repo_added.emit()
             self.init_repo_stats()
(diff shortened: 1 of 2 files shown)


m3nu:
All good. Will add this after the larger #475 . Thanks a lot to @vapaavetehinen ! 👍

src/vorta/views/repo_tab.py
```diff
@@ -59,6 +57,13 @@ def __init__(self, parent=None):
         self.init_repo_stats()
         self.populate_from_profile()
 
+    def set_repos(self):
+        count = self.repoSelector.count()
+        for x in range(4, count):
```
Hofer-Julian, line 62:
```suggestion
        # The repositories are listed after the 4th entry in the repoSelector
        for x in range(4, count):
```

m3nu: Good to add a comment here. These numbers is usually what we forget to change. Finding the index of the separator would work too, but maybe overkill.

m3nu:
Thanks for the review. I'll fix the conflicts locally. Pressed the "re-request review" by accident.

m3nu:
Merged after fixing conflicts locally. Thanks @vapaavetehinen and @Hofer-Julian.

PR #478, opened 2020-05-30
Add new setting to store previously selected profile ID. Fixes #241

### LINKED ISSUE #241: Start with the last used "profile"
I noticed, that Vorta appears to always start with the alphabetically first (or chronologically first set up?) **profile**.

Please let me suggest  - this is what I prefer - that Vorta starts always with the **last** **used** profile.

m3nu:
- adds new col to Settings model to store strings.
- sort profiles alphabetically
- on profile change, store id in DB and use when window is opened

### DIFF
--- a/src/vorta/models.py
+++ b/src/vorta/models.py
@@ -165,8 +165,8 @@ class SchemaVersion(pw.Model):
 class SettingsModel(pw.Model):
     """App settings unrelated to a single profile or repo"""
     key = pw.CharField(unique=True)
-    value = pw.BooleanField()
-    str_value = pw.CharField(null=True)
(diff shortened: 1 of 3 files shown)


src/vorta/models.py
```diff
@@ -224,6 +225,10 @@ def get_misc_settings():
             'label': trans_late('settings',
                                 'Open main window on startup')
         },
+        {
+            'key': 'previous_profile_id', 'str_value': '1', 'type': 'internal',
+            'label': 'Previously selected profile'
```
Hofer-Julian, line 230: I assume we want to translate this as well :)

m3nu:
Merged locally after resolving conflicts.

Issue #314, opened 2019-06-22
Use .nobackup by default, if no other exclusions set


m3nu:
**Is your feature request related to a problem? Please describe.**
Without user action, Vorta will show `.nobackup` under the sources tab, suggesting that this is the default. In reality, we don't exclude anything unless the user adds his own settings in that field.

This lead to some confusion, where the functionality was expected:

> My misunderstanding comes from the fact that I believed that .nobackup was automatically used as a filter, but now I think that it’s just a suggestion in the interface (which is why it’s dimmed). I added it explicitely and now it does work!

**Describe the solution you'd like**
1. Actually exclude folders with `.nobackup` file when the user doesn't change the field.
2. Change the text to make it clear that the placeholder has now effect. Maybe like this: `E.g. .nobackup`

gtivey:
Hi all --- Run into this issue today when I tried to backup with a .nobackup file in an early directory in the backup process without success. After finding this post, I was able to tag the directory with an .exclude file and it worked just fine. Thanks for the post!

m3nu:
I've changed the placeholder text a bit. Since the `.nobackup` file isn't really a standard, I'm reluctant to add it as default. But the misunderstanding described above should still be avoided.

![Screen Shot 2020-06-01 at 13 41 38](https://user-images.githubusercontent.com/3916435/83379299-b798dc80-a40d-11ea-9a6a-ac0d7ace8294.jpg)

sudwhiwdh:
> But the misunderstanding described above should still be avoided.

I am actually a little irritated by the (new) placeholders texts as well.
/.DS_Store looks like a macOS-only thing. And under Linux CACHE would probably be ~/.cache. I have never heard of .nobackup by the way. Maybe we can find some cross-system examples that will give the users an idea of what it is all about?

m3nu:
`.nobackup` and `CACHE` (or `CACHEDIR.TAG`) are both examples from [Borg](https://borgbackup.readthedocs.io/en/stable/usage/create.html) or [Borgmatic](https://torsion.org/borgmatic/docs/reference/configuration/). Not making stuff up here.

How about `**/.cache` and `.nobackup` as examples?

Hofer-Julian:
> How about `**/.cache` and `.nobackup` as examples?

Yeah, that would probably be better.

sudwhiwdh:
`.nobackup` is also macOS only, right?
Is `**/.cache` the macOS equivalent of `~/.cache` on my Linux system?

[...]

PR #480, opened 2020-06-01
Add error message when Borg binary is missing. Fixes #333

### LINKED ISSUE #333: Alert on missing Borg binary instead of showing placeholder version number
I installed Vorta with the pip3 method on GalliumOS (Linux OS, GUI:Xfce) and had the following minor issue:
Since I did **not install borg** before, Vorta complained nicely, when I tried to create a test-repository. 
However, in **Vorta>Misc**   the information about the borg Version was wrong, since it showed: 
**Borg Version: 1.1.10**        
However, borg was not installed at all. 

Expected behaviour: 
If borg is not available Vorta should show this in the Misc tab. Something like: 
**Borg Version: not installed**

- Linux Gallium OS 3.0 
kernel 4.1.18-galliumos    GUI: Xfce

m3nu:
- Display a proper error message when there is an issue with the Borg binary (wrong return code, not found, ..) #333 
- Fix PyQt to versions below 5.15 for now
- Small change to file exclusion placeholder text. #314

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ setup_requires =
 install_requires =
   appdirs
   paramiko
-  pyqt5
+  pyqt5 < 5.15
   peewee
   python-dateutil
   apscheduler
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -2,16 +2,18 @@ import os
 import sys
 import sip
 from PyQt5 import QtCore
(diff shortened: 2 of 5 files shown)


m3nu:
Noticed that tests on Linux fail using the latest PyQt released 5 hours ago. So fixed it to versions below that for now.

Also snuck in a smaller issue #314 .

PR #481, opened 2020-06-01
Update flatpak manifest

Hofer-Julian:
This PR updates the manifest to the current state on https://github.com/flathub/com.borgbase.Vorta

### DIFF
--- a/flatpak/com.borgbase.Vorta.json
+++ b/flatpak/com.borgbase.Vorta.json
@@ -1,7 +1,7 @@
 {
     "app-id": "com.borgbase.Vorta",
     "runtime": "org.kde.Platform",
-    "runtime-version": "5.13",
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Btw, any reason why we test for Python 3.6, when we require 3.7?

m3nu:
> Btw, any reason why we test for Python 3.6, when we require 3.7?

`setup.cfg` and the docs both say `python_requires = >= 3.6`. Where does it say 3.7 or higher?

Hofer-Julian:
> > Btw, any reason why we test for Python 3.6, when we require 3.7?
> 
> `setup.cfg` and the docs both say `python_requires = >= 3.6`. Where does it say 3.7 or higher?

Ah, sorry my mistake

PR #488, opened 2020-06-04
Desktop file: fix syntax, enhance Categories, add a Keyword.

sten0:
Keywords must not duplicate words in Name nor GenericName.  It might
be useful to add additional Keywords, but I haven't been able to find
a package that duplicates entries from Categories.


sten0:
P.S. Keywords are now highly recommended as most desktops provide a mechanism to search for applications by keyword.

PR #501, opened 2020-06-13
Use newest translation data from transifex

Hofer-Julian:
I've translated the remaining German strings.

Also Pavel Borecki completely translated Czech.

### DIFF
--- a/src/vorta/i18n/ts/vorta.cs.ts
+++ b/src/vorta/i18n/ts/vorta.cs.ts
@@ -30,59 +30,39 @@
 <context>
     <name>AddRepoWindow</name>
     <message>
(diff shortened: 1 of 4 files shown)


m3nu:
Great. Pavel is already a contributor. But for new translators we need to add them to the [contributors file](https://github.com/borgbase/vorta/blob/master/CONTRIBUTORS.md) too.

Hofer-Julian:
Good to know!

So I guess, you and me should be add to German translators?

m3nu:
If you like, sure. No need to add me though. Thomas did most translation work in the beginning.

Hofer-Julian:
If you don't want to be added, I also won't add myself.

PR #504, opened 2020-06-13
Save window state

Hofer-Julian:
- stores window width&height in peewee database before closing
- gets the window state during initialization of MainWindow

### DIFF
--- a/src/vorta/models.py
+++ b/src/vorta/models.py
@@ -215,6 +215,14 @@ def get_misc_settings():
             'key': 'previous_profile_id', 'str_value': '1', 'type': 'internal',
             'label': 'Previously selected profile'
         },
+        {
+            'key': 'previous_window_width', 'str_value': '800', 'type': 'internal',
+            'label': 'Previous window width'
+        },
+        {
+            'key': 'previous_window_height', 'str_value': '600', 'type': 'internal',
+            'label': 'Previous window height'
+        },
     ]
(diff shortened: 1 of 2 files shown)


m3nu:
How about using 2 settings, rather than adding more cols? Like `previous_window_width` and `previous_window_height`. That way you don't even need to change the schema.

Hofer-Julian:
Sure, why not.
Do you want to have `previous_profile_id ` moved to this second setting as well?

m3nu:
The settings table has 2 cols to save values currently:

- `value`: the original field for bool
- `value_str`: Added for `previous_profile_id`

You can probably just use `value_str` to save the height and width and typecast it when using.

```
class SettingsModel(pw.Model):
    """App settings unrelated to a single profile or repo"""
    key = pw.CharField(unique=True)
    value = pw.BooleanField(default=False)
    str_value = pw.CharField(default='')
    label = pw.CharField()
    type = pw.CharField()
```

Hofer-Julian:
Okay, I understand what you mean now.
Thought about it as well, but didn't really understand why you've used Strings to store integers in the first place. But let's do it that way!

Hofer-Julian:
@m3nu I think you comments are resolved now.

sudwhiwdh:
What about saving and restoring the window position as well?
Or whether the window is maximized or not.

Issue #378, opened 2019-10-28
Implement --pattern or --patterns-from options


mahikeulbody:
It would be nice to add `--pattern` or `--patterns-from` options in addition to `--exclude` and `--exclude-if-present` options already implemented.

It would be easier to "migrate" to Vorta when one has a `patterns-from` file already set to use with borg.

Thanks for the nice work !

Hofer-Julian:
According to https://borgbackup.readthedocs.io/en/stable/usage/create.html this is still experimental. Not sure if it makes sense to include it in Vorta yet...

palto42:
I'm using the `--pattern` option since many month via `borgmatic`and had no issues. In my view it's much easier to use since it allows a mix of include/exclude which can avoid the usage of complex regular expressions (e.g. exclude some path except specific sub-folder).

Issue #399, opened 2020-01-21
Suggestion: Exclude patterns GUI improvements


geckolinux:
First of all, many thanks for Vorta, this is an extremely important tool to motivate less technical people to take advantage of the all the goodness of Borg.

The only part of the GUI that I found slightly confusing was the exclude patterns, especially after reading the link to the manual, which for a non-technical user doesn't explain clearly what exclude patterns are for or how to use them and how they differ from the other apparently similar methods for excluding certain files and/or directories from backups.

I think it would be nice for Vorta to have a few GUI exclusion improvements, while still allowing to manually type in patterns for experts who prefer that. Maybe something like:

1. A field with directories and files to be excluded, basically the inverse of the **Source Folders and Files to Back Up** pane. This will probably cover 95% of beginners' needs.
2. A checkbox to exclude folders with the `.nobackup` file.
3. A GUI way of building slightly more complex exclusion rules, similar to buildregex.com , but obviously not as comprehensive, just to cover the most common rules of files and/or directories that start with `BLAH`, that contain`BLAH` anywhere in the name, or end with `BLAH`, with a little additional help on the use of wildcards.

Thanks a lot!

m3nu:
I like the checkbox idea, item 2.

We probably don't have the resources to build item 3, but I'd like to have some examples for different OS.

Item 1 would be a list of absolute paths. Also possible, I think.

m3nu:
Just noticed that Nextcloud (desktop client) comes with a nice list of default exclusions. Maybe something for Vorta to assimilate.

<img width="932" alt="Screen Shot 2020-02-09 at 21 17 19" src="https://user-images.githubusercontent.com/3916435/74102763-a3d91a00-4b81-11ea-9059-ee578082f458.png">

I agree that one simple kind of patterns would be enough.

m3nu:
Found some more details on macOS exclusions. They work via extended attributes. I think excluding those should be implemented in Borg directly. It already processes xattrs.

https://apple.stackexchange.com/questions/25779/on-os-x-what-files-are-excluded-by-rule-from-a-time-machine-backup

mauromol:
I also found the exclusions part of the GUI quite confusing.

[...]

PR #507, opened 2020-06-15
Implement experimental pattern support. Fixes borgbase/vorta#378

samuel-w:


m3nu:
This is a larger change and may need some discussion. I'll test it a bit over the next days.

m3nu:
There is a big overlap with the patterns we already support and users are already complaining that this tab is too complicated. I'd rather 

Readme for this Borg feature: https://borgbackup.readthedocs.io/en/stable/usage/help.html

Before adding this, I'd rather work on additional exclude improvements as discussed in #399 and then hide those advanced options in a QToolbox (as we do in the Archive tab)

real-yfprojects:
Since the branch and the commits are lost, I'll close this. Work on the exclude pattern GUI is currently tracked in #907.

PR #508, opened 2020-06-15
Fix issue 450

samuel-w:


m3nu:
You're on a run here. 🤟 Looks good at first sight. Will test over the next days.

m3nu:
I do wonder if we should just set this flag `BORG_UNKNOWN_UNENCRYPTED_REPO_ACCESS_IS_OK ` all the time?

samuel-w:
I believe that this flag exists to prevent accidental data leaks. Setting it to always be on leads to a security risk as the backup would be stored in plain text on a remote server the user has no control over. Having a tick box requires the user to acknowledge the risk.

m3nu:
Then the tooltip is probably the smart way to do it. Will try it again this morning.

m3nu:
Closing this, since user and repos were deleted. It also didn't have much effect.

PR #510, opened 2020-06-15
Fix parsing "changed link" lines

### LINKED ISSUE #509: Crash in DiffResult
**Describe the bug**
Vorta crashes when trying to show diff between backups.

**Additional context**
It seems Vorta logs don't include the crash traceback, but I've found it in system logs:
```
cze 14 22:44:13 vorta: UnboundLocalError: local variable 'full_path_index' referenced before assignment
cze 14 22:44:13 vorta:     full_path = line[full_path_index:]
cze 14 22:44:13 vorta:   File "/app/lib/python3.7/site-packages/vorta/views/diff_result.py", line 60, in parse_line
cze 14 22:44:13 vorta:     files_with_attributes.append(parse_line(line))
cze 14 22:44:13 vorta:   File "/app/lib/python3.7/site-packages/vorta/views/diff_result.py", line 72, in __init__
cze 14 22:44:13 vorta:     window = DiffResult(result['data'], archive_newer, archive_older)
cze 14 22:44:13 vorta:   File "/app/lib/python3.7/site-packages/vorta/views/archive_tab.py", line 460, in list_diff_result
cze 14 22:44:13 vorta: Traceback (most recent call last):
cze 14 22:42:47 vorta: 2020-06-14 22:42:47,325 - vorta.borg.borg_thread - INFO - Running command /app/bin/borg diff --info --log-json XXX@XXX.repo.borgbase.com:repo::XXX-2020-06-14T17:00:00 XXX-2020-06-14T18:00:00
```

ktosiek:
Make sure borg diff lines that start with "changed link" are handled properly. Fixes #509.

This code will still have to work for a little bit longer: https://github.com/borgbackup/borg/issues/3765


src/vorta/views/diff_result.py
```diff
@@ -88,6 +37,62 @@ def parse_line(line):
...
+    files_with_attributes = []
+    nested_file_list = nested_dict()
+
+    def parse_line(line):
+        if line:
+            line_split = line.split()
+        else:
+            return 0, "", "", ""
+
+        if line_split[0] in {'added', 'removed', 'changed'}:
```
Hofer-Julian, line 50: Is it possible that you meant `['added', 'removed', 'changed']`?

Hofer-Julian: I just checked, a dict works as well, but a list or tuple still makes more sense in my opinion.

ktosiek: I think using a set is as good, or maybe even a hair faster, but I'll change it to a list if that's what you prefer.

Hofer-Julian: No, it's okay.
I wasn't aware that curly braces without a colon, are equivalent to set() in Python https://docs.python.org/3/tutorial/datastructures.html#sets

ktosiek: Oh, but that's not a dict - it's a [set](https://docs.python.org/3/tutorial/datastructures.html#sets).

tests/test_diff.py
```diff
@@ -0,0 +1,13 @@
...
+from vorta.views.diff_result import parse_diff_lines
+
+
+@pytest.mark.parametrize('line, expected', [
+    ('changed link        some/changed/link',
+     (0, 'changed', 'link', 'some/changed')),
+    (' +77.8 kB  -77.8 kB some/changed/file',
+     (77800, 'modified', 'file', 'some/changed')),
+])
+def test_parsing_diff_lines(line, expected):
```
Hofer-Julian, line 11: This function should probably be in the same file as https://github.com/borgbase/vorta/blob/936d3c10d09cd601351ed49d280a8d8b2aab99cf/tests/test_archives.py#L135

I have no strong feelings regarding which one of the two should be moved

ktosiek:
Can someone restart the build? It feels like that one failure is a fluke.

Hofer-Julian:
Yeah, we sometimes have falsely failing test cases

PR #511, opened 2020-06-16
Add paste button for source paths

### LINKED ISSUE #466: Allow pasting a list of sources
When adding source folders via the GUI, I have to add them one by one, but I can add exclude patterns all at once. If I have a list of exclude patterns in a text editor, I can just copy and paste the list, and I'd like to be able to do the same for includes, instead of navigating through the file explorer to each of the fifteen sources. Maybe this is related to https://github.com/borgbase/vorta/issues/378

samuel-w:
Implements request in a different way than suggested, I hope it's still OK

Fixes #466

### DIFF
--- a/src/vorta/assets/UI/sourcetab.ui
+++ b/src/vorta/assets/UI/sourcetab.ui
@@ -55,6 +55,13 @@
          </property>
         </widget>
        </item>
+       <item>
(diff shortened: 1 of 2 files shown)


m3nu:
Thanks for adding this! I'll check it out over the next days.

PR #518, opened 2020-06-18
Partially fixes #385

### LINKED ISSUE #385: Passphrase handling improvements
While keyrings are nice and well, it would be even safer if the passphrase weren't stored anywhere at all.

An option to prevent passphrase storage in the keyring, a backup should then prompt for the passphrase.

samuel-w:
I only implemented 2 and 3 of https://github.com/borgbase/vorta/issues/385#issuecomment-635060484.

### DIFF
--- a/src/vorta/views/repo_add_dialog.py
+++ b/src/vorta/views/repo_add_dialog.py
@@ -52,7 +52,7 @@ class AddRepoWindow(AddRepoBase, AddRepoUI):
         keyringName = 'plaintext on disk. Anyone with access to the database can read the password'
         if not plaintextPass:
             keyringName = VortaKeyring.get_keyring().__class__.__name__[5:-7]  # Trims "Vorta" and "Keyring"
-        self.errorText.setText('The password will be stored in ' + keyringName)  # Misusing errorText
+        self.passwordLabel.setText('The password will be stored in ' + keyringName)
 
     def choose_local_backup_folder(self):
         def receive():
@@ -132,18 +132,22 @@ class AddRepoWindow(AddRepoBase, AddRepoUI):
             self._set_status(self.tr('This repo has already been added.'))
             return False
 
-        if self.__class__ == AddRepoWindow:
-            if self.values['encryption'] != 'none':
-                if len(self.values['password']) < 8:
-                    self._set_status(self.tr('Please use a longer passphrase.'))
-                    return False
-
         return True
 
     def validate_passwords(self):
+        msg = "Passwords must be "
         passEqual = self.passwordLineEdit.text() == self.confirmLineEdit.text()
-        self.saveButton.setEnabled(passEqual)
-        self.errorText.setText("" if passEqual else "Passwords must identical")
+        passLong = len(self.values['password']) >= 8
+
+        if not passEqual:
+            msg += "identical"
+        if not (passEqual or passLong):
+            msg += " and "
+        if not passLong:
+            msg += "greater than 8 characters long"
+
+        self.passwordLabel.setText("" if passEqual and passLong else msg)
+        self.saveButton.setEnabled(passEqual and passLong)
 
 
 class ExistingRepoWindow(AddRepoWindow):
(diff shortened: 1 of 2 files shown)


samuel-w:
Tests seem to break because I disable the add button on start, can either change tests or change my code.

Hofer-Julian:
I think changing the test would be the way to go :)

samuel-w:
The Ubuntu tests aren't able to install

samuel-w:
Needs testing on macOS to see if Darwin shows up.

src/vorta/views/repo_add_dialog.py
```diff
@@ -141,7 +141,7 @@ def validate(self):
         if RepoModel.get_or_none(RepoModel.url == self.values['repo_url']) is not None:
             self._set_status(self.tr('This repo has already been added.'))
             return False
-
```
Hofer-Julian: I really think we should move to something like black at some point.
The manual work to make flake8 happy is annoying to everyone^^

m3nu: Can do. There is also autopep8, which is slightly less aggressive than Black. I use that mainly.

Hofer-Julian: Fine as well, it just needs to be aggressive enough to satisfy flake8.
flake8 should probably still be in our CI, since it not only finds style errors :)

Do you have experience with flake8+autopep8?
I know that black and flake8 play well together as long as you force flake8 to accept lines with length <88

m3nu: After autopep8, all code should pass flake8.

For the last new project I used autopep8 + pylint&flake8. The latter two kind of complement each other. But flake8 by itself is also fine.

Hofer-Julian: Perfect!

Hofer-Julian: Just looked up, how to integrate autopep8 into our CI, but I don't get it.
I would have suspected that autopep8 has something similar to ```black --check```, but I can't find it.
@m3nu if you know how to do it, please open the PR :)

m3nu: I think a developer would run autopep before committing his code. It doesn't run automatically, since the changes it makes need to be reviewed.

Hofer-Julian: Absolutely, but the CI should fail if autopep *would* have changed something.
Otherwise, the following PRs will include changes by autopep which have nothing to do with you

m3nu: True. That's probably possible. Would be surprised if not.

Hofer-Julian: Agree, if you find it, please tell me.
Otherwise, I'd like to open a PR for black :)

m3nu: This would return exit code 2, if changes are made, but not change anything.

`autopep8 --exit-code --diff tests/*`

But I'm not sure 3 tools are needed to report errors. And since autopep8 fixes errors reported by [pycodestyle](https://github.com/PyCQA/pycodestyle), that and flake8 may be enough to run automatically. And maybe pylint because it checks some extra stuff. But not sure how big the difference is.

Hofer-Julian: Does autopep8 currently change anything, when being run on Vorta?

m3nu: I would guess no, since the code passes flake8.

samuel-w:
Sorry about the mess of commits.
Last few commits was to support [samuel-w/vorta/KeyManagement](https://github.com/samuel-w/vorta/tree/KeyManagement) password checks

src/vorta/views/repo_add_dialog.py
Hofer-Julian: Just checked, it indeed changes nothing :)

samuel-w:
Can you change the branch to https://github.com/samuel-w/vorta-1/tree/Partial-385? Thanks.

Hofer-Julian:
> Can you change the branch to https://github.com/samuel-w/vorta-1/tree/Partial-385? Thanks.

I don't know how to do that.
Do you @m3nu ?

m3nu:
I think the branch for this PR was renamed or removed. Usually best to reopen a new PR. I can't change it either.

samuel-w:
I don't have the permissions to close this PR, I'll open a new PR once this gets closed.

Issue #436, opened 2020-04-07
Don't override FUSE mount permissions


smartm0use:
Hello, I'm using Vorta (and borg) for the first time.
I have installed the latest version (0.6.24) and I have made a couple of backups.
Now I have two rows in the Archive section and I'm trying to mount any of them, but it seems not to work.

I click on "mount" and I select the folder, then nothing happens, I get this text:
```
fuse: unknown option `umask=0277'
```

After I have selected the folder if I double click on the archive it opens the folder but it is empty.

I'm on Ubuntu 18.04.4 LTS.

m3nu:
Is your FUSE to old or too new to have this option?

This option makes sure that you can read the files, even though the backup was previously done by another user with restricted perms.

smartm0use:
Do you know how to remove that option? I do not need it...

smartm0use:
I have removed the version I installed with pip and I have installed Vorta again with flatpack.
Now I do not have that error, it now says "Mounted successfully" but when I double-click on the folder I get this error message:

![image](https://user-images.githubusercontent.com/5399265/79155502-7dc63500-7dd1-11ea-9bd8-b282f103ae77.png)

What does it mean? What's wrong now?

ThomasWaldmann:
https://github.com/borgbackup/borg/issues/5064 maybe?

Hofer-Julian:
The flatpak uses llfuse-1.3.6.
@ThomasWaldmann could it still be related?

Hofer-Julian:
The flatpak uses llfuse-1.3.6.
@ThomasWaldmann could it still be related?

smartm0use:
Running `pip3 search llfuse` on my system I get this:

```
llfuse (1.3.6)            - Python bindings for the low-level FUSE API
  INSTALLED: 1.3.3
  LATEST:    1.3.6
gridfs-fuse (0.1.2.dev1)  - A FUSE wrapper around MongoDB gridfs using python
                            and llfuse.
```

ThomasWaldmann:
"Transport endpoint not connected" is a not very helpful error message.

When running `borg mount -f ...` in foreground one usually also sees some traceback that makes clear why the borg mount process has died (which is the reason for "not connected").

Without that traceback one can only guess...

smartm0use:
As i said using `borg mount` does work, with no errors.
Doing the same with Vorta gives me that error.

m3nu:
[...]

PR #522, opened 2020-06-19
Adds a checkbox for mount option override, fixes #520

### LINKED ISSUE #520: Use different permissions when mounting archive.
Hi, I'm running Vorta 0.6.26 and Borg 1.1.13 on openSUSE Tumbleweed.

I just discovered a really weird anomaly with Vorta. I tested a backup of my Firefox `~/.mozilla` directory by using Vorta to mount the repo to `~/Desktop/a`, renaming my `~/.mozilla` folder and using Dolphin to copy it from `~/Desktop/a/home/myself/.mozilla` to its normal location. This consistently led to a corrupt Firefox profile, as described [here](https://support.mozilla.org/en-US/kb/fix-bookmarks-and-history-will-not-be-functional). I don't understand all the moving parts in Firefox, but it's either a permissions problem or some other kind of corruption with one of the sqlite databases it uses.

By simply creating a filesystem copy of the `~/.mozilla` profile on my local disk and closing Firefox and renaming the original profile and copying back the backup, Firefox launches as expected as if nothing happened.

So I tested again with the same process as described above, except I manually created the backup and manually mounted it with `borg create` and `borg mount`, and repeated the profile restore process. This also worked as expected.

Then I manually mounted with `borg mount` a backup created with Vorta and restored my profile from that, which also worked correctly.

Then I used Vorta to mount the manually created borg backup described above, and this gave me a corrupted profile.

So Vorta must be doing something weird with the way it's mounting the backups. I confirmed this happens every time I tried to mount it via Vorta and restore the profile, both from a USB HD repo and from my NAS over SSH.

Thanks for the help.

samuel-w:
There are 3 issues with this pull:
1. Adding the checkbox adds clutter to the UI
2. What should the default setting should be, as either setting can break permissions.
3. Does the average user knows what mount permissions are? Possible solution would be add a tooltip along the likes of "Toggle this if any permission denied errors pop up".
I'd like some input before going further, mostly on issue 2

### DIFF
--- a/src/vorta/assets/UI/archivetab.ui
+++ b/src/vorta/assets/UI/archivetab.ui
@@ -179,6 +179,16 @@
         </layout>
        </item>
        <item>
+        <widget class="QCheckBox" name="overrideMountOpts">
+         <property name="text">
+          <string>Override mount permissions</string>
+         </property>
+         <property name="checked">
+          <bool>true</bool>
+         </property>
+        </widget>
+       </item>
+       <item>
         <widget class="QLabel" name="mountErrors">
          <property name="enabled">
           <bool>true</bool>
--- a/src/vorta/borg/mount.py
+++ b/src/vorta/borg/mount.py
@@ -8,7 +8,7 @@ class BorgMountThread(BorgThread):
         self.updated.emit(self.tr('Mounting archive into folder...'))
 
     @classmethod
-    def prepare(cls, profile):
+    def prepare(cls, profile, overrideMountOpts):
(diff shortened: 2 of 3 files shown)


samuel-w:
After thinking about it, I think the default behaviour should be more like unmodified `borg mount` command, so I'll make the checkbox unticked by default

samuel-w:
Not sure if its state should be saved or not.

samuel-w:
Also closes #436

samuel-w:
Replaced by #682

Issue #274, opened 2019-05-02
Add custom options to borg create command


rwaldspur:
I'm looking to add additional arguments to the borg create command from within Vorta, for example, two simple examples:
--exclude-caches
--files-cache=ctime,size

I see there are additional arguments that can be passed while initializing a new repository under the Advanced tab, but my understanding is that there is no current way to add arguments to the "create" command.

This would be (somewhat) similar to issue #96 

I believe this could extend the functionality of Vorta quite a bit in a somewhat general way.

Thanks.

m3nu:
True. Currently there is an option to add arguments to *all* commands. Like global SSH options.

Args for `--create` could go under *Schedule > Shell Commands*. Not sure this is the best place.

flip-bs:
+1 It can be very useful to avoid backing up mounted remote FS with the option `--one-file-system`

adatum:
> Currently there is an option to add arguments to all commands.

@m3nu Is this explained somewhere? 

Options like `--dry-run`, `--one-file-system`, `--exclude-cache` are commonly used and very useful.

mahikeulbody:
and also `--chunker-params` create option.

Ramblurr:
This may be another feature, but I would love to see the ability to add `--dry-run` to test out my backup. Specifically I'm interested in testing my exclusion settings.

PitBullCH:
Looking forward to seeing this one - just had the case myself where I need to add --remote-path=borg1 to the Borg args for Rsync.net, but the only way to do it is delete the profile, unlink the repository, then new profile and re-use existing repository and then can access the Borg args input.

PR #530, opened 2020-06-28
Add create options

samuel-w:
I have an issue before I continue on this:
Should I create a new tab for just the options, as the putting it in the schedule tab as suggested in #274 makes the schedule tab cluttered and create options aren't really part of scheduling, but I also don't want to end up with an interface with 10 tabs

Now that I think about it, I could rename the new tab to Backup Options, and move the shell commands there as well

Looking for input, thanks

### DIFF
--- a/src/vorta/assets/UI/createtab.ui
+++ b/src/vorta/assets/UI/createtab.ui
@@ -1,0 +1,181 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>Form</class>
+ <widget class="QWidget" name="Form">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>791</width>
+    <height>497</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Form</string>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout_2">
+   <item>
+    <widget class="QToolBox" name="toolBox">
+     <property name="currentIndex">
+      <number>0</number>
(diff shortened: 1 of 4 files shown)


samuel-w:
initialui
This is my current UI setup. Should I break it into tabs like it appears in the [documentation](https://borgbackup.readthedocs.io/en/stable/usage/create.html)?

m3nu:
It could be good to have those options somewhere hidden under advanced, but Vorta's goal is not to replicate every Borg option 1:1.

Some of the options in your screenshot are also related to the source tab/exclusions and should go there if they are added.

I also don't think that we should add a new tab to mirror a specific Borg subcommand. A GUI will present information differently than a CLI tool. By that logic we should also move the *Sources* tab to *Create*, which would worsen the UI.

So for each option you should consider, *if* it's needed in the GUI and *where* it should go. E.g. we had very few requests to support storing metadata change times or whatever. 😁

In addition the way you set up the changes, no options are ever saved. So they would only persist until the window is closed. That may be confusing.

To summarize, yes we can add new options if they add value to a significant group of users and are integrated with the UI structure. But we won't implement every single option the Borg CLI provides.

m3nu:
Let me also link to this piece, since it kinda fits here:

m3nu:
Related: https://blog.codinghorror.com/this-is-what-happens-when-you-let-developers-create-ui/

😜

samuel-w:
I didn't implement saving just in case I had to change things. Going to reconsider my approach. Thanks.

samuel-w:
I'll probably just stick a button under the excludes in the sources tab

Issue #532, opened 2020-06-29
Repository Space Warning


hutson:
**Is your feature request related to a problem? Please describe.**

I recently added a new directory to the _Sources_ list that would exceed the 10GB limit available in the free plan on BorgBase to test the behavior of Borg+Vorta+BorgBase.

During the upload process Vorta displayed the following error:

![Screenshot_20200628_192022](https://user-images.githubusercontent.com/6701030/85965913-e90cb580-b983-11ea-9db2-d4f480ce153e.png)

First, there was no indication that adding the new directory would exceed the available storage space. Had this not been intentional I may have been in a situation where I added a directory to be backed up only to find out hours/days later that it had not been successfully uploaded. 

Second, when I looked at the _Archives_ page there was no backup available for today. I would have expected to see a list of files that _were_ successfully uploaded up to the point when space ran out. Unfortunately it appears as if nothing was backed up. Had this been a 401GB directory on a 100GB plan, I _think_ the backup would have failed at 400GB leaving nothing to show for the time and effort to upload.

![Screenshot_20200628_192120](https://user-images.githubusercontent.com/6701030/85966403-48b79080-b985-11ea-8aff-30e3dc569831.png)

**Describe the solution you'd like**

There are several ways activity may create more data than can be accommodated by the remote server. For this particular case, is it possible for Borg to pre-calculate the space that will be consumed by the backup and then check with the remote repository to verify space is available before beginning the backup?

In another backup tool I use, this is the default behavior, and disabling that behavior is an application preference:

![Screenshot_20200628_204939](https://user-images.githubusercontent.com/6701030/85967162-700f5d00-b987-11ea-85ce-6bd7e07e96f7.png)

I believe Borg maintains a lock on the remote repository so that it's not possible for another backup process to consume the space before the backup is complete. However, this may not matter in the case where the remote repository is sharing a single account with multiple repositories that count against the same disk space quota.

samuel-w:
I don't think there is a way for Borg to pre-calculate the space, as deduplication and compression reduce storage usage by a lot (For me, 200GB turns into 30GB). There is a way to disable space calculations, but its a bit hacky IMO https://github.com/borgbackup/borg/commit/43acbb92dabf08b7cae9e7a264c06c92194a484c.

ThomasWaldmann:
borg does checkpoints now and then, default interval is (iirc) every 1800s.

have a look using `borg list --consider-checkpoints REPO`.

[...]

PR #537, opened 2020-06-30
Add disk space override button, fixes #533

### LINKED ISSUE #533: Can't Diff or Mount Archives While Backup Running
**Describe the bug**

I wanted to demonstrate how to run a diff between two archives, but after pressing the _Diff_ button on the diff dialog, nothing happened (The dialog vanished but no diff was shown).

I suspect this is because I am currently running a backup job.

**To Reproduce**
1. Press _Start Backup_.
2. Click on the `Diff` button on the _Archives_ tab.
3. Chose two archives to diff.
4. Press `Diff` and observe that the dialog disappears and no diff is shown.

 - OS: Kubuntu
- Installation method: Flatpak

samuel-w:
Since configuration is stored on a per-repository basis, I had to use a different method of saving and loading the config options. I think the way this is the best option to do it, other than reading and parsing the raw config text. 

I had to exclude the config thread as it has to load the state on startup, otherwise Vorta thinks a backup is running when its not.

### DIFF
--- a/src/vorta/assets/UI/misctab.ui
+++ b/src/vorta/assets/UI/misctab.ui
@@ -22,6 +22,16 @@
      <property name="topMargin">
       <number>10</number>
      </property>
+     <item>
+      <widget class="QCheckBox" name="overrideFreeSpace">
+       <property name="toolTip">
+        <string>Does not work on remote repositories for security reasons</string>
+       </property>
+       <property name="text">
+        <string>Disable free space calculations during backup</string>
+       </property>
+      </widget>
+     </item>
     </layout>
    </item>
    <item>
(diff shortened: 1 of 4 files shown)


m3nu:
Have you seen the `SettingsModel` class? We keep global configs there. https://github.com/borgbase/vorta/blob/master/src/vorta/models.py#L165

samuel-w:
But the configuration is stored in the repository, so that will not work.

m3nu:
What are you trying to achieve again? To avoid duplicate work, you should start with a feature suggestion issue to describe the problem, solution and suggested implementation. Then, when there was sufficient discussion link that to an actual PR with code.

samuel-w:
I am working on issue #532, and one of the requests is to add an option to disable free space calculations.

PR #540, opened 2020-07-02
Add KWallet support

### LINKED ISSUE #236: Support KDE KWallet
We have already implemented macOS Keychain and Gnome Keyring to store passwords. KWallet should be added next.

Some inspiration:

- https://gist.github.com/eberle1080/1304965
- https://stackoverflow.com/questions/2147724/using-kwallet-in-pyqt4

samuel-w:
In the future, we should check if the wallet closes in between detecting the current keyring and setting the password.

Also, is there an alternate way to do keyring detection other than a mountain of try excepts? The code is starting to get hard to read.

Fixes #236


samuel-w:
Wait, QtDBus exists. Have to change code

samuel-w:
Attempting to port to QtDBus, unsucessful.
Python crashes with "No such method 'open' in interface 'org.kde.KWallet' at object path '/modules/kwalletd5' (signature 'sis')"
Even when I convert the wId to qlonglong, it still gets converted back to an int.
However, the existing dbus code still works.
https://gist.github.com/samuel-w/005453d30250265e2c75c167d9b545b3

Hofer-Julian:
Too bad...
But thanks for trying :)

samuel-w:
Tested and it works, woohoo. Had to use command line to get the handle.

m3nu:
Dont want to add this any more? I’m a bit behind merging PRs due to many other things going on, but plan to catch up over Aug.

rejinka:
I really would like to get this feature

samuel-w:
Initially closed this because I thought no one would use it, as more code = more maintenance. Guess I was wrong.

m3nu (dismissed):

samuel-w:
Wait a bit, I may have forgot to catch an exception

samuel-w:
Turns out there is no uncaught exception, but I did find a bug #606

samuel-w:
What's the status of this?

m3nu (approved):

rejinka:
Thank you very much @samuel-w and @m3nu!

Issue #539, opened 2020-07-01
Only run backups when battery is full


ktosiek:
**Is your feature request related to a problem? Please describe.**
On a slow or metered connection backups can take most of the bandwidth.

**Describe the solution you'd like**
It should be possible to skip scheduled backups if the available connection is known to be metered, or even when it's not known to be unmetered. A running backup should be stopped if we lose a suitable connection.

**Describe alternatives you've considered**
The only alternative I know is enabling notifications and stopping the backup when I know I don't want it to run, which is not very handy.

m3nu:
The Wifi settings already attempt to solve this problem. Under *Schedule > Networks > Allowed Networks*. Select only "unmetered" networks to use for automatic backups.

samuel-w:
But the button is not visible if you don't have Wifi, yet it is possible to have a metered wired connection.

ktosiek:
I think that's only implemented for OSX.

On Linux with NetworkManager you can mark connections (of any type, not just WiFi) as metered, and this is honoured by some applications. NetworkManager even tries to guess that status for active connections (for example tethering through Android is marked as "guessed as metered").

The options I feel would make sense are:
* Use any connection
* Only use unmetered connections (which would be a good default, but needs a warning in the UI saying why scheduled backups are paused)
* Use explicit WiFi whitelist (but what about Ethernet connections then? Are those listed too on OSX?)

m3nu:
Using NetworkManager to grab Wifi details on Linux came up before and would be a nice addition.

ktosiek:
How does the networks list work on OS X? Is it just WiFi networks? Will backup run over cable when some WiFi network is selected from the list?

samuel-w:
Communication with NetworkManager takes either another dependency https://pypi.org/project/python-networkmanager/ or DBus. Which one is prefered?

m3nu:
If DBus is not too hard to implement, I'd rather avoid more deps. Else mark it as Linux-only in `setup.cfg`.

And before writing any code, feel free to summarize your planned implementation/changes.

ktosiek:
I think we don't need python-networkmanager: the most interesting part for us is the SSID decoding function (NetworkManager provides raw SSIDs), and the library uses a different DBus implementation than Vorta already uses for notifications.

For the UI I'd just add a "Don't run backup over metered networks", enabled by default, and visible only when NM is available.

The actual implementation should have two parts:
1. An additional check in BorgCreateThread.prepare
[...]

PR #547, opened 2020-07-12
Add an option to stop backups from running on metered networks

ktosiek:
First part of #539 - a backup won't start on a metered connection, but won't stop if the connection status changes later.


samuel-w:
`self.dontRunOnMeteredNetworksCheckBox.setCheckState(QtCore.Qt.CheckState.Checked if profile.dont_run_on_metered_networks else QtCore.Qt.CheckState.Unchecked)` can become         self.dontRunOnMeteredNetworksCheckBox.setChecked(profile.dont_run_on_metered_networks)

samuel-w:
Additionally, flatpak seems to block system dbus access. You need to add `"--system-talk-name=org.freedesktop.NetworkManager",` to flatpak/com.borgbase.Vorta.json

Hofer-Julian:
I'll add the flatpak flag as soon as this PR is merged

m3nu:
Mainly note to self: It would make sense to put macOS/Linux-specific network utility functions as separate modules. Like `utils/{linux,darwin}/network{_status}.py`.

For macOS we could fake the looking for "AndroidAP" in the network name. Anyone know the default iOS SSID? Or how Li

ktosiek:
On Linux with NetworkManager users can explicitly mark network as metered, but most of the time NM just guesses. And that guessing involves a few heuristics:
1. It's a Bluetooth or GSM/CDMA connection.
2. The Wi-Fi declares it's metered by setting a Microsoft's Network cost information EID.
3. The DHCP server sends an ANDROID_METERED vendor-specific extension - an Android phone connected by USB will only set this if it's sharing the GSM connection, but not for unmetered Wi-Fi (Android also has a concept of metered Wi-Fi).

Links to NM code:
1. Most of the guessing logic lives in [nm_device_update_metered](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/f0a39b517e06470c4b176ef0bbd49c2eef1d7ad7/src/devices/nm-device.c#L14890)
2. The Network cost information is [extracted from Wi-Fi beacon](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/f0a39b517e06470c4b176ef0bbd49c2eef1d7ad7/src/nm-core-utils.c#L4546)
3. [Extracting the ANDROID_METERED](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/f0a39b517e06470c4b176ef0bbd49c2eef1d7ad7/src/dhcp/nm-dhcp-systemd.c#L451)

ktosiek:
I've fixed `setCheckState` and the missing flatpack flag.

ktosiek:
There's a fix in my branch that I can't see in the PR - I think GitHub is not feeling well?

Hofer-Julian:
I currently have the exact same problem for a different project...

ktosiek:
I've merged master in to trigger the CI, as it was stuck for a week. It's all green now.

m3nu:
Can we make the feature more OS-agnostic and make the public functions in `src/vorta/network_status.py` less Network Manager-specific? Similar to the way we implement the `keyring` module.

That way the feature doesn't need to be hidden and is always available, but just implemented differently. Until someone looks at the details, it could just look for `AndroidAP` in the SSID.

ktosiek:
An indirection layer like in the `keyring` module makes sense, I'll try adding that.

I'm not sure if we want to invent new concepts on OSes that don't have a "metered network" concept - it sounds confusing. I've just checked 3 different Android devices, and none of them uses `AndroidAP` as the default SSID.
I've looked into how NM guesses if a network is metered, and none of the rules depend on SSID (see https://github.com/borgbase/vorta/pull/547#issuecomment-658573925 a few comments earlier).

m3nu:
Thanks.

Did you see any other pattern in your Android hotspot names? Else macOS can just always return a hardcoded value until we figure something out or they implement it. I'm afraid for now we only have the SSID as data point.

ktosiek:
From what I've seen the default SSID is a name of the device, series, or manufacturer (like "OnePlus", "Redmi", or "ALCATEL ONE TOUCH POP7"). So that's not very helpful :-C

m3nu:
I was able to get the `ANDROID_METERED` DHCP option on macOS:

Normal router:
```
$ ipconfig getpacket en0
op = BOOTREPLY
htype = 1
flags = 0
hlen = 6
hops = 0
xid = 0x8dc8db4d
secs = 0
ciaddr = 0.0.0.0
yiaddr = 172.16.13.237
siaddr = 0.0.0.0
giaddr = 0.0.0.0
chaddr = 8c:85:90:ad:ee:a3
sname =
file =
options:
Options count is 9
dhcp_message_type (uint8): ACK 0x5
subnet_mask (ip): 255.255.252.0
router (ip_mult): {172.16.12.1}
domain_name_server (ip_mult): {172.16.12.1, 8.8.8.8}
domain_name (string): .
lease_time (uint32): 0xe10
interface_mtu (uint16): 0x5dc
server_identifier (ip): 172.16.12.1
end (none):
```

My phone (note the last line)
```
manu@nyx2: ~/Downloads $ ipconfig getpacket en0
op = BOOTREPLY
htype = 1
flags = 0
hlen = 6
hops = 0
xid = 0x8dc8db4e
secs = 0
ciaddr = 0.0.0.0
yiaddr = 192.168.43.223
siaddr = 192.168.43.242
giaddr = 0.0.0.0
chaddr = 8c:85:90:ad:ee:a3
sname =
file =
options:
Options count is 11
dhcp_message_type (uint8): ACK 0x5
server_identifier (ip): 192.168.43.242
lease_time (uint32): 0xe0f
renewal_t1_time_value (uint32): 0x707
rebinding_t2_time_value (uint32): 0xc4d
subnet_mask (ip): 255.255.255.0
broadcast_address (ip): 192.168.43.255
router (ip_mult): {192.168.43.242}
domain_name_server (ip_mult): {192.168.43.242}
vendor_specific (opaque):
0000  41 4e 44 52 4f 49 44 5f  4d 45 54 45 52 45 44     ANDROID_METERED
```

ktosiek:
I've moved getting network status into a separate classes: one for OSX, one for NetworkManager, and one dummy class when none of those is available. I've also moved listing WiFi networks into those classes, I hope to implement listing networks for NM shortly.

ktosiek:
I'm not sure what the test failure is really about, I've seen it happen on ubuntu instead of macos for me: https://github.com/ktosiek/vorta/actions/runs/207355280.
Now both Darwin and NetworkManager implementations support guessing if network is metered and listing wifi networks.

m3nu:
Works as expected on macOS. Just fixed some issues with string type.

Sometimes tests fail due to some DB race conditions that aren't fully fixed yet. This time they time out, which is new. Will have a closer look.

m3nu:
Excluded macOS/Darwin tests to not run on Linux. Is that OK? 😜

ktosiek:
The tests in `test_darwin.py` should be fully isolated from the environment, that's why I've moved calling commands to separate `call_*` functions. Running those tests on Linux was my only way of testing the new code, so they shouldn't need to be skipped :-)

Similarly with the NM tests - they should work as long as QtDBus is available, but shouldn't need access to NetworkManager, or even a working bus.

m3nu:
OK. Unskipping.

PR #548, opened 2020-07-15
Add profile backup/restore functionality

### LINKED ISSUE #462: Make Vorta profiles easy to backup and restore
First -- ***thank you very very much*** for making such great free/libre software! I've used GNU/Linux for about ten years, and whenever I looked at backup solutions, they all seemed too simple or way too complex. Vorta gives me secure backups with detailed control, sane defaults and an easy to understand interface. A delightful creation!

Summary:
**I want Vorta to let me save profiles and then load them on a different system.**

Details:
I just used Vorta to backup my system and user files, then I erased my disk. Now I've installed a fresh operating system and installed Vorta. I open Vorta, and want the option to load my old profiles, but that option doesn't exist. So I create a new profile and "Add Existing Repository" and give it the location and passphrase, and it successfully loads the archives. But now, **if I want to make a new backup, I have to recreate my old profile!** In this case, the most tedious and error-prone part is the Sources tab -- can I remember exactly what sources and exclude patterns I used to create these backups? If I can, great. If not, time for another round of trial and error and honing the details.

Where are my profiles? I looked for the profiles, both on my new system and the mounted backup, and I don't see them anywhere. I see `~/.config/borg/security` has the same number of directories as I had profiles, but the files don't have my sources list. **If I could find my profiles, that would be a first step towards backing them up and loading them.** As it is, Vorta appears to function in a way that I can't back up my backup config :). This means that I can't integrate it with config managers either, like `nix`, `guix`, or `aconfmgr`.

samuel-w:
<del>
This is mostly complete, I just need to add importing a file generated in an old schema version

For schema version upgrades, there are 2 approaches I could take:

1. Create new temporary database, upgrade it to current schema, and merge the two

    -  Can upgrade easily using code from init_db, difficult to implement initially but requires less maintenance

2. Modify the dictionary in memory, then save normally

   - Requires writing new upgrading code twice, easier to implement initially but reduces code quality and needs more maintenance
  - Changes are only needed for column deletion and in place data changes, column addition can use default values

I don't want to exclude options in between schema versions, as it will be confusing when part of the profile is imported and part is not.

</del>

The only way to do schema upgrades is the second way, since dict_to_model will not work with unknown entries.

Also I need to know if WifiSettingModel backup/restore works.

Fixes #462 

Discussion #834


Hofer-Julian:
Sorry about the closed/reopened PR.

I see this PR is still marked as a draft. Is it ready to be reviewed?

m3nu:
> I see this PR is still marked as a draft. Is it ready to be reviewed?

I'll go first. But we'll need some discussion on the way, since it's a complicated feature.

samuel-w:
I left it as draft so I can get opinions on the approach for schema version upgrades. Currently, it only works when the local schema version is the same as the file version.

I think I'll take approach 1, as merging databases doesn't seem easy, and 2/3s of the schema upgrades so far are column additions which seem to be automatically set to default values when using dict_to_model.

samuel-w:
ui
Currently the additional actions button has the default icon. What should this be changed to, because the icon does not really describe the contents of the menu.

samuel-w:
Is there a need for tests? The only way I can see to implement them is to have pregenerated profiles to import as test cases, but those will need to be updated every time there is a schema version update.

samuel-w:
Did you test any other PR that changed the schema version? That might be the cause of the error, since the message only triggers when current_schema < backup_schema. Will add the icons.

samuel-w:
Found a bug with the menu, its 4+ years old https://bugreports.qt.io/browse/QTBUG-50782 https://bugreports.qt.io/browse/QTBUG-49663
Happens on multi screen setups.
Peek 2020-09-05 16-35

The schema upgrade failure message should only display if the upgrading code in `json_to_profile` is wrong

Merge this after #551, otherwise I'll have to convert the backup dates with 80+ lines of code and that will suck.

samuel-w:
What is preventing this from being merged?

samuel-w:
8 month anniversary! This is probably not going to get merged.

PR #550, opened 2020-07-16
Improve password handling

### LINKED ISSUE #385: Passphrase handling improvements

samuel-w:
Changed branch, old PR #518 

d7bb56281c64d190b34600da2421945dacf80478 relates to  #540, but should work without it

How should I format my code for easier translation later?

Fixes #385


src/vorta/utils.py
```diff
@@ -273,3 +273,19 @@ def is_system_tray_available():
...
+    passEqual = firstPass == secondPass
+    passLong = len(firstPass) > 8
+    valid = passEqual and passLong
+
+    if not passEqual:
+        msg += "identical"
+    if not (passEqual or passLong):
+        msg += " and "
+    if not passLong:
+        msg += "greater than 8 characters long"
```
ThomasWaldmann, line 289: that might work for english or languages with similar word order, but will not work for other translations.

you can't assemble from pieces when considering translation.

src/vorta/views/repo_add_dialog.py
```diff
@@ -44,6 +48,22 @@ def values(self):
...
+            messages = {
+                'VortaDBKeyring': 'plaintext on disk.\nVorta supports the secure Secret Service API (Linux) and Keychain Access (macOS)',  # noqa
+                'VortaSecretStorageKeyring': 'the Secret Service API',
+                'VortaDarwinKeyring': 'the Secret Service API',
+                'VortaKWallet5Keyring': 'KWallet5',
+                'VortaKWallet4Keyring': 'KWallet4'}
+            # Just in case some other keyring support is added
+            keyringName = messages.get(keyringClass,
+                                       'somewhere that was not anticipated. Please file a bug report on Github')
+            self.passwordLabel.setText('The password will be stored in ' + keyringName)
```
ThomasWaldmann, line 63: same here. use a placeholder.

samuel-w: Something like 
`self.passwordLabel.setText(f"The password will be stored in {messages.get(keyringClass, 'somewhere that was not anticipated. Please file a bug report on Github')}")`?

src/vorta/utils.py
samuel-w: Would this work for translation?
`

    if not passEqual:

        msg = "Passwords must be identical"

    if not passLong:

        msg = "Passwords must be greater than 8 characters long"

    if not (passLong or passEqual):

        msg = "Passwords must be identical and greater than 8 characters long"
`

samuel-w:
Part of #385

samuel-w:
The password line size changes drastically when visible/hidden
Peek 2020-10-08 20-26
Should I keep the option?

m3nu:
Screen Shot 2020-10-09 at 09 57 22

macOS looks normal. I assume Linux had the oversized circles before?

samuel-w:
Yeah, I think its a weird Linux thing. I'll keep it, since its going to be kept in one mode or the other.

samuel-w:
What's the status of this PR?

src/vorta/utils.py
```diff
@@ -318,3 +320,35 @@ def is_system_tray_available():
...
+        return trans_late('utils', "Passwords must be identical and greater than 8 characters long.")
+    if not passEqual:
+        return trans_late('utils', "Passwords must be identical.")
+    if not passLong:
+        return trans_late('utils', "Passwords must be greater than 8 characters long.")
+
+    return ""
+
+
+def password_transparency(encryption):
```
m3nu, line 339: This is good.

m3nu: Maybe there could be a clearer name for this function? Like `display_password_backend()`

src/vorta/utils.py
```diff
@@ -318,3 +320,35 @@ def is_system_tray_available():
...
+    if encryption != 'none':
+        keyringClass = get_keyring().__class__.__name__
+        messages = {
```
m3nu, line 342: How about moving the Keyring name into the respective Keyring class instead of having a mapping here? The Keyring name itself wouldn't need translation, since it's common to use the English name.

src/vorta/utils.py
```diff
@@ -318,3 +320,35 @@ def is_system_tray_available():
...
+            'VortaDBKeyring':
+            trans_late('utils', 'plaintext on disk.\n'
```
m3nu, line 344: Suggested text for DB fallback could be like this: "Saving password in Vorta settings database. To store password more securely install a supported secret store."

m3nu: (This would be the description text in the `VortaDBKeyring` class)

src/vorta/views/repo_add_dialog.py
```diff
@@ -59,6 +79,27 @@ def receive():
         dialog = choose_file_dialog(self, self.tr("Choose Location of Borg Repository"))
         dialog.open(receive)
 
+    def set_password(self, URL):
```
m3nu, line 82: This is nice. So it will pre-fill saved passwords?

src/vorta/views/repo_add_dialog.py
```diff
@@ -124,21 +165,44 @@ def validate(self):
...
 
-        if self.__class__ == AddRepoWindow:
-            if self.values['encryption'] != 'none':
-                if len(self.values['password']) < 8:
-                    self._set_status(self.tr('Please use a longer passphrase.'))
-                    return False
-
         return True
 
+    def password_listener(self):
```
m3nu, line 170: Again, we could consider simplifying this PR by only validating the password on submission, not in real time.

m3nu (changes requested):

m3nu:
I added some comments. Can be merged with minor changed, I think. Good PR really. Extra points for adding tests. 😄

src/vorta/utils.py
samuel-w: At least for macOS the names are translated https://support.apple.com/fr-fr/guide/keychain-access/kyca1083/mac

src/vorta/utils.py
```diff
@@ -273,3 +274,41 @@ def is_system_tray_available():
         is_available = tray.isSystemTrayAvailable()
 
     return is_available
+
+
+def validate_passwords(firstPass, secondPass):
```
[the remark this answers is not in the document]
samuel-w: When we implement password changing in the future this will need to be reused. Websites do password validation in real time, so it would be nice to have that here as well.

src/vorta/views/repo_add_dialog.py
samuel-w: The code needed to validate passwords is just two lines, not complex at all.
```
        self.passwordLineEdit.textChanged.connect(self.password_listener)
        self.confirmLineEdit.textChanged.connect(self.password_listener)
```

src/vorta/utils.py
samuel-w: I prefer having it centralized in one location, instead of importing the translations in all the keyrings.

src/vorta/views/repo_add_dialog.py
samuel-w: Yes, but only if the current entered password is empty.

src/vorta/utils.py
samuel-w: I prefer not to mention databases, just how its inseecure.

src/vorta/utils.py
m3nu: Also possible. Then keep the English name as attribute of each Keyring class and translate it where you need it. No importing of translations in every Keyring.

src/vorta/utils.py
m3nu: Then "Saving password with Vorta settings. To store password more securely install a supported secret store."

src/vorta/views/repo_add_dialog.py
m3nu: Fine. More interactive.

src/vorta/utils.py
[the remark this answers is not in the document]
m3nu: Good call. I didn't notice it's in utils.

src/vorta/utils.py
samuel-w: That seems fine to me

src/vorta/utils.py
samuel-w: Yeah, that seems more accurate.

src/vorta/utils.py
```diff
@@ -318,3 +320,35 @@ def is_system_tray_available():
...
+    if encryption != 'none':
+        keyringClass = get_keyring().__class__.__name__
+        messages = {
+            'VortaDBKeyring':
+            trans_late(
+                'utils', 'Saving the password to disk. To store password more securely install a supported secret store such as KeepassXC'),
+            'VortaSecretStorageKeyring': trans_late('utils', 'Storing the password in the Secret Service keyring.'),
+            'VortaDarwinKeyring': trans_late('utils', 'Storing the password in in Keychain Access.'),
+            'VortaKWallet5Keyring': trans_late('utils', 'Storing the password in in KWallet 5.'),
+        }
```
samuel-w, line 350: This uses full sentences to make translation easier.

src/vorta/utils.py
samuel-w: That still separates the messages into each keyring, when it really shouldn't be there.

samuel-w: Do translations even work with variable strings?

samuel-w: `translate('utils', keyring.message)`
in VortaDBKeyring:
`message = "Saving the password to disk. To store password more securely install a supported secret store such as KeepassXC"`

Would that import `Saving the password to disk. To store password more securely install a supported secret store such as KeepassXC` to transifex?

samuel-w: I don't want to break the translations, and doing this seems like the easiest way.

samuel-w: A quick check of translations shows that translating variable strings does not work sadly. Have to go with the dictionary or the import in every keyring.
Screenshot_20201201_020459

samuel-w:
I think the keyring messages cannot be changed, see my comments earlier.  https://github.com/borgbase/vorta/pull/550#discussion_r533139509

samuel-w:
Actually, instead of an individual message for each keyring, should it just be one message for secure keyring and one message for insecure keyring, since most people use only one password manager at a time ?

m3nu (approved):

PR #552, opened 2020-07-18
CI Fix-Set 10 minute test job time

samuel-w:
Currently workflows can run for 6 hours before automatically being stopped. 
There is a limit of 5 macOS jobs at a time, while each action runs 3 macOS jobs, meaning that 2 mistakes can block macOS jobs for 6 hours. (I'm surprised no one else has done a similar thing before)
~~This wastes 360 minutes of run time while there are only 2000 minutes per month of time (I assume this is using the free plan).~~

For example, I accidentally pushed 3 infinite loops:
https://github.com/borgbase/vorta/actions/runs/174077306
https://github.com/borgbase/vorta/actions/runs/174058136
https://github.com/borgbase/vorta/actions/runs/174054779

~~Since the timeout is so high, I will end up wasting 1080 minutes of run time, while with this change I would have wasted only 30.~~

10 minutes is a decent compromise, as it allows for some code to be slow and for future expansion while mitigating infinite loops.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 
 jobs:
   test:
+    timeout-minutes: 10
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false


samuel-w:
After further research, the limit only applys to private repos. Sorry!

Hofer-Julian:
Huh, public repos habe

samuel-w:
Yeah, https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions says "GitHub Actions usage is free for public repositories.". Although, there is a limit of 5 macOS jobs at a time, which could be taken up by long running workflows, so this is useful after all.

m3nu:
I'd hope Github already has a limit, in case someone pushes something really stupid. But we can add this one too.

PR #563, opened 2020-08-04
Add correct icon for notifications

Hofer-Julian:
Also optimize icon

### DIFF
--- a/src/vorta/notifications.py
+++ b/src/vorta/notifications.py
@@ -78,7 +78,7 @@ class DBusNotifications(VortaNotifications):
         v = QtCore.QVariant(12321)  # random int to identify all notifications
         if v.convert(QtCore.QVariant.UInt):
             id_replace = v
-        icon = ""
+        icon = "com.borgbase.Vorta"
         title = header
         text = msg
         actions_list = QtDBus.QDBusArgument([], QtCore.QMetaType.QStringList)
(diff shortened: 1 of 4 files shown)


src/vorta/notifications.py
```diff
@@ -78,7 +78,7 @@ def _dbus_notify(self, header, msg, level='info'):
         v = QtCore.QVariant(12321)  # random int to identify all notifications
         if v.convert(QtCore.QVariant.UInt):
             id_replace = v
-        icon = ""
+        icon = "com.borgbase.Vorta-symbolic"
```
samuel-w, line 81: I tested this as "icon = 'com.borgbase.Vorta'" the notification showed the icon. Is there a need for a seperate icon?

Hofer-Julian: I wondered the same.
Our icon is already very simplistic.
But normally symbolic icons are only black-white

m3nu:
Can be merged, @samuel-w ? Waiting for your review.

samuel-w:
It can be merged, I was just confused over the need for a seperate icon, but I understand it now.

SergioRAgostinho:
I was looking into getting a notification image into linux's notifications and noticed this PR was merged.

This PR only works if vorta is installed from flatpack because it makes reference to assets which are only part of the flatpack package. If one tries to install vorta from
```
$ python3  setup.py install
```

this new icon is not shipped anywhere. Is calling 
```
$ python3  setup.py install
```
considered a supported method of installing from source?  If yes, be aware the icon notification is not showing in this case.

Hofer-Julian:
@SergioRAgostinho, installing via setup.py is supported.
Developers use it, distributions will use it and a few users as well.
That you don't see the image is not surprising, since setup.py does not take care of this (also before this PR has been merged).
You can copy them manually, but I would recommend to use the flatpak instead.

PR #564, opened 2020-08-05
Add auto install for applications and icons

samuel-w:
I had to break wheel because of "bdist_wheel makes absolute data_files relative to site-packages".
When testing this, do not install with `pip install -e .`, install with `pip install .`, otherwise the files will not be moved. This is apparently by design. 
`pip install -e .` works normally for other development purposes still, just it doesn't copy icons and desktop files.

Can be disabled by setting `DISABLE_ICONS_DESKTOP` to a non-zero value


setup.py
```diff
@@ -1,8 +1,44 @@
...
+
+DATA_FILES = []
+CURRENT_DIR = os.getcwd()
+
+# Install icons and desktop files, not sure if /usr/local/share is constant
+if sys.platform == 'linux' and os.environ.get("DISABLE_ICONS_DESKTOP", 0) == 0:
+    if 'bdist_wheel' in sys.argv:
+        sys.exit("Building wheels is disabled. This is not a bug")
+
+    if os.geteuid() == 0:
```
Hofer-Julian, line 13: Not sure what this is good for.

setup.py
```diff
@@ -1,8 +1,44 @@
...
     package_dir={'': 'src'},
+    data_files=DATA_FILES,
     package_data={'vorta.i18n': ['qm/*.qm']}
 )
+
+if sys.platform == 'linux':
+    # Unrename files
+    os.rename(os.path.join(CURRENT_DIR, 'package/com.borgbase.Vorta.svg'),
+              os.path.join(CURRENT_DIR, 'package/icon.svg'))
+    os.system("gtk-update-icon-cache")
```
Hofer-Julian, line 44: Is this then GNOME specific?

samuel-w: On second thought, this line is not actually needed. Might result in the icon not appearing until relog, but should be fine.

setup.py
samuel-w: This is because when installing from source `pip install .` pip tries to build a wheel for Vorta using `bdist_wheel`. When downloading from a pypi release it should not be an issue because the package is built as `sdist`.

Hofer-Julian: I've meant: if os.geteuid() == 0:
:)

samuel-w: That is to check if the user is installing it as root, and sets the local variable appropriately, since the default install directory when running pip as root is /usr/local/

Hofer-Julian (changes requested):
The necessary changes are noted in the comment above 😊

samuel-w:
On second thought, I should make the autostart copy the icon over instead of doing this, since its simpler.

PR #567, opened 2020-08-07
Use PyQt 5.15

samuel-w:
Ubuntu tests work, but macOS has test_add_folder fail. and that test alone.

Edit: Had to change the test to clicking the button using click() rather than qtbot for macOS. Everyting works now, yay.


Hofer-Julian:
macOS with Python 3.8 seems to be failing though...

m3nu:
Merged via #573.

PR #571, opened 2020-08-08
Add rename functionality

### LINKED ISSUE #570: Allow renaming for specific archive snapshots
If there's a snapshot in the archive list that is special for the user in some way, it would be very convenient to give it a unique name, instead of the generic, so it could be found very easy. Example names: "Before OS upgrade", "Before moving x" or whatever.

Add an option to rename specific archives. For example, add an additional button ("Rename") to the Extract, mount, check etc. group, maybe an addition binding to F2? The renaming could take place within the table control directly in the "Name" section, if that's possible.

Keep a textfile somewhere where you store the full generic name of specific backups, next to a meaningful description. I'm not even sure if that counts as a "real" alternative.

samuel-w:
Fixes #570 

This is starting to get a bit crowded. Should we split the left side into two lines? Thoughts?
crowded
Something like this? Although then the UI looks too complex
split_sides

Edit: I'll remove the button and try to make the name editable from the table.
Edit again: Actually, nevermind, since implementing renaming through the table will make the functionality vague.


XXXBold:
Wow, that was fast. Does this also implement renaming with the F2-Key?

samuel-w:
I didn't add it since the other functions do not have keybindings.

m3nu:
> Edit again: Actually, nevermind, since implementing renaming through the table will make the functionality vague.

How about adding archive-specific actions to the table as button with submenu, similar to what you did for import/restore profile? And only keeping repo-wide options at the bottom.

samuel-w:
OK, this is currently what I have:
menu
I had to change the tests to calling the methods directly, as qtbot doesn't work well with menus.

m3nu:
> I had to change the tests to calling the methods directly, as qtbot doesn't work well with menus. All buttons work normally.

It's fine to call the methods directly. Makes it more of a unit test than a full integration test (which is more prone to breaking due to small changes).

This is for right-clicks now? I'd add a button for this, since the right-click isn't totally obvious.

samuel-w:
Would a tooltip work as well? I don't want to put the same functionality in two different places.

m3nu:
I'm afraid, nobody every notices those tooltips.

But a button is fairly easy to add: https://stackoverflow.com/questions/12009134/adding-widgets-to-qtablewidget-pyqt

I can add some code, if you allow me to push to your branch. There should be a checkbox somewhere at the top of the PR that says "Allow contributors to push to this branch (or so)".

m3nu:
*Allow edits from maintainers*

allow-maintainers-to-make-edits-sidebar-checkbox

samuel-w:
I always set that checkbox. Don't know why you can't modify it.

m3nu:
I'll try again.

m3nu:
Here one option using `QToolButton` and the submenu already there. (needs some beatification and menu needs to know what archive it belongs to)

Screen Shot 2020-08-11 at 13 56 32

samuel-w:
I don't think we should put a massive Edit button on every archive. For me it prevents the entire archive name from being seen. Maybe replace the text with an icon and find some way to resize the top bar so that only the icon visible.

m3nu:
I know. That's why I wrote "needs some beatification" ;-)

samuel-w:
Instead of creating a new button, we could just set the text on the bottom to "Right click on a row to show more actions", since that text will always be visible.

samuel-w:
Added a button to demonstrate. Takes up less space than adding a small button per row.

samuel-w:
Added multiple selections.

samuel-w:
So what UI are we settling on for right click menu?

1. Nothing
2. Tooltip
menu
3. Edit button in table
table
4. Menu button in bottom bar + tooltip
mmenu
5. Some other idea?

m3nu:
Tooltips/right-click aren't obvious enough. Edit button in table? But smaller with just an icon?

samuel-w:
Menu button in bar is a left click action, and takes up the least screen space.

m3nu:
That's correct, but it's also less clear than the per-row action button. Unless we find a better name. *Menu* is too broad. *Archive Action*?

samuel-w:
_Archive Actions_ would work

samuel-w:
This is what it looks like now:
now

m3nu:
Buttons look good, I'd see.

m3nu:
Can you rebase this PR on the current master branch and then force-push to your branch?

Also I'd make sure:

- [ ] Error message when no archive selected?
- [ ] Maybe move *Check* to the right side, since it applies to whole repo. Or is there any logic to having it on the left?

samuel-w:
Actually, I wonder if there a way to have all the buttons on the bar (no menu), and have the two sides of buttons automatically become two rows only if the screen is too small

I found this https://doc.qt.io/qtforpython/overviews/qtwidgets-layouts-flowlayout-example.html that we should use instead of a menu

samuel-w:
I gave up on the flowlayout thing. My concern was that the small menu button would look weird on a large display, but I guess it looks file.

m3nu:
Let's save this for 0.7.3 and see how to reorganize the Archive buttons.

m3nu:
Continue in #864

Issue #593, opened 2020-08-22
Improve Backup source management


XXXBold:
**Is your feature request related to a problem? Please describe.**
Implementing this will improve usability when modifying the backup source files/directories.

**Describe the solution you'd like**
I've got some ideas, which will improve the handling of the backup source list. If you want these issues splitted, let me know.

- [x] **1. Allow selection of multiple files/folders on the file browser.**
For now, when multiple files/folders within the same folder are desired for backup, they have to be added each one-by-one, which is very inefficient and annoying for multiple items, because you have to browse to this folder again and again, for each item.

Implementation suggestion:
I think there's a flag for the file browser function to allow multiple item selection. Add an entry for each selected.

- [x] **2. Allow selection of multiple items withon the sources list**
This is easier if one wants to delete multiple entries at once.

Implementation suggestion:
Shouldn't be difficult to implement I think.

- [x] **3. Add an option to sort the entries within the sources list**
Since each new entry is added to the end of the list, it can become confusing. I'd suggest that there should be at least an option to sort the entries in alphabetical order if desired.
In general, I'd appreciate if there'd be some kind of separation between files and folders.

Implementation suggestion:
Add an additional button to sort the entries. Maybe buttons for moving an entry up/down within the list will also be useful?

- [x] **4. Show occupied space of entries + size of all files**
It would be nice if the size on disk of each entry would be shown aswell. Additional, the size of all entries could be shown somewhere aswell.
This would be very convenient to the enduser, because for now one can only guess how big the backup will be (excluding compression), without running it.

Implementation suggestion:
Replace listcontrol with e.g. tablecontrol (Maybe there's a better control in QT, IDK).
Create two columns (This could also be useful for filtering I guess):
[Path][Size]
When doing this way, it's easy to add other columns aswell if required.

The sum of all entries' size could be shown below the list.

**Describe alternatives you've considered**
1.: There's no alternative to do this faster afaik.
2.: Delete one by one
3.: Not possible right now
4.: N/A

samuel-w:
For number 4, it wouldn't be that useful after the first backup because of de-duplication.

m3nu:
> Edit: For 3, it would be better to automatically sort the entries, to reduce the number of buttons.

Yes, can be sorted automatically. No button for this.

![wgetgui](https://user-images.githubusercontent.com/3916435/91652406-aa9d8f00-eac9-11ea-82e0-19e731bf3a06.png)

[...]

PR #596, opened 2020-08-23
Enable multiselection on source list

XXXBold:
This PR enables as proposed in #593 as number 2.

It's now possible to select & delete multiple entries at once.


m3nu:
Wow. Very simple implementation. 👍

m3nu (approved):

PR #597, opened 2020-08-27
Add remote ratelimiting

### LINKED ISSUE #590: Add Rate Limit, already supported by Borg
Offer an option to set a rate limit via [`--remote-ratelimit`](https://borgbackup.readthedocs.io/en/stable/faq.html#is-there-a-way-to-limit-bandwidth-with-project-name)

Suggested by Boris via email.

samuel-w:
This change requires another schema upgrade. However, we already have five or so 14 -> 15 upgrades. How should we handle this? 

Fixes #590, fixes #817

### DIFF
--- a/src/vorta/keyring/kwallet.py
+++ b/src/vorta/keyring/kwallet.py
@@ -52,7 +52,7 @@ class VortaKWallet5Keyring(VortaKeyring):
         try:
             self.handle = int(output)
         except ValueError:  # For when kwallet is disabled or dbus otherwise broken
-            return -2
+            self.handle = -2
 
 
 class KWalletNotAvailableException(Exception):


m3nu:
We just change the DB version before merging it. Just a big annoying while testing.

m3nu:
A nice and compact PR that addresses one thing only. Very refreshing to see. 👍

m3nu:
> This change requires another schema upgrade. However, we already have five or so 14 -> 15 upgrades. How should we handle this?

I'll deal with this when merging. No worries.

m3nu:
I don't think this feature is important enough to go on the first tab. Suggest to put it somewhere in the *Schedule* tab, where we already have other network settings.

m3nu:
Sorry for the delay in merging. I've now approved or assigned all your PRs and will merge them one-by-one or comment.

samuel-w:
Moving issues to discussions prevents automatically closing them in pull requests. :(

m3nu:
That's OK. If a discussion yields a good feature idea, we open an issue for it.

The discussion can be continued after the issue is closed and implemented.

Old discussions that aren't interesting for many users will sink to the bottom or not get upvoted. So it's more self-correcting than having it all in issues.

PR #600, opened 2020-08-29
Add selection of multiple files

samuel-w:
Implements number 1 of #593 
I don't think there's a way to add multiple directories, https://doc.qt.io/qt-5/qfiledialog.html#FileMode-enum
Changed wording of the UI:
wording


m3nu (approved):
Good job.

PR #605, opened 2020-08-31
Add desktop file installation button (Linux only)

samuel-w:
Replacement of #564, closed due to complexity and possibly breaking Debian packaging guidelines

I moved the icons because otherwise it would be out of the scope of the package.


Hofer-Julian:
I don't really see what icon copying has to do with autostart. Could you please elaborate?

samuel-w:
The desktop file has a line `Icon=com.borgbase.Vorta`. Normally a user would have to manually copy the icon. This automates that process, along with adding the icon for notifications.

Hofer-Julian:
But this code would:
- be only called when you change autostart settings
- be called every time you change autostart settings.

samuel-w:
Changed it so now:
Autostart functionality is same as before
New button copies desktop file and icons.

Hofer-Julian:
I appreciate your efforts. But I think we should either find a different way than doing this manually during runtime or just accepting that people who use pip directly have to do some extra work. I currently tend towards the latter 😊

samuel-w:
Yeah, I'm going to go with copying things manually, since if a user is using pip its safe to assume they have command line knowledge.

PR #607, opened 2020-09-03
Check if keyring is unlocked before setting password

### LINKED ISSUE #606: When adding a repo, setting password fails without feedback when password manager is locked
**Describe the bug**
When adding a repo, setting password fails without feedback when password manager is locked

**To Reproduce**
1. Open the initialize/add repository with kwallet/secretstorage unlocked
2. Lock kwallet/secretstorage.
3. Enter a password and repo that uses encryption
4. Click add

Set password invisibly fails for KWallet, or does nothing while throwing exception in log for secretstorage.
Not sure about macOS keychain.

Suggested fix: Make set_password return whether or not the password was set correctly with get, and either:

- Fallback to VortaDB (not advised due to security concerns)
- Tell user to unlock password manager ("Please unlock your password manager")

Should check if exception is thrown, otherwise generic fix method is to check with get_password after set_password
`return bool(self.get_password(sevice, repo_url))`

 - OS: Ubuntu 20.04/Kubuntu 20.04

**Additional context**

Log for secretstorage:
```
2020-08-31 20:53:34,066 - vorta.borg.borg_thread - DEBUG - Using VortaSecretStorageKeyring keyring to store passwords.
2020-08-31 20:53:34,066 - asyncio - DEBUG - Using selector: EpollSelector
  File "/home/user/.local/lib/python3.8/site-packages/vorta/views/repo_add_dialog.py", line 64, in run
    params = BorgInitThread.prepare(self.values)
  File "/home/user/.local/lib/python3.8/site-packages/vorta/borg/init.py", line 20, in prepare
    ret = super().prepare(profile)
  File "/home/user/.local/lib/python3.8/site-packages/vorta/borg/borg_thread.py", line 121, in prepare
    ret['password'] = keyring.get_password('vorta-repo', profile.repo.url)
  File "/home/user/.local/lib/python3.8/site-packages/vorta/keyring/secretstorage.py", line 30, in get_password
    collection = secretstorage.get_default_collection(self.connection)
  File "/usr/lib/python3/dist-packages/secretstorage/collection.py", line 155, in get_default_collection
    return Collection(bus)
  File "/usr/lib/python3/dist-packages/secretstorage/collection.py", line 42, in __init__
    self.collection_props_iface.Get(COLLECTION_IFACE, 'Label',
  File "/usr/lib/python3/dist-packages/secretstorage/util.py", line 31, in function_out
    return function_in(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.UnknownObject: No such object path '/org/freedesktop/secrets/aliases/default'
```

samuel-w:
I need someone to test the macOS keyring. Should display `Please unlock your password manager.` if locked, nothing if unlocked when adding a repository. I have no idea whether it will work or not since I don't have a mac to test it on.

Could still break since lock check happens a few seconds before adding a repository, but I think this covers most cases. The only way to fix it would be to set the password and then delete it if the action failed which is a bit clunky and could result in data loss. Just don't lock the keyring in the 10 seconds it takes to add a repo. 

Fixes #606


m3nu:
macOS already shows a prompt to unlock the keychain. So I only got the new message when adding a _new_ repo, not an existing one. Still OK to add this check, as it's common with keychains.

Screen Shot 2020-09-04 at 19 33 33

samuel-w:
It's only should pop up if you dismiss the keyring unlock prompt.

samuel-w:
The secretstorage depends on either `dbus-python` or `jeepney`. How do I catch both `dbus.exceptions.DBusException` and `jeepney.wrappers.DBusErrorResponse` without having both libraries?

samuel-w:
Is this good to go?

m3nu:
This overlaps with #550. So let's merge that first and then rebase to avoid any conflicts.

m3nu:
After that I'll test on macOS. Though most changes are for Linux. Locked keychains are possible, but very unusual for the main login keychain in macOS.

m3nu (approved):

PR #611, opened 2020-09-03
Display progress information

### LINKED ISSUE #594: Display Borg Progress data, different progress indicator
It's very annoying when one runs a bigger backup, which will take some time, and there's absolutely no visible progress and ETA.

- Change the progress bar on the "Repository" tab to actually show the progress in percent. For an easy solution, simply count the files, ignoring their size, that's better than it is now.
- Add status text, containing the following: Current speed, bytes copied, bytes left, and depending on the speed, give an ETA (Yes, this might be inaccurate, I'm aware of that)

Disclaimer:
I'm not familiar with the borg API, so probably there's no (easy) way to get progress/files count/data sizes etc. If so, maybe some changes could be made to borg first?

Can't find one

SergioRAgostinho:
An implementation of what was described in 
https://github.com/borgbase/vorta/issues/594#issuecomment-686363495

See it live
output

Closes #594 

On an unrelated note: the repo stats reporting a deduplicated size larger than the compressed size contradicts a bit the information from the progress report. Is this expected?

### DIFF
--- a/src/vorta/borg/borg_thread.py
+++ b/src/vorta/borg/borg_thread.py
@@ -13,7 +13,7 @@ from subprocess import Popen, PIPE
 
 from vorta.i18n import trans_late
 from vorta.models import EventLogModel, BackupProfileMixin
-from vorta.utils import keyring, borg_compat
+from vorta.utils import keyring, borg_compat, pretty_bytes
 from vorta.keyring.db import VortaDBKeyring
 
 mutex = QtCore.QMutex()
@@ -200,6 +200,14 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
                             logger.log(level_int, parsed["message"])
                         elif parsed['type'] == 'file_status':
                             self.log_event(f'{parsed["path"]} ({parsed["status"]})')
+                        elif parsed['type'] == 'archive_progress':
+                            msg = (
+                                f'Original: {pretty_bytes(parsed["original_size"])}, '
+                                f'Compressed: {pretty_bytes(parsed["compressed_size"])}, '
+                                f'Deduplicated: {pretty_bytes(parsed["deduplicated_size"])}, '
+                                f'Files: {parsed["nfiles"]}'
+                            )
+                            self.progress_event(0, msg)
                     except json.decoder.JSONDecodeError:
                         msg = line.strip()
                         if msg:  # Log only if there is something to log.
@@ -242,6 +250,9 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
 
     def log_event(self, msg):
         self.updated.emit(msg)
+
+    def progress_event(self, value, fmt):
+        pass
 
     def started_event(self):
         self.updated.emit(self.tr('Task started'))
--- a/src/vorta/borg/borg_thread.py
+++ b/src/vorta/borg/borg_thread.py
@@ -207,7 +207,7 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
                                 f'Deduplicated: {pretty_bytes(parsed["deduplicated_size"])}, '
                                 f'Files: {parsed["nfiles"]}'
                             )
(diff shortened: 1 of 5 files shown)


src/vorta/borg/borg_thread.py
```diff
@@ -200,6 +200,14 @@ def read_async(fd):
                             logger.log(level_int, parsed["message"])
                         elif parsed['type'] == 'file_status':
                             self.log_event(f'{parsed["path"]} ({parsed["status"]})')
+                        elif parsed['type'] == 'archive_progress':
+                            msg = (
+                                f'Original: {pretty_bytes(parsed["original_size"])}, '
+                                f'Compressed: {pretty_bytes(parsed["compressed_size"])}, '
+                                f'Deduplicated: {pretty_bytes(parsed["deduplicated_size"])}, '
```
SergioRAgostinho, line 207: Potentially change order here to match the repo-tab order.

XXXBold:
One more thing: It would be nice if the total file count could also be shown, if available, means extend the string at the end to "...Files: 10/1234"

SergioRAgostinho:
That information is not available. Here's a sample of the json payload.
```
{'original_size': 229620919, 'compressed_size': 37674787, 'deduplicated_size': 37674787, 'nfiles': 205, 'time': 1599160038.8519056,  'type': 'archive_progress', 'path': 'tmp/data/ModelNet40/curtain/train/curtain_0130.off'} 
```

XXXBold:
Okay, too bad, but I might have another method: Will it help if you can get the current files count from the sources control? Or do you need another number?

I'm asking cause I'm working on getting the directory/file sizes within the source list, and if it helps you I could also simply count all files/folders.

SergioRAgostinho:
> I'm asking cause I'm working on getting the directory/file sizes within the source list, and if it helps you I could also simply count all files/folders.

Is there something already implemented?

If yes, can you point me to the method? I can cache this at the beginning of the backup and use it not only to report the number but to update the progress bar value as well.

If not, can we move this to a different PR?

XXXBold:
It's not yet available, but I'm working on it, see #593 point no. 4, will take a while, though.
I guess using for the progress bar could also be added later, so you don't have to wait with this PR.

m3nu:
Wow. Love it. 😍

m3nu:
MacOS can't have text inside the progress bar. Maybe we use a different element instead of it.

> Note that whether or not the text is drawn is dependent on the style. Currently CleanLooks and Plastique draw the text. Mac, Windows and WindowsXP style do not. https://doc.qt.io/archives/qt-5.8/qprogressbar.html#text-prop

m3nu:
Screen Shot 2020-09-04 at 14 52 09

Using a label field, so it works everywhere.

m3nu:
Basic GIF animation to replace progress bar.

Screen Shot 2020-09-04 at 15 26 20

m3nu:
We should probably call this primary- and secondary status display and use it for different things. E.g. the primary one could display our own status messages, while secondary shows Borg's logs.

SergioRAgostinho:
> Basic GIF animation to replace progress bar.

It's looking a bit static :sweat_smile:

m3nu:
Because I'm too lazy to make a movie, like you. But it does do something.

SergioRAgostinho:
How it currently looks on my setup
output


> Because I'm too lazy to make a movie, like you.

Don't let that stop you. Here's the one-liner in case you already have ffmpeg installed
```
ffmpeg -i input-video.mp4 -vf "fps=15,scale=640:-1:flags=lanczos,split[s0][s1];[s0]palettegen[p];[s1][p]paletteuse" -loop 0 output.gif
```

SergioRAgostinho:
:+1: I will not touch this one then. I think it's pretty much finished if we exclude the potential ui revision you might want to perform now, with the two existing text labels.

m3nu:
Closing in favor of #615

PR #615, opened 2020-09-04
Improved progress updates. By @SergioRAgostinho

m3nu:
Avoid merge conflicts from #611


m3nu:
Change summary:

- This removes the (unused) progress bar and adds a second text field instead. We now have 2 fields for displaying status messages: **progress** (for Vorta's high-level messages, like "Starting Backup") and **log** (to pass through Borg's more detailed messages).
- I also removed some unneeded `log_event()` functions that weren't really used.
- Since we will never be able to display a %-status, I added a stateless progress indicator just to the button.

All credit goes to @SergioRAgostinho for the idea and initial implementation.

src/vorta/assets/UI/mainwindow.ui
```diff
@@ -159,41 +159,17 @@
...
        <widget class="QPushButton" name="cancelButton">
         <property name="enabled">
          <bool>false</bool>
         </property>
+        <property name="maximumSize">
+         <size>
+          <width>150</width>
+          <height>16777215</height>
+         </size>
+        </property>
```
SergioRAgostinho, line 172: You can avoid having to explicitly set widths for all of the surrounding elements if you set the horizontal policy of the new text label, exactly like the progress bar previously had
https://github.com/SergioRAgostinho/vorta/blob/5752233ab010299189ffae0845da6f5dd470df16/src/vorta/assets/UI/mainwindow.ui#L228-L233

m3nu: Good tip. Changed.

src/vorta/views/main_window.py
```diff
@@ -156,19 +167,17 @@ def profile_add_action(self):
             self.profileSelector.setCurrentIndex(self.profileSelector.currentIndex())
 
     def backup_started_event(self):
-        self.set_status(progress_max=0)
         self._toggle_buttons(create_enabled=False)
+        self.set_log('')
```
samuel-w, line 172:
```suggestion
        font = self.progressText.font()
        font.setBold(True)
        self.progressText.setFont(font)

```

src/vorta/views/main_window.py
```diff
@@ -156,19 +167,17 @@ def profile_add_action(self):
...
 
     def backup_finished_event(self):
-        self.set_status(progress_max=100)
         self._toggle_buttons(create_enabled=True)
         self.archiveTab.populate_from_profile()
         self.repoTab.init_repo_stats()
```
samuel-w, line 177:
```suggestion
        font = self.progressText.font()
        font.setBold(False)
        self.progressText.setFont(font)

```

src/vorta/views/main_window.py
```diff
@@ -156,19 +167,17 @@ def profile_add_action(self):
...
 
     def backup_cancelled_event(self):
         self._toggle_buttons(create_enabled=True)
-        self.set_status(progress_max=100)
-        self.set_status(self.tr('Task cancelled'))
+        self.set_log(self.tr('Task cancelled'))
```
samuel-w, line 181:
```suggestion
        font = self.progressText.font()
        font.setBold(False)
        self.progressText.setFont(font)

```

samuel-w (changes requested):
Functionality seems fine, but all text in the bottom bar is bold
bold
Conditionally set it with something like:
```
        font = self.progressText.font()
        font.setBold(STATE)
        self.progressText.setFont(font)
```

src/vorta/assets/UI/mainwindow.ui
```diff
@@ -211,65 +181,49 @@
...
+          <verstretch>0</verstretch>
+         </sizepolicy>
         </property>
-        <property name="textVisible">
-         <bool>true</bool>
+        <property name="font">
+         <font>
+          <pointsize>11</pointsize>
+          <weight>75</weight>
+          <bold>true</bold>
```
samuel-w, line 195:
```suggestion
```

samuel-w (commented):
And remove this

samuel-w:
Could insert it in set_progress, with the condition being:
`text.startswith("Files")`
but it might act weirdly with translations.

src/vorta/assets/UI/mainwindow.ui
m3nu: I actually wanted to make it bold. You think it's a bad idea?

src/vorta/views/main_window.py
m3nu: The font sizes seem funny on Linux. "Start Backup" button seems too small.

src/vorta/assets/UI/mainwindow.ui
m3nu: You mean the issue is that sometimes the font isn't bold? That wasn't intended.

src/vorta/views/main_window.py
m3nu: I'm also not happy with the button changing size. Better we use a min-size again.

m3nu: Made it non-bold in standard font size. Logs below slightly smaller. Hope this doesn't cause issues on Linux. Min size for the button.

Screen Shot 2020-09-05 at 19 28 11

src/vorta/views/main_window.py
m3nu: Fully removed the bold font. Since we update the progress very often, we should avoid extra instructions.

src/vorta/assets/UI/mainwindow.ui
m3nu: Removed all bold formatting.

m3nu:
I ended up removing absolute font sizes and formatting, since they look different on Linux and macOS. This is with undefined size:

Screen Shot 2020-09-05 at 19 45 04

Screen Shot 2020-09-05 at 19 45 31

m3nu:
I'm aware that progress updates aren't translated yet. Since we also translate logs coming from Borg directly, that's fine for now. Could be added later, if performance allows it.

SergioRAgostinho:
> Peek 2020-09-05 01-55

@m3nu it might be worth to set a fixed (minimum) width for the buttons. So the size doesn't change once the spinner appears/disappears.

Hofer-Julian (approved):
> @m3nu it might be worth to set a fixed (minimum) width for the buttons. So the size doesn't change once the spinner appears/disappears.

That sounds like a good idea.
Otherwise ready to merge from my side :)

m3nu:
> @m3nu it might be worth to set a fixed (minimum) width for the buttons. So the size doesn't change once the spinner appears/disappears.

Already did that. It's in the last commit. (needs forced-pull, since I rebased)

m3nu:
Merged. Thanks everyone for helping out. @SergioRAgostinho for adding most of the implementation and @samuel-w and @Hofer-Julian for the review and feedback.

PR #617, opened 2020-09-05
Improve source management Control

XXXBold:
This PR aims to change the source tab to show more information, as suggested in #593 no. 4.
I think some discussion is required in regards of design/implementation details

State:
PR is feature complete, some issue on Mac OS remaining...

First Preview of new source UI:
vorta_srcmgmt_new_draft


XXXBold:
Further ideas:
- Why are there two different buttons to add files or folders? Couldn't that be combined, since multiple selection is possible now?

samuel-w:
Not according to QFileDialog's enums. It's either files or a folder.
https://doc.qt.io/qt-5/qfiledialog.html#FileMode-enum

m3nu:
He's right. I already tried that. Can't mix files and folders unless there is a custom file selector class (maybe worth it).

samuel-w:
Found this for files and folders:
https://www.qtcentre.org/threads/43841-QFileDialog-to-select-files-AND-folders

Thiis for multiple directories:
https://www.qtcentre.org/threads/34226-QFileDialog-select-multiple-directories?p=220108#post220108

XXXBold:
Okay, didn't know that this isn't possible, so I'll leave this point then.

@m3nu @samuel-w 
What do you think about the UI changes? Are you fine with them, or do you have other/more ideas?

m3nu:
I think it's great if it works without causing lots of background disk activity. And the calculation needs to happen async in the background. You'll need a placeholder before it's done. If people add network drives it could be even slower.

Also suggest to add an option under *Misc* to disable it.

samuel-w:
Not UI related, but:
The file/folder checking takes up all of my disk IO, and takes quite a long time. Try using `du -sh` and `du --inodes -s` for faster folder size/file count.

XXXBold:
It doesn't cause disk activity in the background (Well, it does once, when it calculates the sizie initial).
The placholder already exists (just the text "load..." for now), and it's processed in the background (1 thread/path), once when a source is added or when vorta is started.

An additional button to manually update the sizes could be placed somewhere...

m3nu:
Yeah. That will be the issue. If this runs whenever the profile changes, it will cause heavy disk activity.

XXXBold:
> Not UI related, but:
> The file/folder checking takes up all of my disk IO, and takes quite a long time. Try using `du -sh` and `du --inodes -s` for faster folder size/file count.
> Both are standard Unix, so shouldn't be an issue.

Hm, are you sure you want to go the unix way? I sticked to QT for portability (Future windows version, for example), but I'm up for discussion. Also, currently there's an artificial delay for 5 seconds (for testing) before the each thread returns.

m3nu:
I would only calculate when adding or when it's manually refreshed. Whenever it's opened will be too often. E.g. I have multiple profiles to do different home folder backups.

Also: how about Borg exclusions below? Once I exclude large folders, the calculated total will be meaningless. :-(

samuel-w:
Maybe calculate the exclusion size separately and subtract? Would lead to double the IO worst case.

XXXBold:
I didn't think about exclusions for now tbh, since the idea was to only show the current size of each entry, but true, that's an issue aswell.

If you don't want the size to be calculated on startup, it needs to be saved in the database, along with the path I guess?

samuel-w:
Yeah, should be part of SourceFileModel

m3nu:
> Yeah, should be part of SourceFileModel

Yes. Good idea. We don't have much in that table yet. At least it will be used for something. So I'd only calculate the size when adding it or manually refreshing. Together with a checkbox in *Misc* to disable calculating it when adding. E.g. "Calculate source folder statistics when adding." Enabled by default.

XXXBold:
Okay, I'll do some reworks. BTW what's the best practice in python for interfacing with native APIs (the mentioned `du` command for example)? Running as a cmd and parse the output string? Or is there a better method?

prefixtry:
@XXXBold Marginal observation, which I would like to share when I look at the screenshot above:
Add Folder -> singular
Add File(s) -> singular + plural
Paste Folders/Files -> singular + plural but both forms written out.
->  Would it be possible to make the wording more consistent here?

m3nu:
> Okay, I'll do some reworks. BTW what's the best practice in python for interfacing with native APIs (the mentioned `du` command for example)? Running as a cmd and parse the output string? Or is there a better method?

If a lot of work is done by the CLI tool and the CLI tool has no mode for json (or similar) structured output, then calling it from `subprocess` can be surprisingly efficient.

E.g. I used to size folders in Python, but also use `du` now and didn't have issues. Just note that the args may be different between macOS and Linux. For Linux I use:

```
def get_folder_size(path):
    res = subprocess.run('du -sb .', cwd=path, shell=True, check=False,
                         stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
    return int(res.stdout.split(b'\t')[0])
```

macOS is most likely just `du -s`. (need to be careful with block sizes, apparent size)

XXXBold:
I've added new attributes for the SourceFileModel. However, I think this requires a new DB-Version? Would appreciate if someone could help me with this, no idea what to do there :P

Also, if this changes anyway, should the "dir" attribute be renamed to "dirPath"? Would just be more obvious then.

src/vorta/models.py
```diff
@@ -106,6 +106,9 @@ class Meta:
 class SourceFileModel(pw.Model):
     """A folder to be backed up, related to a Backup Configuration."""
     dir = pw.CharField()
```
XXXBold, line 108: Change this name to dirPath?

samuel-w: Don't, it will require a very awkward db upgrade.

samuel-w:
You need to add schema upgrading code in init_db,  for this:
```
    if current_schema.version < 16:
        _apply_schema_update(
            current_schema, 16,
            migrator.add_column(SourceFileModel._meta.table_name,
                                'dirSize', pw.IntegerField()),
            migrator.add_column(SourceFileModel._meta.table_name,
                                'dirFilesCount', pw.IntegerField()),
            migrator.add_column(SourceFileModel._meta.table_name,
                                'dirType', pw.BooleanField())
        )
```
Also multi word variables are in formatted as snake_case not camelCase in Python.

XXXBold:
Thanks! I've also considered your hint for the naming convention, and changed the variable names, will also look over the other files...

src/vorta/models.py
```diff
@@ -106,6 +106,9 @@ class Meta:
...
+    dir_size = pw.IntegerField()
+    dir_files_count = pw.IntegerField()
+    dir_type_isdir = pw.BooleanField()
```
XXXBold, line 111: This variable name isn't good, might change it...

samuel-w: Change to `path_isdir`?

XXXBold: changed, first wanted to stick to the "dir_" prefix, but nvm

XXXBold:
Update:

Made some progress, datasizes are now stored within the DB, I've also added the option in the misc tab to disable automatic calculation when added. Calculation of a folder's size is updated to run async and use the proposed unix command by @m3nu 
I've also made it this way to get the total files count with another command (`find . | wc -l`), see the utils.py-file.

the UI now looks like that:
vorta_srcmgmt_new_draft2
For the path "/home/test/Musik": This is how it looks like when a path is added without enabled automatic calculation.
For the path "/home/test/Schreibtisch": This is how it looks while the (async) calculation is in progress. As soon as it's finished, the values on the UI are updated (there's still an artificial delay for 5s while testing)

Still missing: New button to manually update the sizes.

Also, do you want to take the exclude patterns into account for the sizes as mentioned before? I wouldn't, personally. If, add another column to e.g. show the "estimated size for backup" or something, but for the current "Size" row it should simply show the entry's size, imo.

Any thoughts on this, and/or is there more I'm missing?

src/vorta/views/source_tab.py
```diff
@@ -20,17 +39,52 @@ def __init__(self, parent=None):
...
+        if source.path_isdir == True:
+            self.sourceFilesWidget.setItem(indexRow,1,QTableWidgetItem("<DIR>"))
+        else:
+            self.sourceFilesWidget.setItem(indexRow,1,QTableWidgetItem("<File>"))
+        if update_data == True:
+            self.sourceFilesWidget.setItem(indexRow,2,QTableWidgetItem("load..."))
+            self.sourceFilesWidget.setItem(indexRow,3,QTableWidgetItem("load..."))
+            getDir = FilePathInfoAsync(source.dir)
+            getDir.signal.connect(self.set_path_info)
+            self.updateThreads.append(getDir) # this is ugly, is there a better way to keep the thread object?
```
XXXBold, line 68: Is there a better way to prevent the thread object from destruction, than adding it to the array?

XXXBold:
Added more functionality, seems pretty much complete from my side.

samuel-w:
Few issues:
- The size is incorrect, `du -sb /home/user` produces 270764M when run from command line, shows 447.4M in Vorta.
- Upgrading from a previous version makes all my folders show as files
- Tests are broken
- You should run autopep8 on your code, or use an IDE that does it for you for linting.

Some readability changes:
- Elements count should be something like "File and Folder Count"
- "Update sources" should be something like "Refresh sizes". Also the buttons space themselves apart, which didn't happen before.
- <File> should become File, <DIR> should become Folder

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,135 @@
...
 
 
 class SourceTab(SourceBase, SourceUI, BackupProfileMixin):
+    updateThreads = []
+    
     def __init__(self, parent=None):
         super().__init__(parent)
         self.setupUi(parent)
+        
+        headerTxt=["Path","Type","Size","Elements Count"]
```
samuel-w, line 20:
```suggestion
        headerTxt=["Path","Type","Size","File and Folder Count"]
```

src/vorta/assets/UI/sourcetab.ui
```diff
@@ -71,13 +95,20 @@
...
         <widget class="QPushButton" name="sourceRemove">
          <property name="text">
           <string>Remove</string>
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QPushButton" name="sourcesUpdate">
+         <property name="text">
+          <string>Update Sources</string>
```
samuel-w, line 108:
```suggestion
          <string>Recalculate sizes</string>
```

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,135 @@
...
+        self.updateThreads.append(getDir) # this is ugly, is there a better way to keep the thread object?
+        getDir.start()
+
+    def add_source_to_table(self,source,update_data):
+        index_row = self.sourceFilesWidget.rowCount()
+        self.sourceFilesWidget.insertRow(index_row)
+        item_path = QTableWidgetItem(source.dir)
+        self.sourceFilesWidget.setItem(index_row,0,item_path)
+        if source.path_isdir == True:
+            self.sourceFilesWidget.setItem(index_row,1,QTableWidgetItem("<DIR>"))
```
samuel-w, line 74:
```suggestion
            self.sourceFilesWidget.setItem(index_row,1,QTableWidgetItem("Folder"))
```

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,135 @@
...
+        else:
+            self.sourceFilesWidget.setItem(index_row,1,QTableWidgetItem("<File>"))
```
samuel-w, line 76:
```suggestion
            self.sourceFilesWidget.setItem(index_row,1,QTableWidgetItem("File"))
```

src/vorta/utils.py
```diff
@@ -31,6 +33,48 @@
...
+    signal = pyqtSignal(str,int,int)
+    def __init__(self,path):
+        self.path = path
+        QThread.__init__(self)
+        self.exiting = False
+
+    def run(self):
+        self.files_count = 0
+        self.size, self.files_count = get_path_datasize(self.path)
+        self.signal.emit(self.path,self.size,self.files_count)
```
samuel-w, line 46:
```suggestion
        self.signal.emit(self.path,str(self.size),str(self.files_count))        
```

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,135 @@
...
         self.sourceAddFile.clicked.connect(lambda: self.source_add(want_folder=False))
         self.sourceRemove.clicked.connect(self.source_remove)
+        self.sourcesUpdate.clicked.connect(self.sources_update)
         self.paste.clicked.connect(self.paste_text)
         self.excludePatternsField.textChanged.connect(self.save_exclude_patterns)
         self.excludeIfPresentField.textChanged.connect(self.save_exclude_if_present)
         self.populate_from_profile()
+        
+    def set_path_info(self,path,data_size,files_count):
+        items = self.sourceFilesWidget.findItems(path,QtCore.Qt.MatchExactly)
```
samuel-w, line 45: Insert here
```
data_size = int(data_size)
files_count = int(files_count)
```

samuel-w (changes requested):
Changes from previous comment

src/vorta/assets/UI/sourcetab.ui
```diff
@@ -71,13 +95,20 @@
          </property>
         </widget>
        </item>
-       <item alignment="Qt::AlignTop">
+       <item>
```
samuel-w, line 98: Change this back to \<item alignment="Qt::AlignTop">

samuel-w, line 98:
```suggestion
       <item alignment="Qt::AlignTop">
```

src/vorta/views/source_tab.py
XXXBold: Maybe like this, to save some space in horizontal direction?
```suggestion
        headerTxt=["Path","Type","Size","Folder /\nFile Count"]
```

src/vorta/utils.py
XXXBold: Don't really get this: is that related to "The size is incorrect, du -sb /home/user produces 270764M when run from command line, shows 447.4M in Vorta. This is caused by the size being greater than the C++ int limit, convert it to pretty_bytes before returning it."?

XXXBold:
First, thanks for the suggestions, I'll have a look.

I'm having some issues with the align on top for the buttons, after your change it looks like this: 
Bildschirmfoto vom 2020-09-27 10-52-31

samuel-w:
Add a spacer at the bottom
Screenshot_20200927_150228

src/vorta/utils.py
samuel-w: Yeah, converting it to string returning it, and then converting back to int prevents overflow.

This is because Python 3 int does not have a maximum, while C++ signed int does have a maximum (2^31-1). 
Qt 5 is C++ based, PyQt5 is just a binding that calls the C++ code.

XXXBold:
Dammit, I definitely messed up something... Let me fix this...

samuel-w:
Linted it and changed to a QTableWidget, as well as fixing tests.
[vorta.tar.gz](https://github.com/borgbase/vorta/files/5310445/vorta.tar.gz)
https://gist.githubusercontent.com/samuel-w/88271b17ef8b397f0d8c6b270ad80b59/raw/c2985a684598958864f92978298da32c9e6fb503/vorta.patch

The file/folder count should only display for folders, since files will always have a count of 1.

I recommend using an IDE that has git integration, or the git command line. Uploading files to the Github Web UI will be more likely to have errors.

src/vorta/views/source_tab.py
samuel-w: It would look strange on wide screens. Qt table automatically wrap words I believe. https://doc.qt.io/qt-5/qtableview.html#wordWrap-prop

XXXBold:
@samuel-w First, thanks a lot for your patches, really appreciate that!

I'm only using a regular texteditor for this right now, since I thought I'll only make some little changes, I'm also doing all the stuff inside a VM, to avoid breaking my real Vorta configuration. I'll stick to this method for now, will take better care in the future for things like this :)

How do you want to only show the Files count for folders? The cell for the file count exists for all entries, what should be shown instead of the "1" if it's a file?

samuel-w:
I would put nothing (empty string) there. Same for N/A, replace it with nothing. This is to avoid UI clutter.

samuel-w:
Also, instead of using a virtual machine, you can use a python virtual enviroment. 
Guide I used to set it up: https://docs.python-guide.org/dev/virtualenvs/

Alternative guide, assuming in root of vorta repository:
Install virtualenv (pip install --user virtualenv)
`virtualenv venv`
`source venv/bin/activate`
`pip install -e .`
To exit `deactivate`

XXXBold:
Phew, that sounds like it could complicate it even more, since I'm not very familiar with python and it's tools. Might look at this when I develop more in python, thanks for the hint :)

BTW, added the empty strings as suggested.

samuel-w:
Created a PR against your branch to fix the last few bugs https://github.com/XXXBold/vorta/pull/1

XXXBold:
Cool thanks! From my side, there's only one thing left that doesn't work right now: Sorting the entries by clicking on the header has no effect (Sort indicator is shown, but nothing is sorted).

samuel-w:
For sorting, set sortingEnabled to False at the start of populate_from_profile, and then set sortingEnabled to True at the end.
https://doc.qt.io/qt-5/qtableview.html#sortingEnabled-prop

samuel-w:
I think you have to override the sort method to work with the size labels https://doc.qt.io/qt-5/qabstractitemmodel.html#sort

XXXBold:
Thanks for the pointers, yes, needed to add an additional event for the sort indicator.

It's functional right now, however, sorting by size is not working, because it's sorted in alphabetical order, and the size is the formatted pretty_bytes string.

samuel-w:
I created some code to sort the pretty_bytes strings.
https://gist.github.com/samuel-w/02c8bbc368b4a014842c422b371cf907

XXXBold:
Phew, that looks complicated... Wouldn't it be easier to keep track of the original data sizes somehow?

samuel-w:
du has different syntax in macOS and Linux. You'll have to add coreutils as a dependency, or change the size command to `du -s -B 1`

samuel-w:
Also sorting by amount of files is broken, you'll have to override `__lt__` and convert the strings back to ints and compare them

XXXBold:
Implemented the sorting tuff, also cleaned up the code a bit.

Not sure if the changed command line now works for mac. If I use the command suggested by you `du -s -B 1 .`, the size reported is different to `du -s -b .`.

samuel-w:
According to this: https://askubuntu.com/questions/104848/how-can-i-see-a-folders-exact-size-in-bytes

`-b` is short for `-B1 --apparent-size`

macos uses the BSD du, https://www.freebsd.org/cgi/man.cgi?du

The equivalent of Linux `du -sb DIR` is `du -s -B 1 -A` I think

samuel-w:
Apparently BSD du does not output a prefix
Example output:
`567868 .`

samuel-w:
Apparently macOS does not use BSD utilities. They use one that doesn't have apparent size calculations https://ss64.com/osx/du.html.
My solution is to add the GNU utilities as a dependency, patch below.
```
From 4caf598a584b47af3c271afa8a2630d00f5ded0f Mon Sep 17 00:00:00 2001
From: samuel-w <samuel.woon@protonmail.com>
Date: Mon, 12 Oct 2020 20:21:26 -0500
Subject: Fix macOS tests, add coreutils dependency

---
 .github/workflows/main.yml | 1 +
 requirements.d/Brewfile    | 1 +
 src/vorta/utils.py         | 2 +-
 3 files changed, 3 insertions(+), 1 deletion(-)

diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index 1d6dce2..356df0d 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,6 +35,7 @@ jobs:
     - name: Install system dependencies (macOS)
       if: runner.os == 'macOS'
       run: |
+        brew install coreutils
         brew upgrade openssl readline xz  # pyenv pyenv-virtualenv
     - name: Install Vorta
       run: |
diff --git a/requirements.d/Brewfile b/requirements.d/Brewfile
index 6cb384e..5976b25 100644
--- a/requirements.d/Brewfile
+++ b/requirements.d/Brewfile
@@ -4,4 +4,5 @@
 brew 'qt'
 brew 'hub'
 brew 'xmlstarlet'
+brew 'coreutils'
 cask 'qt-creator'
diff --git a/src/vorta/utils.py b/src/vorta/utils.py
index c6affa1..d4a65f7 100644
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -59,7 +59,7 @@ def get_directory_size(dirPath):
 
     if files_count > 1:  # files count on empty directory is still 1 because of '.', ignore it
         if sys.platform == 'darwin':
-            res = subprocess.run('du -s -B 1 -A .', cwd=dirPath, shell=True, check=False,
+            res = subprocess.run('gdu -sb .', cwd=dirPath, shell=True, check=False,
                                  stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
             data_size = int(res.stdout.split()[0])
         else:
-- 
2.27.0
```

m3nu:
`Brewfile` is just for development dependencies. None of this gets installed for ordinary users.

samuel-w:
Is there a way to get apparent size with du on macOS then? That's the thing preventing using du to get the folder sizes.

m3nu:
Probably not. This is the man-page for it: https://www.unix.com/man-page/osx/1/du/

samuel-w:
I wrote this based off to replace the du and find commands, but its 3 times as slow.
```
def get_directory_size(source_dir_path):
    ''' Get number of files and total size in bytes from a path. Based off https://stackoverflow.com/a/17936789 '''
    data_size = 0
    seen = set()

    for dir_path, _, file_names in os.walk(source_dir_path):
        for file_name in file_names:
            file_path = os.path.join(dir_path, file_name)

            # Ignore symbolic links, since borg doesn't follow them
            if os.path.islink(file_path):
                continue

            stat = os.stat(file_path)

            # Visit each file once
            if stat.st_ino not in seen:
                seen.add(stat.st_ino)
                data_size += stat.st_size

    files_count = len(seen)

    return data_size, files_count
```

m3nu:
Better a slower Python implementation than the hassle of supporting different `du` binaries if you ask me.

XXXBold:
Added the python implementation provided by @samuel-w as a fallback for non-linux OS. However, I get different results, in both size and elements count.
Using the linux branch:
vorta_srcmgmt_data_du
Using the fallback:
vorta_srcmgmt_data_fallback

samuel-w:
The file counter counts files and folders for Linux, while the python code counts only files. 

The Linux one follows symbolic links, while the python code does not.

I recommend using only the python code, rather than switching implementations based on OS. You will have to change "File/Folder Count" to "File Count".

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,176 @@
...
+    def set_path_info(self, path, data_size, files_count):
+        items = self.sourceFilesWidget.findItems(path, QtCore.Qt.MatchExactly)
+        # Conversion int->str->int needed because QT limits int to 32-bit
+        data_size = int(data_size)
+        files_count = int(files_count)
+
+        for item in items:
+            db_item = SourceFileModel.get(dir=path)
+            if QFileInfo(path).isDir():
+                self.sourceFilesWidget.item(item.row(), SourceColumn.Type).setText("Folder")
```
samuel-w, line 75:
```suggestion
                self.sourceFilesWidget.item(item.row(), SourceColumn.Type).setText(self.tr("Folder"))
```

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,176 @@
...
+                self.sourceFilesWidget.item(item.row(), SourceColumn.FilesCount).setText(format(files_count))
+                db_item.path_isdir = True
+            else:
+                self.sourceFilesWidget.item(item.row(), SourceColumn.Type).setText("File")
```
samuel-w, line 79:
```suggestion
                self.sourceFilesWidget.item(item.row(), SourceColumn.Type).setText(self.tr("File"))
```

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,176 @@
...
+        # Remove thread from list when it's done
+        for thrd in self.updateThreads:
+            if thrd.objectName() == path:
+                self.updateThreads.remove(thrd)
+
+    def update_path_info(self, index_row):
+        path = self.sourceFilesWidget.item(index_row, SourceColumn.Path).text()
+        self.sourceFilesWidget.item(index_row, SourceColumn.Type).setText("Calculating...")
+        self.sourceFilesWidget.item(index_row, SourceColumn.Size).setText("Calculating...")
+        self.sourceFilesWidget.item(index_row, SourceColumn.FilesCount).setText("Calculating...")
```
samuel-w, line 97:
```suggestion
        self.sourceFilesWidget.item(index_row, SourceColumn.Type).setText(self.tr("Calculating..."))
        self.sourceFilesWidget.item(index_row, SourceColumn.Size).setText(self.tr("Calculating..."))
        self.sourceFilesWidget.item(index_row, SourceColumn.FilesCount).setText(self.tr("Calculating..."))
```

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,176 @@
...
+        self.sourceFilesWidget.setItem(index_row, SourceColumn.FilesCount, FilesCount(""))
+
+        if update_data:
+            self.update_path_info(index_row)
+        else:  # Use cached data from DB
+            if source.dir_size > -1:
+                self.sourceFilesWidget.item(index_row, SourceColumn.Size).setText(pretty_bytes(source.dir_size))
+
+                if source.path_isdir:
+                    self.sourceFilesWidget.item(index_row, SourceColumn.Type).setText("Folder")
```
samuel-w, line 120:
```suggestion
                    self.sourceFilesWidget.item(index_row, SourceColumn.Type).setText(self.tr("Folder"))
```

src/vorta/views/source_tab.py
```diff
@@ -1,64 +1,176 @@
...
+                    self.sourceFilesWidget.item(index_row,
+                                                SourceColumn.FilesCount).setText(format(source.dir_files_count))
+                else:
+                    self.sourceFilesWidget.item(index_row, SourceColumn.Type).setText("File")
```
samuel-w, line 124:
```suggestion
                    self.sourceFilesWidget.item(index_row, SourceColumn.Type).setText(self.tr("File"))
```

src/vorta/assets/UI/sourcetab.ui
```diff
@@ -37,12 +40,50 @@
...
+        </property>
+       </column>
+       <column>
+        <property name="text">
+         <string>Size</string>
+        </property>
+       </column>
+       <column>
+        <property name="text">
+         <string>File/Folder Count</string>
```
samuel-w, line 84:
```suggestion
         <string>File Count</string>
```

samuel-w (commented):
You should mark any strings that will be visible to the user for translation.

> Notes for Developers
> 
>     Original strings in .ui and .py must be American English (en_US) and ASCII.
>     In English, not translated:
>         log messages (log file as well as log output on console or elsewhere)
>         other console output, print().
>         docs
>         py source code, comments, docstrings
>     Translated:
>         GUI texts / messages
>     In Qt (sub)classes, use self.tr(“English string”), scope will be the instance class name.
>     Elsewhere use vorta.i18n.translate(“scopename”, “English string”)
>     To only mark for string extraction, but not immediately translate, use vorta.i18n.trans_late function. Later, to translate, use vorta.i18n.translate (giving same scope).

XXXBold:
All changes made/approved as requested.
Now, only the python-method is used to get the folder size/files count.
For me, the PR is now feature complete. Thanks once again @samuel-w and @m3nu for the great assistance on various topics!

m3nu:
Thanks for the contribution, @XXXBold . Great to see everyone tackling a problem to arrive at the best solution.

Let's proceed to review and then merge.

m3nu:
Just tried this locally. Works well and not intrusive, if someone doesn't want/need to calculate the sizes. Some thoughts I had:

- [ ] The table could use rows with different (striped) background colors, as we already do in the Archive tab.
- [ ] Would Qt support resizing the source table, while making the exclude area smaller?

And for future improvements:

- We could consider using multiple horizontal "Toolbox" screens on the _Source_ tab, like we do for _Archives_ and _Schedule_. To give enough space to the source table and re-arrange the buttons like we have on the *Archives* tab. Then it would also easier to expand the different Exclude options. E.g. manage exclusions in a better way than just lines in a textbox. See also https://github.com/borgbase/vorta/issues/399#issuecomment-583844419

src/vorta/models.py
```diff
@@ -215,6 +218,11 @@ def get_misc_settings():
             'label': trans_late('settings',
                                 'Open main window on startup')
         },
+        {
+            'key': 'get_srcpath_datasize', 'value': True, 'type': 'checkbox',
+            'label': trans_late('settings',
+                                'Get size of file/folder when added')
```
samuel-w, line 224:
```suggestion
                                'Get statistics of file/folder when added')
```
Since its size + count.

samuel-w, line 224:
```suggestion
                                'Get statistics of file/folder when added')
```
Since its size + count.

src/vorta/utils.py
```diff
@@ -26,10 +28,52 @@
 QApplication.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling, True)  # enable highdpi scaling
 QApplication.setAttribute(QtCore.Qt.AA_UseHighDpiPixmaps, True)  # use highdpi icons
 
+network_status_monitor = NetworkStatusMonitor.get_network_status_monitor()
```
samuel-w, line 31:
```suggestion
```

src/vorta/utils.py
```diff
@@ -132,7 +206,7 @@ def get_sorted_wifis(profile):
 
     from vorta.models import WifiSettingModel
 
-    system_wifis = get_network_status_monitor().get_known_wifis()
+    system_wifis = network_status_monitor.get_known_wifis()
```
samuel-w, line 209:
```suggestion
    system_wifis = get_network_status_monitor().get_known_wifis()
```

samuel-w: Keep the runtime network monitor determining.

samuel-w (changes requested):
Functionality is fine, just some minor style changes.

samuel-w:
For the alternating colours, IMO it looks cleaner with the same colour throughout.

src/vorta/utils.py
```diff
@@ -26,10 +28,51 @@
 QApplication.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling, True)  # enable highdpi scaling
 QApplication.setAttribute(QtCore.Qt.AA_UseHighDpiPixmaps, True)  # use highdpi icons
 
+logger.info('Using %s NetworkStatusMonitor implementation.', network_status_monitor.__class__.__name__)
+
```
samuel-w, line 32:
```suggestion
```

src/vorta/utils.py
```diff
@@ -98,10 +155,24 @@ def get_private_keys():
     return available_private_keys
 
 
+def sort_sizes(size_list):
+    """ Sorts sizes with extensions. Assumes that size is already in largest unit possible """
+    final_list = []
+    for suffix in [" B", " KB", " MB", " GB", " TB"]:
+        sub_list = [Decimal(size[:-len(suffix)])
```
samuel-w, line 162:
```suggestion
        sub_list = [float(size[:-len(suffix)])
```

src/vorta/utils.py
```diff
@@ -17,7 +17,9 @@
 from paramiko.ed25519key import Ed25519Key
 from paramiko.rsakey import RSAKey
 from PyQt5 import QtCore
+from PyQt5.QtCore import QFileInfo, QThread, pyqtSignal
 from PyQt5.QtWidgets import QApplication, QFileDialog, QSystemTrayIcon
+from decimal import Decimal
```
samuel-w, line 22:
```suggestion
```

src/vorta/utils.py
samuel-w, line 162:
```suggestion
        sub_list = [float(size[:-len(suffix)])
```

samuel-w (commented):
Now that I think about it, Decimal is not really needed.

m3nu:
What's the status here? Are we good to merge? Since this is a larger PR, would be good to get it merged before we get conflicts with other PRs.

samuel-w (approved):
Looks good to me.
Tested:
- Empty folder (Size 0B File Count 0)
- Folder with one file (Size 2B File Count 1)

Might want to add a note on the sources tab about the size reducing with compression.
Later we could use the exclude patterns and exclude if present to get a more accurate total size.

m3nu:
Works for me as well. Thanks for the great contribution, @XXXBold !

> Might want to add a note on the sources tab about the size reducing with compression.
> Later we could use the exclude patterns and exclude if present to get a more accurate total size.

Let's save those for later, when we re-arrange the whole Source tab to use multiple toolbox areas (source folders, exclusions). The PR is already large enough.

m3nu (approved):

PR #622, opened 2020-09-07
Recolour dark tray icon for dark tray

samuel-w:
As of https://github.com/borgbase/vorta/commit/79ef4717d7fc1a56cb29f43cf7189a812441b055 the system dark mode is automatically detected. This works in some cases, but for my case (Arc Darker on XFCE) the dark icon is impossible to see (its in the top right).
This restores the option for light tray icon. 
darj
Restored:
darj


m3nu:
Can't you detect your desktop env somehow and then use darkmode? Like we do for Gnome?

I'm reluctant to bring back all those options that we removed and that not many users need (macOS and Gnome are detected automatically)

samuel-w:
In XFCE, you can make your panel white, so autodetect would break in that case.
darj

Hofer-Julian:
And that can not be detected somehow? How do other apps do it?

samuel-w:
They surround the icon with a border like Discord:
darj

m3nu:
Then we should do the same for the tray icon? I already do the same for the BorgBase favicon.

samuel-w:
I recoloured them, this is what they look like now:
hdd-o-active-dark
hdd-o-dark
They look the same because I just added white in the middle

samuel-w:
This might eliminate the need for a dedicated light icon. Thoughts?

samuel-w:
For comparison, here are the light icons with black in the middle:
hdd-o-light
hdd-o-active-light

m3nu:
Not terrible.

Screen Shot 2020-09-07 at 13 11 27
Screen Shot 2020-09-07 at 13 11 06

m3nu:
hdd-o-active-light
hdd-o-light

Redid them from SVG

samuel-w:
Can you add them to the branch? Because they display as png in Github.

m3nu:
Those look alright on macOS and Gnome in dark- and light mode.

m3nu:
I added them as PNG. Don't think we can safely use SVG. I just exported them form SVG to get the best quality.

So we are sure we want to use only one icon for everything? Then I also need to adjust the icon loading function.

m3nu:
Screen Shot 2020-09-07 at 13 56 39
Screen Shot 2020-09-07 at 13 56 17
Screen Shot 2020-09-07 at 13 56 02
Screen Shot 2020-09-07 at 13 56 54

(Somehow VMWare doesn't have great resolution)

samuel-w:
darj
darj2
Looks good in the tray.

samuel-w:
I already adjusted the icon loading

m3nu:
I know, but I'd still suggest removing the `light` part to avoid any confusion, if we ever re-add a similar feature.

I can take care of it later when I merge it. Rest is good.

m3nu (approved):

PR #625, opened 2020-09-07
Avoid formatting date and time twice. Fixes #624

### LINKED ISSUE #624: Vorta 0.7.0 not starting in Mojave
After updating from Vorta 0.6.26 on macOS Mojave (10.14.6) the application is not starting.

The application will not start if running from the GUI, returning no errors.

From the command line, I get this:

/Applications/Vorta.app/Contents/MacOS/vorta-darwin
2020-09-07 14:07:06,961 - vorta.i18n - DEBUG - Loading translation succeeded for ['en', 'en-US', 'en-Latn-US'].
2020-09-07 14:07:09,217 - apscheduler.scheduler - INFO - Scheduler started
2020-09-07 14:07:09,220 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-09-07 14:07:09,220 - vorta.scheduler - DEBUG - New job for profile Default was added.
  File "__main__.py", line 44, in <module>
  File "__main__.py", line 37, in main
  File "application.py", line 56, in __init__
  File "application.py", line 90, in open_main_window_action
  File "views/main_window.py", line 50, in __init__
  File "views/schedule_tab.py", line 32, in __init__
  File "views/schedule_tab.py", line 68, in populate_from_profile
  File "views/schedule_tab.py", line 72, in init_wifi
  File "utils.py", line 130, in get_sorted_wifis
  File "network_status/darwin.py", line 48, in get_known_wifis
  File "peewee.py", line 4871, in format_date_time
TypeError: strptime() argument 1 must be str, not datetime.datetime
[74960] Failed to execute script __main__

- OS: macOS Mojave (10.14.6)
- Vorta version: 0.7.0
- Borg version: 1.1.11

m3nu:
Fix a bug where it tried to convert a datetime object to datetime again, when it should be a string.

### DIFF
--- a/src/vorta/network_status/darwin.py
+++ b/src/vorta/network_status/darwin.py
@@ -43,13 +43,13 @@ class DarwinNetworkStatus(NetworkStatusMonitor):
         result = []
         if wifis is not None:
             for wifi in wifis.values():
-                raw_last_connected = wifi.get('LastConnected', None)
-                last_connected = None if not raw_last_connected \
-                    else format_date_time(raw_last_connected, WifiSettingModel.last_connected.formats)
                 ssid = wifi.get('SSIDString', None)
-
                 if ssid is None:
                     continue
+
+                last_connected = wifi.get('LastConnected', None) or wifi.get('LastAutoJoinAt', None)
+                if isinstance(last_connected, str):  # TODO: Maybe not needed any more?
+                    last_connected = format_date_time(last_connected, WifiSettingModel.last_connected.formats)
 
                 result.append(SystemWifiInfo(ssid=ssid, last_connected=last_connected))
 
@@ -71,7 +71,7 @@ def call_ipconfig_getpacket(bsd_device):
     try:
         return subprocess.check_output(cmd)
     except subprocess.CalledProcessError:
-        logger.warn("Command %s failed", shlex.join(cmd))
+        logger.debug("Command %s failed", shlex.join(cmd))
         return b''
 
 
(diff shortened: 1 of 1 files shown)


src/vorta/network_status/darwin.py
```diff
@@ -43,14 +43,14 @@ def get_known_wifis(self):
...
-                raw_last_connected = wifi.get('LastConnected', None)
-                last_connected = None if not raw_last_connected \
-                    else format_date_time(raw_last_connected, WifiSettingModel.last_connected.formats)
                 ssid = wifi.get('SSIDString', None)
-
                 if ssid is None:
                     continue
 
+                last_connected = wifi.get('LastConnected', None) or wifi.get('LastAutoJoinAt', None)
+                if isinstance(last_connected, str):  # TODO: Maybe not needed any more?
```
samuel-w, line 51: Could this become `if last_connected is not None:`?

m3nu: No, because if it's `None` we can just save it to the DB.

There seem to be cases where `plistlib` returns a string, rather than a datetime object. Maybe for older versions. So we convert the string to datetime, as the original author did. For me it always returned a datetime object, but older macOS versions may use a different version of this file. So we keep it for now.

If `last_connected` is None, we don't want to convert the string to datetime, but just save the None as-is.

So we want either a None or a datetime, but never a string.

samuel-w (commented):
I don't have a mac to test these changes, so I'm going to assume it works. :)

m3nu:
Oh, I thought you had a Mac. Never mind then.

It's hard to test. I suspect the internal Apple file we use to read Wifi networks changed in some version.

PR #626, opened 2020-09-07
Use naked en locale rather than en_US. Fixes #623

### LINKED ISSUE #623: Vorta 0.7.0 is using German although device is set to English
**Describe the bug**
I have updated Vorta to the latest 0.7 release. My Mac is set to English as the primary language (see menu bar). German is setup as a secondary language. Vorta shows everything in German though, this did not happen with the release before 0.7.

**To Reproduce**

Screenshot 2020-09-07 at 13 12 24

- OS: macOS 10.15.6
- Vorta version: 0.7.0
- Borg version: 1.1.13

**Additional context**

m3nu:
When using English with a non-US region, our `en-US` default language file wasn't loaded.

This change renames the source language to just `en`.

See also https://doc.qt.io/qt-5/qtranslator.html#load-1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@ vorta.egg-info
 *.log
 htmlcov
 *.qm
-src/vorta/i18n/ts/vorta.en_US.ts
+src/vorta/i18n/ts/vorta.en.ts
 flatpak/app/
 flatpak/.flatpak-builder/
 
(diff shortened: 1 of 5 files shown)


m3nu:
Doesn't change actual files, just the exclusions. We shouldn't even ship `en`, but it's required for translations. So I assume it still made it into the macOS build.

Issue #619, opened 2020-09-06
Vorta 0.70 crashes when GUI is opened via tray icon while a backup is running


deermo:
**Describe the bug**
When I start a backup in Vorta 0.70 via system tray and  then open the GUI via system tray, Vorta crashes.
Might overlap with #386?
(OT: Thanks to the team for the many improvements in the 0.7 release!!)

**To Reproduce**
1. Open vorta (with window closed as last state)
2. Start backup via tray
3. Click try icon and click "Vorta" to open GUI 

**Desktop (please complete the following information):**
- OS: macOS 10.15.6
- Vorta version: 0.70
- Borg version: 1.1.13

**Additional context**
```
2020-09-06 19:01:19,107 - vorta.i18n - DEBUG - Loading translation succeeded for ['de-DE'].
2020-09-06 19:01:19,173 - apscheduler.scheduler - INFO - Scheduler started
2020-09-06 19:01:19,175 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-09-06 19:01:19,175 - vorta.scheduler - DEBUG - New job for profile PROFILE was added.
2020-09-06 19:01:19,175 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-09-06 19:01:19,176 - vorta.scheduler - DEBUG - New job for profile PROFILE was added.
2020-09-06 19:01:19,225 - vorta.borg.borg_thread - INFO - Running command /usr/local/bin/borg --version
2020-09-06 19:01:27,835 - vorta.borg.borg_thread - DEBUG - Using VortaDarwinKeyring keyring to store passwords.
2020-09-06 19:01:27,948 - root - WARNING - Command ipconfig getpacket en6 failed
2020-09-06 19:01:27,955 - root - WARNING - Command ipconfig getpacket en1 failed
2020-09-06 19:01:27,962 - root - WARNING - Command ipconfig getpacket en2 failed
2020-09-06 19:01:27,969 - root - WARNING - Command ipconfig getpacket en3 failed
2020-09-06 19:01:27,975 - root - WARNING - Command ipconfig getpacket en4 failed
2020-09-06 19:01:27,982 - root - WARNING - Command ipconfig getpacket bridge0 failed
[...]

Issue #613, opened 2020-09-03
Diff doesn't always work


harmlessdrudge:
**Describe the bug**

Diff function works sometimes, but often does nothing.

**To Reproduce**

Click on Archives, then DIff, select last two backups and then click diff

![vorta-diff](https://user-images.githubusercontent.com/3849521/92176000-89280480-ee35-11ea-8efc-df32cfefc46d.jpg)

Result is nothing shown here, despite hundreds of Mb of additional files.

**Desktop (please complete the following information):**

Linux Mint 19.3 w Cinnamon desktop

**Additional context**
If appropriate include logs. Can be found in Main Window > Misc Tab > Log.
[vorta.log](https://github.com/borgbase/vorta/files/5171512/vorta.log)

samuel-w:
I've had this happen too, even after just opening, but it goes away with a restart of Vorta.

ktosiek:
Do you see an entry for `borg-run diff` in Schedule -> Log? If so, what's the Returncode?

harmlessdrudge:
Yes, Returncode seems to be 0 regardless of whether or not any differences are displayed (every diff Returncode is 0).

ThomasWaldmann:
@harmlessdrudge can you reproduce an issue when just using borg directly? if so, please file a bug on the borg issue tracker.

harmlessdrudge:
Maybe this will help

`2020-09-04 17:27:23,639 - vorta.i18n - DEBUG - Loading translation failed for ['en-IE', 'en-Latn-IE'].
2020-09-04 17:27:23,663 - apscheduler.scheduler - INFO - Scheduler started
2020-09-04 17:27:23,665 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-09-04 17:27:23,665 - vorta.scheduler - DEBUG - New job for profile Default was added.
2020-09-04 17:27:23,715 - vorta.borg.borg_thread - INFO - Running command /usr/bin/borg --version
2020-09-04 17:28:06,454 - vorta.borg.borg_thread - DEBUG - Using VortaSecretStorageKeyring keyring to store passwords.
2020-09-04 17:28:06,455 - asyncio - DEBUG - Using selector: EpollSelector
2020-09-04 17:28:06,460 - root - DEBUG - Found 1 passwords matching repo URL.
[...]

Issue #521, opened 2020-06-19
Running diff on backups that have files with changed permissions causes a crash


ghost:
Running diff on backups that have files with changed permissions causes a crash

**To Reproduce**
1. Create a new repository
2. Create a backup that has at least one file in it
3. Change the ownership of a file in the backup, in my case user to root (I also tested root to user, still crashes)
4. Create another backup
5. Run a diff of the two backups in Vorta
Program crashes with this error:
```
 **Additional context**
 Traceback (most recent call last):
   File "/app/lib/python3.7/site-packages/vorta/views/archive_tab.py", line 472, in list_diff_result
     window = DiffResult(result['data'], archive_newer, archive_older)
   File "/app/lib/python3.7/site-packages/vorta/views/diff_result.py", line 73, in __init__
     files_with_attributes.append(parse_line(line))
   File "/app/lib/python3.7/site-packages/vorta/views/diff_result.py", line 61, in parse_line
     full_path = line[full_path_index:]
```

**Desktop (please complete the following information):**
 - OS:Ubuntu 20.04
 -  Python 3.8.2
 - Vorta 0.6.26, running in flatpak

Output of running diff manually:
> [user:user -> nfsnobody:nfsnobody] home/user/arrays/test.txt

PR #628, opened 2020-09-08
Improve exception handling

### LINKED ISSUE #621: Exceptions not logged in logs
**Describe the bug**
When Vorta crashes, the exceptions are not stored in the logs, making them harder to diagnose.
Not really that big of an issue, since the crash _should_ be reproducable and it is for reporting only.

**To Reproduce**
1. Find a crash that happens.
2. Produce the crash while running from the terminal.
3. Read the logs, and note that the crash is visible in the terminal, but not in the logs.

- OS: Ubuntu 20.04
- Vorta version: 0.7.0
- Borg version: 1.1.13

**Additional context**
Example crash #619 
Log file:
```
2020-09-06 13:11:37,544 - vorta.i18n - DEBUG - Loading translation failed for ['en', 'en-US', 'en-Latn-US'].
2020-09-06 13:11:37,556 - apscheduler.scheduler - INFO - Scheduler started
2020-09-06 13:11:37,558 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-09-06 13:11:37,558 - vorta.scheduler - DEBUG - New job for profile Backblaze was added.
2020-09-06 13:11:37,990 - vorta.borg.borg_thread - INFO - Running command /home/user/.local/bin/borg --version
2020-09-06 13:11:43,373 - vorta.borg.borg_thread - DEBUG - Using VortaSecretStorageKeyring keyring to store passwords.
2020-09-06 13:11:43,373 - asyncio - DEBUG - Using selector: EpollSelector
2020-09-06 13:11:43,378 - root - DEBUG - Found 2 passwords matching repo URL.
2020-09-06 13:11:43,411 - vorta.borg.borg_thread - INFO - Running command /home/user/.local/bin/borg create --list --progress --info --log-json --json --filter=AM -C zstd,8 --exclude-if-present .nobackup REPO_URL /home/user
2020-09-06 13:11:43,821 - vorta.borg.borg_thread - INFO - Creating archive at "REPO_URL"
```

Terminal output:
```
2020-09-06 13:11:37,544 - vorta.i18n - DEBUG - Loading translation failed for ['en', 'en-US', 'en-Latn-US'].
2020-09-06 13:11:37,556 - apscheduler.scheduler - INFO - Scheduler started
2020-09-06 13:11:37,558 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-09-06 13:11:37,558 - vorta.scheduler - DEBUG - New job for profile Backblaze was added.
2020-09-06 13:11:37,990 - vorta.borg.borg_thread - INFO - Running command /home/user/.local/bin/borg --version
2020-09-06 13:11:43,373 - vorta.borg.borg_thread - DEBUG - Using VortaSecretStorageKeyring keyring to store passwords.
2020-09-06 13:11:43,373 - asyncio - DEBUG - Using selector: EpollSelector
2020-09-06 13:11:43,378 - root - DEBUG - Found 2 passwords matching repo URL.
2020-09-06 13:11:43,411 - vorta.borg.borg_thread - INFO - Running command /home/user/.local/bin/borg create --list --progress --info --log-json --json --filter=AM -C zstd,8 --exclude-if-present .nobackup REPO_URL /home/user
2020-09-06 13:11:43,821 - vorta.borg.borg_thread - INFO - Creating archive at "REPO_URL"
  File "/home/user/.local/lib/python3.8/site-packages/vorta/application.py", line 90, in open_main_window_action
  File "/home/user/.local/lib/python3.8/site-packages/vorta/views/main_window.py", line 94, in __init__
    self.set_status(self.tr('Backup in progress.'))
AttributeError: 'MainWindow' object has no attribute 'set_status'

```

samuel-w:
Still exits on uncaught exception, which is same as before.
Should we translate the log message? Just the "Uncaught exception, file a report on GitHub"
Fixes #621 

Example, from #619:
```
2020-09-07 20:52:17,851 - root - ERROR - Uncaught exception, file a report on GitHub:
  File "/home/user/Projects/vorta/src/vorta/application.py", line 90, in open_main_window_action
  File "/home/user/Projects/vorta/src/vorta/views/main_window.py", line 94, in __init__
    self.set_status(self.tr('Backup in progress.'))
AttributeError: 'MainWindow' object has no attribute 'set_status'
```


m3nu:
Works as expected

```
2020-09-08 10:59:17,093 - root - ERROR - Uncaught exception, file a report on GitHub:
  File "/Users/manu/Workspace/vorta/src/vorta/borg/borg_thread.py", line 162, in run
    self.started_event()
  File "/Users/manu/Workspace/vorta/src/vorta/borg/create.py", line 41, in started_event
    raise Exception('Buha')
Exception: Buha
QSocketNotifier: Socket notifiers cannot be enabled or disabled from another thread
zsh: segmentation fault  vorta
```

m3nu:
How about putting the issue URL to make it even easier?

https://github.com/borgbase/vorta/issues/new

samuel-w:
The issue I'm having is some exceptions don't cause a crash, while others do cause a crash, and I can't differentiate them. 

Should we always crash on uncaught exceptions?

m3nu:
Which exceptions don't crash? Maybe they are caught somewhere? I wouldn't crash the app more often than we have to.

Another idea would be to open a window showing the exception and offering to direct it to Github? (prefill?)

src/vorta/__main__.py
```diff
@@ -5,12 +5,19 @@
...
 from vorta.updater import get_updater
 from vorta.utils import parse_args
 
 
+def exception_handler(type, value, tb):
+    # https://stackoverflow.com/questions/49065371/why-does-sys-excepthook-behave-differently-when-wrapped
+    logger.critical("Uncaught exception, file a report at https://github.com/borgbase/vorta/issues/new:",
+                    exc_info=(type, value, tb))
+    sys.__excepthook__(type, value, tb)
+
```
samuel-w, line 19: If I were to add `sys.exit(1)` here, then any exception, even if it didn't cause a crash, would cause a crash.

If I don't add it here, they any exception that would cause a crash, does _not_ cause a crash.

This behaviour only appears if Vorta is in the background.

samuel-w: I think I'll leave this as is, since if vorta in the background it's closable by the tray.

samuel-w: Turns out the different behaviour is because I was testing with different PyQt5 versions (5.14 vs 5.15)

samuel-w:
Now it double prints the exception to the terminal. Should be fine for reading bug reports.

samuel-w:
I've decided to always close on uncaught exception because it makes the behaviour more consistent across platforms, such as #613 and #521, where one crashes and the other does not, but both lead to undesired behaviour.

src/vorta/__main__.py
```diff
@@ -5,13 +5,24 @@
...
 from vorta.utils import parse_args
 
 
+def exception_handler(type, value, tb):
+    # https://stackoverflow.com/questions/49065371/why-does-sys-excepthook-behave-differently-when-wrapped
+    # This double prints the exception, want to only print the log entry
+    logger.critical("Uncaught exception, file a report at https://github.com/borgbase/vorta/issues/new:",
+                    exc_info=(type, value, tb))
+    sys.__excepthook__(type, value, tb)
+    sys.exit(1)
```
samuel-w, line 20: Remove this line to make uncaught exceptions never crash, may lead to undesired behaviour.

m3nu: Fine. Lets crash it all the time. At least we get bug reports quickly.

samuel-w: I've made the exception print a notification. Should I still create a crash?

samuel-w:
Sample notification:
example

m3nu:
How about a real UI window in Qt that has the exception and Github link?

Else we start with just adding it to the logs. Not sure a notification is useful since it can't be clicked and may be disabled.

samuel-w:
This is what is displayed now, exception is fake:
example
It only happens when the exception is in the foreground, if its in the background it just prints to log.

m3nu:
Good solution. Last suggestion: Move the "File a bug report on github.com/..." into the alert, so it can be a real link? Then only the actual exception is logged.

samuel-w:
It gets cutoff though.
example

m3nu:
> It gets cutoff though.
> example

Better add it inside the message, rather than as title. Title could be "Uncaught exception" or something?

samuel-w:
Adding a link destroys formatting.
example

m3nu:
Because HTML removes line breaks unless there is a `<br>`.

Still possible to make the alert a more complex UI element with two separate text widgets. Probably even possible programmatically without doing a .ui file.

samuel-w:
There is a way to automate filling out a bug report with https://github.com/borgbase/vorta/issues/new?title=&body=. We could use that, but it seems like too much work compared to copying and pasting.

m3nu:
I expected there to be some URL parameters, like for `mailto` links. I think for now it's already a great improvement if those exceptions end up a) in the logs and b) in a message window when a crash happens.

src/vorta/__main__.py
```diff
@@ -5,13 +5,33 @@
...
+        if app and app.main_window:
+            full_exception = ''.join(format_exception(type, value, tb))
+            try:
+                QMessageBox.critical(app.main_window,
+                                     app.main_window.tr("Fatal Error"),
+                                     app.main_window.tr(
+                                         "Uncaught exception, please file a report with this text at\n"
+                                         "https://github.com/borgbase/vorta/issues/new\n") + full_exception)
+            except RuntimeError:
+                # Window is closed, only log is available
```
m3nu, line 29: Not possible to show a messagebox without main window?

samuel-w: It would have the strings untranslated, since the translation comes from main window.

m3nu: Probably not an issue and new translations are added over time.

samuel-w: I mean that the translation is impossible, since it is inherited from the crashed application

m3nu: Then we may have a message window title that can be translated and the details on the exception below are always as-is untranslated, correct?

samuel-w: Found a way to do it, now it doesn't show translations when there is no `app`

m3nu (approved):

Issue #629, opened 2020-09-08
0.7.1 bugfix release and RCs


lvoogdt:
**Describe the bug**
On MacOS 10.15.5 Vorta 0.7.0 launches, but when i want to open the GUI it crashes immediately without any error. Removing the settings.db has no effect, so clean install has same bug. 0.6.26 seems to be working fine.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open Vorta
2. Open GUI via tray menu
3. Crash

**Desktop (please complete the following information):**
- OS: OSX 10.15.5 
- Vorta version: 0.7.0
- Borg version: 1.1.13

**Additional context**
If appropriate include logs. Can be found in Main Window > Misc Tab > Log.

```
2020-09-08 10:36:01,156 - vorta.i18n - DEBUG - Loading translation failed for ['en-NL'].
2020-09-08 10:36:01,184 - apscheduler.scheduler - INFO - Scheduler started
2020-09-08 10:36:01,185 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-09-08 10:36:01,185 - vorta.scheduler - DEBUG - New job for profile Default was added.
2020-09-08 10:36:01,389 - vorta.borg.borg_thread - INFO - Running command /usr/local/bin/borg --version
```

Btw is there a way to donate to this project?

siu00adg:
Same issue here except I'm on Manjaro Linux with XFCE. Was working great until the last update to 0.7.0-1.

```
$ vorta
2020-09-08 13:15:55,913 - vorta.i18n - DEBUG - Loading translation failed for ['en-GB', 'en-Latn-GB'].
2020-09-08 13:15:55,922 - apscheduler.scheduler - INFO - Scheduler started
2020-09-08 13:15:55,924 - apscheduler.scheduler - INFO - Added job "VortaScheduler.create_backup" to job store "default"
2020-09-08 13:15:55,924 - vorta.scheduler - DEBUG - New job for profile Default was added.
2020-09-08 13:15:55,963 - vorta.borg.borg_thread - INFO - Running command /usr/bin/borg --version

Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/vorta/application.py", line 90, in open_main_window_action
    self.main_window = MainWindow(self)
  File "/usr/lib/python3.8/site-packages/vorta/views/main_window.py", line 50, in __init__
    self.scheduleTab = ScheduleTab(self.scheduleTabSlot)
[...]

PR #634, opened 2020-09-08
Initialize the network status monitor lazily

ktosiek:
This avoids problems with `vorta -d` and fork. The problems were mentioned here: https://github.com/borgbase/vorta/issues/629#issuecomment-688998973

### DIFF
--- a/src/vorta/borg/create.py
+++ b/src/vorta/borg/create.py
@@ -4,7 +4,7 @@ from dateutil import parser
 import subprocess
 
 from vorta.i18n import trans_late
-from vorta.utils import format_archive_name, borg_compat, network_status_monitor
+from vorta.utils import format_archive_name, borg_compat, get_network_status_monitor
 from vorta.models import SourceFileModel, ArchiveModel, WifiSettingModel, RepoModel
 from .borg_thread import BorgThread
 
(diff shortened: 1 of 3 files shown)


samuel-w (approved):
Works great, thanks!

samuel-w:
It fixes the no daemon, but now metered networks are not showing up

Result of `dbus-send --print-reply --system --dest=org.freedesktop.NetworkManager "/org/freedesktop/NetworkManager/Settings/1" org.freedesktop.NetworkManager.Settings.Connection.GetSettings`: https://pastebin.com/xmE6q5Hz

ktosiek:
Is that the only "/org/freedesktop/NetworkManager/Settings/X" that `dbus-send --print-reply --system --dest=org.freedesktop.NetworkManager /org/freedesktop/NetworkManager/Settings org.freedesktop.NetworkManager.Settings.ListConnections` gives you? This one is an ethernet connection.
I don't have a WiFi card in my current PC, but I'll try testing on real hardware shortly (today/tomorrow).

The "Allowed Network" box only lists WiFi connections, by SSID. This was done to mirror the OSX behaviour, but maybe for NetworkManager it'd be better to show all networks? Expanding that list might not be simple - currently Vorta only looks at the [PrimaryConnection](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.html#gdbus-property-org-freedesktop-NetworkManager.PrimaryConnection), so some networks (like VPNs) don't make sense on that list.

samuel-w:
It should show all networks, since it is possible to have wired metered internet, such as satellite internet.

ktosiek:
All networks would include things that are not really internet connections though, I've asked on NM IRC and will see if there's a good heuristic we could use. For now you can mark a network as metered in NM settings - Vorta will honour this setting, independently of the networks list.

m3nu (approved):

PR #642, opened 2020-09-10
Keep version number in one place only. Clean up Makefile

m3nu:


setup.cfg
```diff
@@ -68,7 +69,8 @@ max-line-length = 120
 exclude =
     build,dist,.git,.idea,.cache,.tox,.eggs,
     ./src/vorta/views/dark/collection_rc.py,
-    ./src/vorta/views/light/collection_rc.py
+    ./src/vorta/views/light/collection_rc.py,
```
samuel-w, line 72: These files don't exist anymore, so these lines should be deleted

m3nu: Right. Good catch.

PR #661, opened 2020-09-20
Save window size correctly

samuel-w:
Fixes case where main window gets larger every time Vorta is opened.


sudwhiwdh:
What about saving and restoring the window state across sessions as well?

samuel-w:
I don't understand what you mean. The bug is caused by the frameGeometry() taking into account the title bar, which adds to the height.
```
print(f'{self.frameGeometry().width()},{self.frameGeometry().height()}')
print(f'{self.width()},{self.height()}')
```
Results
```
800,629
800,600
```

m3nu (approved):

PR #668, opened 2020-09-24
Give error message on unmount failure

samuel-w:
Also changed the grammar of success message, since no other text uses "un-mount"


src/vorta/views/archive_tab.py
```diff
@@ -299,14 +299,17 @@ def umount_action(self):
...
-            archive_name = result['params']['current_archive']
+            self._set_status(self.tr('Unmounted successfully.'))
             del self.mount_points[archive_name]
             self.update_mount_button_text()
             row = self.row_of_archive(archive_name)
             item = QTableWidgetItem('')
             self.archiveTable.setItem(row, 3, item)
+        else:
+            self._set_status(
+                self.tr(f'Unmounting failed. Make sure no programs are using {self.mount_points.get(archive_name)}'))
```
m3nu, line 312: I'm not sure f-strings are good to use here due to the translation. We don't want the Python code to end up in the translation string.

m3nu (changes requested):

m3nu (approved):

m3nu (approved):

PR #672, opened 2020-09-27
Update check message on completion

### LINKED ISSUE #671: [bug] Starting consistency check... doesn't clear the message
**Describe the bug**

Super minor display issue... it seems like the messages don't clear out correctly after a successful check... I would expect the 'Starting...' message to disappear once the 'No problems found' message came up.

image

- OS: OSX 10.15.5
- Vorta version: 0.7.1
- Installed from: brew

**Additional context**

image

samuel-w:
Other messages claim to complete without checking the returncode, for example BorgExtractThread. Should we change that?

Fixes #671


m3nu:
Good idea. Now that we have 2 areas for status updates, we could add more messages on success/failure based on return code without hiding the last error log message.

m3nu (approved):

Issue #340, opened 2019-08-09
Crash when re-opening main window during repo setup


fantom-x:
**Describe the bug**
Crash

**To Reproduce**
Start Vorta in MacOS
Click on the icon and bring up the Vorta interface
Select "Initialize New Backup Repository"
Leave it open, do not enter anything
Here I went to a different window (Files, Safari, etc) but it is not necessary: the next step will crash Vorta
Go back to the icon and bring the Vorta interface again: when it is hidden, using the icon is a convenient way of getting Vorta back in the view
Vorta crashes now

**Desktop (please complete the following information):**
MacOS latest
Vorta latest

m3nu:
I can reproduce this. The precise error is:

```
Traceback (most recent call last):
  File "vorta/views/repo_tab.py", line 159, in repo_select_action
RuntimeError: wrapped C/C++ object of type QComboBox has been deleted
[1]    8147 abort      ./dist/Vorta.app/Contents/MacOS/vorta-darwin
```

samuel-w:
Is this bug still present in 0.7.1?

m3nu:
Still present.

```
Traceback (most recent call last):
  File "/Users/manu/Workspace/vorta/src/vorta/views/repo_tab.py", line 171, in repo_select_action
    self.repoSelector.setCurrentIndex(0)
RuntimeError: wrapped C/C++ object of type QComboBox has been deleted
```

The issue is that we shouldn't redo the main window, but only bring it to the foreground. One line change, but it may be possible to simplify `open_main_window_action()` and `toggle_main_window_visibility()` further.

```
    def open_main_window_action(self):
        if not self._main_window_exists():
            self.main_window = MainWindow(self)
        self.main_window.show()
        self.main_window.raise_()
```

m3nu:
Added the fix suggested above.

PR #685, opened 2020-10-10
Fix crash when window is closed while thread running 

### LINKED ISSUE #684: Crash occurs whenever the main window is closed while an archive action is running
**Describe the bug**
A crash occurs whenever the main window is closed while a thread is running
**To Reproduce**
1. Start an action that takes a long time, such as checking a backup (Not creating a backup)
2. Close the main window

A crash occurs with `QThread: Destroyed while thread is still running`

- OS: Kubuntu 20.04, Xubuntu 20.04, Ubuntu 18.04
- Vorta version: 0.7.1
- Installed from: PPA, pip and flatpak

**Additional context**
Same error message as #603, but we never discovered why that happens

This is caused by https://github.com/borgbase/vorta/pull/208, since undoing it fixes this crash. The original purpose, to save memory, is replaced by #656. So it should be save to undo

```
2020-10-09 23:31:50,760 - vorta.i18n - DEBUG - Loading translation failed for ['en-CA', 'en-Latn-CA'].
2020-10-09 23:31:50,772 - apscheduler.scheduler - INFO - Scheduler started
2020-10-09 23:31:51,362 - root - INFO - Using NetworkManagerMonitor NetworkStatusMonitor implementation.
2020-10-09 23:31:51,442 - vorta.borg.borg_thread - INFO - Running command /usr/bin/borg --version
2020-10-09 23:31:53,987 - asyncio - DEBUG - Using selector: EpollSelector
2020-10-09 23:31:53,987 - vorta.borg.borg_thread - DEBUG - Using VortaSecretStorageKeyring keyring to store passwords.
2020-10-09 23:31:53,988 - asyncio - DEBUG - Using selector: EpollSelector
2020-10-09 23:31:53,990 - root - DEBUG - Found 1 passwords matching repo URL.
2020-10-09 23:31:54,020 - vorta.borg.borg_thread - INFO - Running command /usr/bin/borg check --info --log-json /repo
2020-10-09 23:31:54,282 - vorta.borg.borg_thread - INFO - Starting repository check
QThread: Destroyed while thread is still running
Aborted
```

samuel-w:
Fixes #684 

I only opened the issue to confirm that it wasn't a weird setup issue, since I remember this not crashing in the past.

Opening and closing the window is so much faster now, as an upside.

Check to make sure this doesn't reintroduce #340, doesn't in my testing.


samuel-w:
Takes up about 6-10MB more memory when window losed.

m3nu:
Sounds good. You want to do a PR for those changes, since you already have them working? I'll test on macOS.

samuel-w:
This is the pull request.
Here is what I have tested, all working:

- Closing/opening window with repository window open
- Running check and opening/closing window
- Creating backup from tray and opening/closing
- Creating backup from window and opening/closing

samuel-w:
Could also solve this by making the parent of the threads VortaApp.

samuel-w:
Went with the alternate method of fixing it, see comment above.

m3nu:
> Closing and opening the window clears the messages. Going to fix that. Would be easier to keep the messages using original method.

I also think that keeping the window state would have benefits.

m3nu:
Tested on macOS. Didn't see any issues. Actually nice to keep the window state when reopening it.

samuel-w:
I've decided to both keep the window state, and correctly set the borg thread parents. Technically the second one is not needed, but its nice to have.

m3nu (approved):

m3nu:
Merged. I like it how window state, logs, etc are preserved. Good change!

PR #696, opened 2020-10-31
Cleanup database at exit, use write-ahead logging

samuel-w:
https://www.tutorialspoint.com/sqlite/sqlite_vacuum.htm

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[run]
+omit = tests/*
+source=vorta
+relative_files=true
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
 
       matrix:
-        python-version: [3.6, 3.7, 3.8]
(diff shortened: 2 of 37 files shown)


m3nu (dismissed):

m3nu:
We should look into async/await. Maybe some startup tasks can be offloaded.

samuel-w:
Async is apparently slower for databases https://stackoverflow.com/questions/39803746/peewee-and-peewee-async-why-is-async-slower

m3nu:
Is it worth making startup slower for this low-priority task? How about doing it in the background with the scheduler or when closing?

samuel-w:
I moved it so it runs right before exiting.

m3nu (approved):

samuel-w:
Added write-ahead logging https://www.sqlite.org/wal.html. It speeds up tests a lot, 30 sec to 16 sec.

samuel-w:
Strange. The tests aren't faster in the CI, but they are faster on my local machine. Still useful though.

m3nu (approved):

m3nu (approved):

PR #706, opened 2020-11-07
Ignore flake8 W503

samuel-w:
Patches https://gitlab.com/pycqa/flake8/-/issues/139, 
since W503 will not be removed https://gitlab.com/pycqa/flake8/-/issues/539


samuel-w:
Why don't we use the default flake8 ignores, since W503 is already in there. The `ignore =` in setup.cfg disables it

husisusi (approved):

m3nu (approved):

PR #710, opened 2020-11-13
Clean up autostart

samuel-w:
Removed os library, so it only uses pathlib


Hofer-Julian (approved):
Changes are fine and tests succeed

PR #721, opened 2020-11-19
Refactor add and rename to not use exec_

samuel-w:
exec_ not recommended per documentation https://doc.qt.io/qt-5/qdialog.html#exec.
Now the tests run on the actual client

rootdir is copied from #548


m3nu (approved):

m3nu:
Nice to see some refactoring.

PR #722, opened 2020-11-19
Remove last exec_ instances, add archive deletion test, retain archive deletion message

samuel-w:
That should be the last of the dialog exec_.

I'm doing this to prepare for PyQt6 https://riverbankcomputing.com/software/pyqt/download.


m3nu (approved):

m3nu:
Thanks! Looks like a good cleanup.

PR #723, opened 2020-11-21
Link ssh key to repository, allow unsetting of profile repository.

samuel-w:
It makes more sense, since ssh keys are used to access remote repositories.

However, it does delete the existing ssh key setting per profile. I don't think its feasible to migrate, since a two profiles can have different ssh keys and be linked to the same repo.

### DIFF
--- a/src/vorta/borg/borg_thread.py
+++ b/src/vorta/borg/borg_thread.py
@@ -142,7 +142,7 @@ class BorgThread(QtCore.QThread, BackupProfileMixin):
                 logger.warning('Found password in database, but secure storage was available. '
                                'Consider re-adding the repo to use it.')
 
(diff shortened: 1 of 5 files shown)


m3nu:
There are arguments for both profile>sshKey and repo>sshKey. Would be good to discuss such changes before adding code.

What are the benefits of this change, except to break existing config?

samuel-w:
I see it like this:
1. Repo>SSH Key
Convenient for multiple remotes, multiple key
2. Profile>SSH Key
Convenient for multiple remotes, same key

The behaviour of 1 can be done using multiple profiles but is a bit clunky.

Both are kind of a set once thing, so now that I think about it its more annoying to lose all your key settings out of nowhere. Closing.

PR #729, opened 2020-11-27
Sort archive sizes in archive tab

samuel-w:


m3nu:
So this allows sorting by size or always sorts by size?

samuel-w:
It sorts by size correctly. Before it alphabetically sorted the size strings leading to the wrong result.

samuel-w:
Also it pushes the empty rows to the bottom when sorting descendingly, and pushes them to the top when sorting ascendingly

m3nu:
I think the sorting should be by date by default since this is how Borg sorts. By size isn't meaningful.

Sorting by alphabet was an approximation of date-based sorting. Maybe not fully correct, so I'm open to improve on that. By size doesn't make sense, since I will generally look for an archive done on a specific day. I really don't care about the size. I just want my data back on a specific day.

Of course sorting using other columns could be possible too and there may be a case where size is useful.

samuel-w:
This doesn't change the sorting behaviour of anything else. It just makes the sizes sort correctly.
Before:
Before
After:
After

m3nu:
Great. That was my question before. If you fixed sorting by size, that's perfect.

m3nu (approved):

PR #734, opened 2020-11-30
CI: Add python 3.9, labels, disable macOS mount

samuel-w:


m3nu (approved):

m3nu:
Can you just exclude the mount test from platform `darwin`? Like we exclude macOS from DBus tests. Thanks!

samuel-w:
Should be good now.

m3nu (approved):

m3nu:
Thank you, Sir!

PR #741, opened 2020-12-08
Sort tree dialogs alphabetically

### LINKED ISSUE #646: archive: extract: file order, folder order, hidden file order
When we want to extract a folder or file from the archive, we get to a window, where we can select the files and folders to extract.

My expectation was, that the order of the files and folders in a folder is like this:

Either:
A) First the folders by alphabet and then the files by alphabet.
Or:
B) All files and folders mixed, but strictly alphabetical.

But what I found is different. I can not find a rule, how it is ordered.

Folders and files are mixed.
Either folder nor files nor the mix is strictly alphabetical:
Vorta file order

It should be ordered in a way, that we can find a particular file or folder quickly.
If this has any appropriate order, please let me know.

If not, I would suggest, to implement method A) . See above.

### LINKED ISSUE #664: Add sorting to the restore window
Files are randomly ordered in the restore window. It would be better to at least have them sorted alphabetically, or even better sort them using the column headers.

samuel-w:
Fixes #646, fixes #664.

Sorts both diff and restore due to inheritance. Puts folders first then files, but doesn't allow changing of sorting order.

### DIFF
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -7,7 +7,7 @@ import platform
 import re
 import sys
 import unicodedata
-from collections import defaultdict, OrderedDict
+from collections import defaultdict
 from datetime import datetime as dt
 from functools import reduce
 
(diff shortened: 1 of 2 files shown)


samuel-w:
It moves file of size zero to the top, since folders and files have the same size.

m3nu (approved):

PR #757, opened 2021-01-14
Fix label colour not changing on dark mode change

### LINKED ISSUE #756: Labels not using text color from Plasma color scheme
**Describe the bug**
When using a Plasma color scheme on KDE, the labels "Repository:", "SSH Key:" and "Compression:" from the Repository tab still show up as black instead of using the color provided by the theme.

img-2021-01-14-185055

**To Reproduce**
1. Open the KDE system settings (`systemsettings5`)
2. Go to the "Colors" tab
3. Change the theme to a dark theme so the text colour is set to something bright

I don't really know what `systemsettings5` does in the background 

- OS: Manjaro
- Vorta version: 0.7.1
- Installed from: AUR via pamac: https://aur.archlinux.org/packages/vorta/

**Additional context**
I don't really know what happens in the background with Plasma color schemes, or whether they automatically set up a gtk3-theme as well, but I'm sure this can be reproduced on other distros with KDE and maybe even other desktop environments.

samuel-w:
This makes the labels slightly off center, but its clear enough which labels go where and most Qt themes seem to automatically align the text anyways. Fixes #756

### DIFF
--- a/src/vorta/assets/UI/repotab.ui
+++ b/src/vorta/assets/UI/repotab.ui
@@ -51,14 +51,11 @@
          <height>0</height>
         </size>
        </property>
-       <property name="styleSheet">
-        <string notr="true">margin-top: 4;</string>
-       </property>
        <property name="text">
         <string>Repository:</string>
        </property>
        <property name="alignment">
-        <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignTop</set>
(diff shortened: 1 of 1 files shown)


m3nu:
Looks good. Never had the darkmode issue on macOS, but good if it fixes the reported issue.

Alignment is the same as before.

Screen Shot 2021-01-15 at 09 08 46

m3nu (approved):

samuel-w:
With the fusion theme its slightly off.
align

Issue #758, opened 2021-01-15
Can't select multiple folders


sbaragnaus2:
**Describe the bug**
I can't add multiple sources directories at once (CTRL+left mouse button). I don't know what's wrong, because I can add multiple files add once. 
_If you had a drive with 100 folders and you have to add 70 folders it's a time consuming task._ 

**To Reproduce**
1. Go to 'sources' tab
2. Click on 'add folder'
4. See error: I can't select multiple folders

**Desktop (please complete the following information):**
- OS: Fedora 33 Kde spin (Plasma 5.20.5, Frameworks 5.78, Qt 5.15.2)
- Vorta version: 0.7.1
- Installed from: Flathub

This also happens on Solus Plasma with Vorta installed from Solus repository

**Additional context**
n/a

samuel-w:
Unfortunately there is no way to do this, Qt supports selecting only one folder at a time. As an alternative, I'll modify the paste files/folders button so that you can copy many folders and press paste to input all of them.

PR #759, opened 2021-01-15
Change paste to allow copying files/folders from file manager, set DB defaults

samuel-w:
Possibly fixes https://github.com/borgbase/vorta/issues/758. Also changed default behaviour of add_source_to_table, fixed paste button.

Edit: The test is because paste caused a crash due to the new source tab behaviour.


m3nu:
Looks good, but can't test it.

Hofer-Julian (approved):
Works fine on GNOME with nautilus

PR #771, opened 2021-01-19
Clean up supporting files

m3nu:


m3nu:
Moved `coverage` settings and dependency to `setup.cfg` and `requirements.d/dev.txt`. @samuel-w

setup.cfg
```diff
@@ -62,6 +63,8 @@ filterwarnings =
 
 [coverage:run]
 source = src
+omit = tests/*
+relative_files = true
```
samuel-w, line 67:
```suggestion
omit = tests/*
source = vorta
relative_files = true
```
`source = vorta` has to be there, otherwise coverage gathers no data.

m3nu: Great. Changed. Will see if it works on this repo here.

PR #772, opened 2021-01-19
Re-enable buttons on cancel

samuel-w:
Solves this problem: https://github.com/borgbase/vorta/issues/332#issuecomment-762762559


m3nu (approved):

m3nu:
Thanks. 🤩

For 0.7.2 let's focus on stability. Then more risky PRs after the release.

PR #782, opened 2021-02-06
Reset table on refresh

### LINKED ISSUE #781: Uncaught exception while deleting an archive
the following error popped up while deleting an archive. The only things that sticks out to me is that I had the archive list sorted by name when this happened.

  File "/app/lib/python3.7/site-packages/vorta/views/archive_tab.py", line 442, in delete_result
    deleted_row = self.archiveTable.findItems(self.archive_name, QtCore.Qt.MatchExactly)[0].row()
IndexError: list index out of range

samuel-w:
Fixes #781

### DIFF
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -111,6 +111,7 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         """Populate archive list and prune settings from profile."""
         profile = self.profile()
         if profile.repo is not None:
+            self.archiveTable.setRowCount(0)  # Clear entries
             self.mount_points = get_mount_points(profile.repo.url)
             self.toolBox.setItemText(0, self.tr('Archives for %s') % profile.repo.url)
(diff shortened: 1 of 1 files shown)


m3nu:
This doesn't fix #781 and I wouldn't know how. `populate_from_profile()` is not run again after deleting a repo.

A better solution may be to make picking a line more stable. We may have a bad way of doing that.

Error after deleting two archives and randomly sorting some cols.

```
2021-02-06 15:40:31,417 - root - CRITICAL - Uncaught exception, file a report at https://github.com/borgbase/vorta/issues/new
  File "/Users/manu/Workspace/vorta/src/vorta/views/archive_tab.py", line 443, in delete_result
    deleted_row = self.archiveTable.findItems(self.archive_name, QtCore.Qt.MatchExactly)[0].row()
IndexError: list index out of range
```

samuel-w:
It does for me. Deleting archives causes duplication. This solves it.

Before:
Before

After:
After

m3nu:
You also need to do some col sorting to trigger the bug. Just deleting something isn’t enough.

Issue #783, opened 2021-02-06
Sorted archive list gets corrupted on refresh


rblenis:
**Describe the bug**
Sorting the archive list will result in corrupted list after it is refreshed.

**To Reproduce**
1. Sort the archive list by clicking on one of the columns, such as 'Name'.
2.  Then refresh the list by clicking delete, prune or refresh buttons.
3. The updated list will be corrupt.
4. 
![Screenshot from 2021-02-06 11-29-06](https://user-images.githubusercontent.com/10763518/107123964-a93d7300-686e-11eb-8dbc-d74d8160d4ba.png)

**Desktop (please complete the following information):**
- OS: linux (Ubuntu 20.10)
- Vorta version: 0.7.2 (borg 1.1.15)

- Installed from: Ubuntu "Software" app (flatpak)

m3nu:
Seems to be the same as #781, which also edits the table. At least your PR fixes that as well. 👍

m3nu:
Closing this, since #784 seems to fix it too.

PR #784, opened 2021-02-06
fix corrupted archive table list during refresh if sorting is enabled

rblenis:
fix issue #783


rblenis:
The reason for disabling sorting while setting the row items is to prevent sorting while changing items.  Here is a section of documentation for QTableWidget::setItem()

"Note that if sorting is enabled (see sortingEnabled) and column is the current sort column, the row will be moved to the sorted position determined by item.

If you want to set several items of a particular row (say, by calling setItem() in a loop), you may want to turn off sorting before doing so, and turn it back on afterwards; this will allow you to use the same row argument for all items in the same row (i.e. setItem() will not move the row)."

m3nu:
Interesting. So this ensures the table isn't messed up while editing it?

It seems to solve the reported bug. I couldn't trigger it while randomly sorting and deleting archives.

This solves it for you as well, @samuel-w ?

m3nu (approved):

m3nu:
Merged. Thanks for solving this! 👍

PR #861, opened 2021-02-16
Simplify default archive name

m3nu:
This makes the default timestamp for archive names used more readable and is inspired by Apple's snapshots:

```
$ tmutil deletelocalsnapshots 2021-02-16-075646
Deleted local snapshot '2021-02-16-075646'
```


rblenis:
It looks like this PR also changed the default prune prefix, which could be bad. Removing the "-{profile-slug}" portion means pruning could remove archives from other backup sets, which is not expected!

m3nu:
Thought about this too. I'll add a note about it in the release logs.

The current change just brings us in line with the default of Borg and Borgmatic's archive naming and nothing changes for existing backup profiles. The new default only applies to *newly created* backup profiles.

But thanks for checking. If you like, I'll run all my PRs by you first. None of the usual contributors has time to review things currently.

rblenis:
Thanks, but don't run your PR's by me. I'm not sure I'll have the bandwidth!  I've got way too much work, and a few other projects that need some attention too. I'll keep checking in periodically, though.

PR #867, opened 2021-02-17
Improve break-lock dialog.

m3nu:
Ensure user thinks twice before breaking a lock.


ThomasWaldmann:
Much better! Users should not think that breaking locks is usual procedure.

PR #877, opened 2021-02-21
Test fixes, avoid segfault at end

m3nu:
Fixes #456
Fixes #748

Issue #338, opened 2019-07-29
After deletion of a backup, I still see cached statistic data of the previous backup


Wikinaut:
After deletion of a backup, I still see cached statistic data of the previous backup even after stopping Vorta and restarting.

The Archive page is empty (as expected).

![grafik](https://user-images.githubusercontent.com/1151915/62084741-4707b900-b25a-11e9-806d-e00ea0d780d0.png)

![grafik](https://user-images.githubusercontent.com/1151915/62084902-9fd75180-b25a-11e9-8754-fc53b6211c6e.png)

* Vorta 0.6.21
* borgbackup 1.1.5

m3nu:
We only get those numbers form `borg create` and `borg info`. That means we need to run 3 commands when you delete a repo. (we already run `borg list` after each delete or prune.)

Worth it to have those numbers right?

sudwhiwdh:
Sounds reasonable!

m3nu:
I suggest to just clear those repo numbers on prune or delete and show `N/A` or something. Then after the next create, the correct numbers will be avail again.

sudwhiwdh:
show `N/A` would be nice, yes!

PR #878, opened 2021-02-21
add option to refresh individual archives

rblenis:
allow individual archives to do a more complete refresh, pulling in/updating size and duration. (see discussion #828)

### DIFF
--- a/src/vorta/borg/info_archive.py
+++ b/src/vorta/borg/info_archive.py
@@ -2,7 +2,7 @@ from .borg_thread import BorgThread
 from vorta.models import ArchiveModel, RepoModel
 
 
-class BorgRefreshArchiveThread(BorgThread):
+class BorgInfoArchiveThread(BorgThread):
 
     def started_event(self):
         self.app.backup_started_event.emit()
--- a/src/vorta/borg/info_repo.py
+++ b/src/vorta/borg/info_repo.py
@@ -3,7 +3,7 @@ from vorta.i18n import trans_late
 from vorta.models import RepoModel
 
 
-class BorgInfoThread(BorgThread):
+class BorgInfoRepoThread(BorgThread):
 
     def started_event(self):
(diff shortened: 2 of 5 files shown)


m3nu:
Minor comment on renaming the file and class. Rest looks great. Thanks for the useful addition!

rblenis:
Added parsing of repo stats that comes back with the archive results - this provides a solution to issue #338 as well, though doing a 'borg info' after prune and delete should also be added to fully address #338.

rblenis:
@m3nu, what are you suggesting I rename the file and class to?

rblenis:
not sure why checks timed-out. I reran the action in my fork and it completed, but can't figure out how to get it to re-run in the PR unless I push a dummy change.

m3nu (commented):
> not sure why checks timed-out. I reran the action in my fork and it completed, but can't figure out how to get it to re-run in the PR unless I push a dummy change.

Some macOS checks can hang *after* the tests were successful. I'm not sure about the reason, but addressed a few possible reasons in #877. Could be the combo of pytest + coverage when using threads.

m3nu:
> @m3nu, what are you suggesting I rename the file and class to?

Usually we name our commands after the Borg subcommand used. In `vorta.borg` we already have:

- `list_archive.py`: list files in an archive
- `list_repo.py`: list archives in a repo
- `info`: list info on a repo

So the consistent choice would be to rename `info.py` to `info_repo.py` and your new module to `info_archive.py`?

m3nu:
The renaming is mostly logistics. I really like this PR because it only changes what's necessary and is easy to understand. If you prefer, I can do the renaming before merging.

m3nu:
Is it OK if I finish the renaming and then just merge?

rblenis:
sure

m3nu (approved):

m3nu:
Thanks for the contribution. Merged.

PR #889, opened 2021-02-24
Improve kwallet DBus detection

samuel-w:
Forgot that non-empty strings cast to True.


m3nu (approved):
Was never able to reproduce any error, but maybe this helps a few people.

PR #898, opened 2021-02-27
Allows to fully disable using the system keychain.

m3nu:
This PR tries to avoid common issues with system keychains that were reported:

- Add option to avoid using system keychain.
- Prioritize keychains first. Then try to start them.
- Maintenance: Possible fixes for hung tests and segfault on exit on some setups.

To test:

`$ pip install --force-reinstall git+https://github.com/m3nu/vorta.git@issue/make-keyring-optional`

In *Misc* tab disable the setting "Store repository passwords in system keychain...". After that all repo passwords will be stored with Vorta only, independent of the system keychain.

Screen Shot 2021-02-27 at 10 58 39

### DIFF
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -22,7 +22,6 @@ from PyQt5.QtWidgets import QApplication, QFileDialog, QSystemTrayIcon
 
 from vorta.borg._compatibility import BorgCompatibility
 from vorta.i18n import trans_late
-from vorta.keyring.abc import VortaKeyring
 from vorta.log import logger
 from vorta.network_status.abc import NetworkStatusMonitor
 
@@ -337,14 +336,3 @@ def validate_passwords(first_pass, second_pass):
         return trans_late('utils', "Passwords must be greater than 8 characters long.")
 
     return ""
-
-
-def display_password_backend(encryption):
-    ''' Display password backend message based off current keyring '''
-    # flake8: noqa E501
-    if encryption != 'none':
-        keyring = VortaKeyring.get_keyring()
-        return trans_late('utils', "Storing the password in your password manager.") if keyring.is_system else trans_late(
-            'utils', 'Saving the password to disk. To store password more securely install a supported secret store such as KeepassXC')
-    else:
-        return ""
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -344,7 +344,7 @@ def display_password_backend(encryption):
     # flake8: noqa E501
     if encryption != 'none':
         keyring = VortaKeyring.get_keyring()
-        return trans_late('utils', "Storing the password in your password manager.") if keyring.is_primary else trans_late(
+        return trans_late('utils', "Storing the password in your password manager.") if keyring.is_system else trans_late(
             'utils', 'Saving the password to disk. To store password more securely install a supported secret store such as KeepassXC')
     else:
         return ""
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,8 +48,8 @@ jobs:
         pip install -e .
         pip install -r requirements.d/dev.txt
 
-    - name: Setup tmate session
-      uses: mxschmitt/action-tmate@v1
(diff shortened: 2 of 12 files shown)


m3nu:
@samuel-w . This allows disabling keychains due to the many issues we have with them and some users don't want to use them at all.

samuel-w:
This feels like a workaround to the buggy code, and not an actual fix. I'll just remove kwallet.

m3nu:
Not really. People have actually asked to provide this option. Even myself, I don't really use the macOS keychain and back it up. And on Linux often a keychain is installed, but people don't use it.

src/vorta/utils.py
```diff
@@ -344,7 +344,7 @@ def display_password_backend(encryption):
     # flake8: noqa E501
     if encryption != 'none':
         keyring = VortaKeyring.get_keyring()
-        return trans_late('utils', "Storing the password in your password manager.") if keyring.is_primary else trans_late(
+        return trans_late('utils', "Storing the password in your password manager.") if keyring.is_system else trans_late(
```
samuel-w, line 347: '''
    keyring = VortaKeyring.get_keyring()
    if encryption != 'none' and not keyring.is_system:
        return trans_late('utils', "Storing the password in your password manager.") if keyring.is_primary else trans_late(
            'utils', 'Saving the password to disk. To store password more securely install a supported secret store such as KeepassXC')
    else:
        return ""
'''
Otherwise it always shows as using password manager when system keychain disabled.

samuel-w:
We should break up the settings into subcategories, its getting long.

src/vorta/utils.py
m3nu: True. Will change.

m3nu: Looks convoluted. Moved this to the view where it's actually used. We have too many circular imports already.

samuel-w:
Don't move the backend, I'm reusing it in key management

src/vorta/utils.py
samuel-w: Don't move the backend, I'm reusing it in key management. Utils prevents circular imports.

m3nu: Sorry, but I didn't see any other uses of `display_password_backend()`. Which file do you mean?

samuel-w: It's a pull request.

m3nu: I see. Which one? Duplicating a few lines isn't an issue.

Next task I want to do is to make imports more hierachical. Many imports are only possible within functions because everything imports from everywhere. So e.g. I don't want models or utils to import too much. They should be at the bottom.

I may have fixed the "hanger" during macOS testing as well. According to the debugger it was stuck waiting for the BorgThread mutex. So I tried a few things there, like using the native Python mutex, which has a nicer API.

m3nu: Never mind. Found `display_password_backend()` in your PR. Will try to put that function into the keyring class instead.

m3nu: Moved to `VortaKeyring().get_backend_warning()`

m3nu:
You notice any other issues? Else I merge it tomorrow to encourage some testing in the master branch.

Also one more [request](https://github.com/borgbase/vorta/issues/903#issuecomment-787448733) to select the keychain. So giving the option to turn it off is probably just a first step.

m3nu:
I also put back Coverage for macOS, since it didn't make a difference after all. 😅

m3nu:
@samuel-w , we now use KWallet only if it's KDE. There were too many reports where people had it installed, but didn't use it. python-keyring has roughly the [same rule](https://github.com/jaraco/keyring/blob/cca9efce749b273b453573d047b887382d8c6896/keyring/backends/kwallet.py#L56).

Would appreciate your feedback.

m3nu:
Did a larger rewrite on how keyrings are selected. This avoids the chaotic if-else and just assigns a priority within each keyring. @samuel-w

samuel-w (approved):

Issue #190, opened 2019-02-12
Remove keyring package for Linux


yatesco:
**Describe the bug**
`pip3 install vorta --user` has a pretty impressive back-trace when starting (see the stack trace at the bottom). `coliny@dev-os:~$ pip3 install --upgrade keyrings.alt` makes the world a better place:

```
Collecting keyrings.alt
  Downloading https://files.pythonhosted.org/packages/f2/97/e0601ee8e936a13245686af0ac9ed2e2bb61706b4155e4e629cf063ad248/keyrings.alt-3.1.1-py2.py3-none-any.whl
Collecting six (from keyrings.alt)
  Using cached https://files.pythonhosted.org/packages/73/fb/00a976f728d0d1fecfe898238ce23f502a721c0ac0ecfedb80e0d88c64e9/six-1.12.0-py2.py3-none-any.whl
Installing collected packages: six, keyrings.alt
Successfully installed keyrings.alt-3.1.1 six-1.12.0
coliny@dev-os:~$ 
coliny@dev-os:~$ ./.local/bin/vorta 
Forking to background (see system tray).
```

**To Reproduce**
**Desktop (please complete the following information):**
OS: Ubuntu Budgie 18.10 - all up to date and `pip3 install vorta --user`

**Additional context**
```
coliny@dev-os:~$ pip3 install vorta --user
Collecting vorta
  Cache entry deserialization failed, entry ignored
  Cache entry deserialization failed, entry ignored
  Downloading https://files.pythonhosted.org/packages/45/c4/8b5ad0d7547a48f229ad1feef582d83bf16c399dd7f4c89b1ff1c21a8aa6/vorta-0.6.8.tar.gz (167kB)
    100% |████████████████████████████████| 174kB 4.3MB/s 
Collecting appdirs (from vorta)
  Cache entry deserialization failed, entry ignored
  Cache entry deserialization failed, entry ignored
  Downloading https://files.pythonhosted.org/packages/56/eb/810e700ed1349edde4cbdc1b2a21e28cdf115f9faf263f6bbf8447c1abf3/appdirs-1.4.3-py2.py3-none-any.whl
Collecting apscheduler (from vorta)
  Cache entry deserialization failed, entry ignored
  Cache entry deserialization failed, entry ignored
[...]

PR #900, opened 2021-02-27
Fix kwallet again

samuel-w:
One last try to fixing kwallet. If this doesn't work just remove the kwallet code, its causing more bugs than solutions.

Slightly based it off python-keyring's implementation, https://github.com/jaraco/keyring/blob/main/keyring/backends/kwallet.py#L96.

### DIFF
--- a/src/vorta/keyring/kwallet.py
+++ b/src/vorta/keyring/kwallet.py
@@ -42,18 +42,20 @@ class VortaKWallet5Keyring(VortaKeyring):
 
     @property
     def is_unlocked(self):
-        self.try_unlock()
-        return self.handle >= 0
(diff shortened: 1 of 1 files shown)


samuel-w:
Wait, why don't we just use python-keyring?

m3nu:
Doesn't change the functionality of the code, right? Just re-arranges a few things.

m3nu:
> Wait, why don't we just use python-keyring?

We did in the beginning, but there were other issues, like Pyobjc segfaults.

https://github.com/borgbase/vorta/issues/190#issuecomment-463037623

m3nu:
Thanks, cherry-picked a change into #898.

PR #904, opened 2021-03-01
Disable codecov comments

samuel-w:


m3nu:
Good idea.

m3nu:
I usually don't look at those. Would prefer to do a sprint now- and then to add missing tests.

Issue #901, opened 2021-02-27
'Diff' command does not handle filename/paths with CR/LF characters in them.


rblenis:
If you backup files with a CR or LF in the name or path, then performing a diff will not correctly show a directory tree of the diffs. The CR/LF can cause the later portion of the filename to be interpreted as another file, etc. The problem needs to be addressed in borg first, since it does not support json output mode for the diff command. It's probably impossible to properly parse all legal filenames until json mode is supported in the diff command.

OS: linux (Ubuntu 20.10
Vorta version: 0.7.2
Installed from: 1.1.15

I split this off from issue #883, since the filename problem affects two borg commands, and the list/extract one can be fixed without new borg features.

m3nu:
Good stuff, thanks for taking care.

m3nu:
Fixed by @rblenis via #909 when Borg 1.16+ is used.

PR #909, opened 2021-03-02
add support for 'diff --json-lines', available in borg >= 1.1.16

rblenis:
fixes issue #901 (diff not handling CR/LF in filenames)


Hofer-Julian:
Nice!
Did not know that borg diff supports json output now 😊

rblenis:
yes, hopefully my PR on borg will be completed soon. I think they're trying to get it in for the 1.1.16 release.

rblenis:
borg 1.1.16 released with 'diff --json-lines' support.

rblenis:
@m3nu, should the test workflow also use several versions of borg, this way some features which behave differently depending on borg version (like this feature), will get thoroughly testeed?

m3nu:
Most calls to Borg are mocked and it just uses pre-defined json output files in `/tests/borg_json_output`.

m3nu:
Yeah, let's get this feature out. I was waiting with merges until the Debian package is frozen. That should be done by now.

m3nu:
That would be perfect. Thanks!

rblenis:
I added some tests, copying the non-json-lines test.  I also made small changes to report the changes in the tree dialog to match the non-json-lines.  Later, I may look at improving the difference info displayed in the tree, especially when multiple types of changes occur to the same item.

m3nu:
Looks all good to me. This is ready to merge?

rblenis:
yes it's ready.

src/vorta/utils.py
```diff
@@ -169,20 +169,21 @@ def sort_sizes(size_list):
     return final_list
 
 
-def pretty_bytes(size):
-    """from https://stackoverflow.com/questions/12523586/
-            python-format-size-application-converting-b-to-kb-mb-gb-tb/37423778"""
+def pretty_bytes(size, metric=True, sign=False, precision=1):
```
m3nu, line 172: Nice improvements to `pretty_bytes()`. But you don't use the new `metric` and `sign` args anywhere?

m3nu: I also removed the SO reference, since the function substantially changed.

src/vorta/borg/diff.py
```diff
@@ -17,17 +18,17 @@ def prepare(cls, profile, archive_name_1, archive_name_2):
...
+            'borg', 'diff', '--info', '--log-json',
+            f'{profile.repo.url}::{archive_name_1}',
+            f'{archive_name_2}']
         ret['ok'] = True
-        ret['cmd'] = cmd
+        ret['archive_name_older'] = archive_name_1
+        ret['archive_name_newer'] = archive_name_2
+        ret['json_lines'] = False
+        if borg_compat.check('DIFF_JSON_LINES'):
+            ret['cmd'].insert(4, '--json-lines')
```
m3nu: This is dangerous because the hard-coded position is often overlooked when someone makes changes later. Minor change to get rid of it.

m3nu:
Few administrative edits.

src/vorta/utils.py
rblenis: In an earlier of my changes, I was using some of the enhancements to pretty_bytes, but after adding the test code and realizing the displayed tree did not match exactly with the non-json-lines version, I didn't use it, and forgot to revert those changes.  There  are two 'bug-fixes', but I don't think either one would be triggered. 1. better handling of something larger than 1000TB - my version with the loop index check will at least attempt to print the number using 'TB' units, instead of NaN (due to array index exception). 2. better handling of negative 'size' - print correct value with appropriate units instead of using base/unscaled units ('B'). Currently, I think the tree only shows the '+' change in file content, and ignores the '-' change in file content, so #2 probably isn't triggered either. And then a 3rd 'issue', is when a number is close to the threshold of using the next larger units, but doesn't use them, but then due to rounding prints as '1000' ...  Example: size = 999.95 KB -- it is not large enough to go through the loop twice and use 'MB' units, but when printed, it is rounded up and prints as '1000.0 KB' instead of  '1.0 MB'. It seems like the preferred display would limit the significand to 4 total digits and not sometimes round up and display 5 digits.

rblenis: I could split off the pretty_bytes changes as a separate PR and indicate the issues it addresses.

m3nu: Nah, it's fine. The new args might be useful in the future and it fixes some issues as you say.

m3nu (approved):

PR #910, opened 2021-03-02
fix 'list --json-lines' command when only a single result line is out…

rblenis:
if only a single line of output comes from borg --json-lines, then the default processing will already have converted it to a dict, whereas if multiple lines are output, it leaves it as a string and lets the command specific code deal with it.  This fixes the immediate problem of handling it in the 'list --json-lines' command, but will need to handle in the 'diff --json-lines' option when added to borg.

### DIFF
--- a/src/vorta/views/extract_dialog.py
+++ b/src/vorta/views/extract_dialog.py
@@ -18,12 +18,10 @@ class ExtractDialog(ExtractDialogBase, ExtractDialogUI):
         super().__init__()
         self.setupUi(self)
 
-        files_with_attributes = []
         nested_file_list = nested_dict()
         self.selected = set()
 
-        def parse_json_line(line):
-            data = json.loads(line)
+        def parse_json_line(data):
             size = data["size"]
             # python >= 3.7
             # modified = datetime.fromisoformat(data["mtime"]).ctime()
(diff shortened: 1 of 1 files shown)


src/vorta/views/extract_dialog.py
```diff
@@ -39,11 +37,11 @@ def parse_json_line(line):
                 d[name] = {}
             return size, modified, name, dirpath
 
-        for line in fs_data.split("\n"):
-            try:
-                files_with_attributes.append(parse_json_line(line))
-            except ValueError:
-                pass
+        # handle case of a single line of result, which will already be a dict
```
m3nu, line 40: What happens to the value error?

List comprehension will be faster than a for-loop. So that's good.

I do worry about the `isinstance()` check in the inner loop. Maybe it's possible to deal with this special case of one line outside the inner loop?

rblenis: previously, I passed every line to json.loads, which would through error on empty line, but now I only pass non-empty lines to json parser.

rblenis: The isinstance is only done once, to determine if the entire fs_data object is a string or a dict, then if it is a dict, put in a single item list (only a single line returned from borg), or split the string and convert each line to dict.

m3nu: True. Got it.

m3nu (dismissed):

m3nu:
I'll merge this next week, since the Debian guys asked me to do a feature freeze until Mar 10.

m3nu:
Thanks for the addition! 🙏

PR #945, opened 2021-04-09
Open window on left-click. Fixes #938

### LINKED ISSUE #938: Left-click doesn't open tray menu on Gnome
**Describe the bug**
clicking on the notification bar icon doesnt open the application

**To Reproduce**
1. have vorta running, so an icon in the notification area is displayed
2. Click on the notification icon, nothing happens

expectation is to get vorta to open, instead i need to right click and select "Vorta for Borg Backup"

- OS: Debian unstable
- Vorta version: 0.7.5
- Installed from: debian package

m3nu:
XDG-compliant tray behavior: right click to open a context menu, and left click to restore the application.

If XDG_CURRENT_DESKTOP isn't set, always open the tray menu (macOS)

### DIFF
--- a/src/vorta/tray_menu.py
+++ b/src/vorta/tray_menu.py
@@ -25,7 +25,7 @@ class TrayMenu(QSystemTrayIcon):
 
     def on_activation(self, reason):
         if reason == QSystemTrayIcon.Trigger:
-            if os.environ.get('XDG_CURRENT_DESKTOP', '') == 'KDE':
+            if os.environ.get('XDG_CURRENT_DESKTOP', '').startswith(('KDE', 'GNOME')):
                 self.app.toggle_main_window_visibility()
             else:
                 self.on_user_click()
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
If I remember correctly, this change was introduced to accommodate KDEs human interface guidelines.
GNOME does not have something for tray icons, since tray icons are discouraged there.

If you ask for my opinion, I would find it surprising if the window would open after a left click.
It also doesn't work well with a touchpad, since the settings would then be accessible with a right click instead.

If we change the current behaviour at all, I'd rather drop the special case for KDE and go back to the default, than including GNOME into the special case.

m3nu:
Agreed. Not merging this. The main issue I saw is that on Gnome, it won't open the menu when left-clicking. Looks like a Qt issue. Didn't find a good workaround.

sandrotosi:
@Hofer-Julian why is that surprising to open the main app window when left licking on an icon? i got that exact behavior with the system monitor applet (included with Gnome); i'm using gnome flashback, and application in the tray icon are still supported there.

Hofer-Julian:
The last commit seems like a good compromise

Hofer-Julian (approved):

m3nu:
Then let's go with that. Thanks everyone for reviewing and weighing in!

Issue #940, opened 2021-04-07
diff: KeyError in get_dict_from_list when diffing archives with a removed folder that now has .nobackup


rxgh99:
**Describe the bug**
A folder was removed as a source using the source window in Vorta. The file .nobackup was touched in the removed folder and a subsequent backup was run. 

When diffing the .nobackup archive run with the previous archive (which included the folder), the following error occurred (note: from Vorta logs). In this example, the folder "Music" was removed and .nobackup placed in the folder. 

2021-04-06 23:25:36,589 - vorta.borg.borg_thread - INFO - Running command /Applications/Vorta.app/Contents/Resources/borg-dir/borg.exe diff --info --log-json xxxxxx@xxxxxx.repo.borgbase.com:repo::hostname-2021-04-06-185815 hostname-2021-04-06-190533
2021-04-06 23:25:42,110 - root - CRITICAL - Uncaught exception, file a report at https://github.com/borgbase/vorta/issues/new
Traceback (most recent call last):
  File "views/archive_tab.py", line 520, in list_diff_result
  File "views/diff_result.py", line 22, in __init__
  File "views/diff_result.py", line 94, in parse_diff_lines
  File "views/diff_result.py", line 87, in parse_line
  File "utils.py", line 112, in get_dict_from_list
KeyError: 'Music'

Note: an additional backup was run and a diff on the two most recent archives (those with the .nobackup exclude) was successful. The failure occurs diffing the transition to .nobackup. 

**To Reproduce**
1. Include a folder as a source
2. Run a backup
3. Touch .nobackup in the folder
4. Remove the folder from the source list
5. Run a backup
6. Diff the two backups

**Environment (please complete the following information):**
- OS: macOS 11.2
- Vorta version: 0.7.5
- Installed from: vorta.borgbase.com

m3nu:
The diff feature was completely rewritten for the upcoming 0.7.6 version. So the issue you are describing is probably already resolved in the current `master` branch. Can you test if the issue still occurs there?

rxgh99:
Thank you. Apologies, but not clear how to get the current latest from master branch (I was using binary install for Vorta). I'd like to test if I can get a couple of pointers on installing the latest master. I have brew installed. Thanks.

rxgh99:
Are these the recommended steps to get master branch? Thanks.

https://borgbackup.readthedocs.io/en/stable/installation.html#git-installation

rxgh99:
[...]

Issue #925, opened 2021-03-28
Uncaught exception when doing a 'Diff' between latest and 2nd latest archives


aromanelli:
Fedora 33
Vorta 0.7.5

1) Started Vorta.
2) Did a backup to an existing repo on a Seagate 5TB USB drive (that had less than 1TB space left).
3) Added a new 'Exclude Patterns' patten (/home/adrian/.gradle/wrapper/dists). Did _not_ do another backup after that!
4) Switched to 'Archives' tab.
5) Selected 'Diff' button.
6) Selected latest (newly done same run) archive and the most previous to that archive.
7) Diff process was running for a bit, then the crash happened and an error dialog was shown. No diff results were shown.

I've done other diffs in the past successfully, but never after doing step #3.

Log entry below ...
```
2021-03-27 21:10:33,745 - asyncio - DEBUG - Using selector: EpollSelector
2021-03-27 21:10:33,748 - vorta.borg.borg_thread - DEBUG - Using VortaSecretStorageKeyring keyring to store passwords.
2021-03-27 21:10:33,751 - asyncio - DEBUG - Using selector: EpollSelector
2021-03-27 21:10:33,755 - root - DEBUG - Found 1 passwords matching repo URL.
2021-03-27 21:10:33,849 - vorta.borg.borg_thread - INFO - Running command /app/bin/borg diff --info --log-json /run/media/adrian/Seagate 5TB USB/Borg Backup::593150c8 ar-laptop-fedora-default-2021-03-27T21:04:25
2021-03-27 21:11:28,772 - root - CRITICAL - Uncaught exception, file a report at https://github.com/borgbase/vorta/issues/new
Traceback (most recent call last):
  File "/app/lib/python3.8/site-packages/vorta/views/archive_tab.py", line 520, in list_diff_result
    window = DiffResult(result['data'], archive_newer, archive_older)
  File "/app/lib/python3.8/site-packages/vorta/views/diff_result.py", line 22, in __init__
    files_with_attributes, nested_file_list = parse_diff_lines(fs_data.split('\n'))
  File "/app/lib/python3.8/site-packages/vorta/views/diff_result.py", line 94, in parse_diff_lines
    files_with_attributes.append(parse_line(line))
  File "/app/lib/python3.8/site-packages/vorta/views/diff_result.py", line 87, in parse_line
    d = get_dict_from_list(nested_file_list, dir.split('/'))
[...]

PR #947, opened 2021-04-12
fix issue #940 - KeyError in get_dict_from_list

rblenis:
- cause of error : defaultdict added more defaultdict while attempting to traverse the tree, but once a 'potential leaf' node was added, it was added as a dict, not a defaultdict.

example: (annotated output from 'borg diff --info --log-json repo::archive1 archive2')
added directory     home/rblenis/tmp/borgtest                              <-- 'home','rblenis','tmp' added as defaultdict, 'borgtest' added as dict
added       6.93 kB home/rblenis/tmp/borgtest/CircleFit.cpp       <-- 'CircleFit.cpp' added as dict
added       1.61 kB home/rblenis/tmp/borgtest/CircleFit.h           <-- 'CircleFit.h' added as dict
added directory     home/rblenis/tmp/borgtest/folder2                <-- 'folder2' added as dict
added       6.93 kB home/rblenis/tmp/borgtest/folder2/CircleFit.cpp   <-- 'CircleFit.cpp' added as dict
added       1.61 kB home/rblenis/tmp/borgtest/folder2/CircleFit.h       <-- 'CirclFit.h' added as dict
removed directory   home/rblenis/tmp/borgtest/foo/bar1                     <-- 'borgtest' is dict, so trying to traverse to non existent foo/bar1 throws KeyError.

- two possible solutions:
    - 1 - change everywhere that adds a 'potential' leaf node to add a defaultdict (ie nested_dict()) - this occurs in several places, but not many.
    - 2 - change get_dict_from_list to add a default dict (not defaultdict) when traversing the tree, for the case where a multi-level node is added on top of an existing node. This requires only changing a single location, and means that the dictionaries returned by accessing the tree will behave like normal dict (ie, won't by default add missing keys).

### DIFF
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -1,13 +1,11 @@
 import argparse
 import errno
 import getpass
-import operator
 import os
 import platform
 import re
 import sys
 import unicodedata
-from collections import defaultdict
 from datetime import datetime as dt
 from functools import reduce
 
@@ -105,11 +103,11 @@ def nested_dict():
     - https://stackoverflow.com/a/16724937/3983708
     - https://stackoverflow.com/a/14692747/3983708
     """
-    return defaultdict(nested_dict)
+    return dict()
 
 
 def get_dict_from_list(dataDict, mapList):
-    return reduce(operator.getitem, mapList, dataDict)
+    return reduce(lambda d, k: d.setdefault(k, {}), mapList, dataDict)
 
 
 def choose_file_dialog(parent, title, want_folder=True):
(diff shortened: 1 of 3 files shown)


rblenis:
Also, probably addresses #925

m3nu:
Good stuff. I see you went with option 2.

I'll try to add a test for the failed case tomorrow and then merge it.

Thanks for handling it! 👍

m3nu:
I added a test case following the steps [here](https://github.com/borgbase/vorta/issues/940#issuecomment-816992935) and can confirm that this PR fixes the exception. 👍

m3nu (approved):

rblenis:
Yes, I went with option 2 so that all nodes in the nested dictionary were the same type (dict vs defaultdict), and didn't run into potential future confusion over the dictionaries auto-adding keys when not expected. This option only auto-adds keys while building the nested dictionaries.  Thanks for looking into the test case. I ran out of time this weekend, but wanted to at least present the PR so people knew it was being worked on.

m3nu:
It's perfect. 2 issues closed with 2 lines 😄

PR #954, opened 2021-04-20
Import/export settings (or bootstrap with default config)

m3nu:
Here a work-in-progress of the import/export feature. I cherry-picked the main files since other files have diverged since this was added. Left to do:

- Not necessary to export archive details, since they can be fetched from the repo.
- Discuss file name/extension to use. Currently `.vortabackup`. Maybe just `vorta-profile.json` is clearer (see example test file)
- You wanted to bootstrap from a default location. This could be triggered by looking for a specific file, if there are no repos (in `srv/vorta/models.py`) or with a command line argument. The former might be better.
- Most logic is in `backup_window.py` This needs to be linked to some button in the main window. Maybe next to the profile changer. Roughly like this:
```
def __init__():
    ...
    self.backupAction = menu.addAction("Backup profile", self.profile_backup_action)
    self.restoreAction = menu.addAction("Restore profile backup", self.profile_restore_action)
    ...

def profile_backup_action(self):
    window = BackupWindow(parent=self)
    self.window = window
    window.setParent(self, QtCore.Qt.Sheet)
    window.show()

def profile_restore_action(self):
    def profile_restored_event(new_profile, returns):
        if returns.get('repo'):
            self.repoTab.set_repos()
        if returns.get('overrideExisting'):
            self.scheduleTab.init_logs()
            self.scheduleTab.init_wifi()
            self.miscTab.populate_from_profile()
        self.add_profile_entry(new_profile.name, new_profile.id)
    window = RestoreWindow(parent=self)
    self.window = window
    window.setParent(self, QtCore.Qt.Sheet)
    window.profile_restored.connect(profile_restored_event)
    window.show()
```

### DIFF
--- a/src/vorta/assets/UI/backupwindow.ui
+++ b/src/vorta/assets/UI/backupwindow.ui
@@ -1,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>Dialog</class>
+ <widget class="QDialog" name="Dialog">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>483</width>
+    <height>245</height>
+   </rect>
+  </property>
+  <layout class="QFormLayout" name="formLayout">
+   <item row="0" column="0">
+    <widget class="QLabel" name="label_2">
+     <property name="text">
+      <string>Location:</string>
+     </property>
+    </widget>
+   </item>
+   <item row="0" column="1">
+    <layout class="QHBoxLayout" name="horizontalLayout">
+     <item>
+      <widget class="QLabel" name="locationLabel">
+       <property name="text">
+        <string>None</string>
+       </property>
+      </widget>
+     </item>
+     <item>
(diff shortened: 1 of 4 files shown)


phihos:
Thanks for putting this together.
I already started with the integration of the existing windows:
vorta-buttons
Did you have something like this in mind?
If yes, I would like to push my changes and go on with fixing the export. It currently crashes when I trigger it.

How do I add commits to this PR? Do I have permission to push to this branch?

About the backup file extension: From a technical perspective I think json is more fitting because it is actually json and not a special format invented for this program.

m3nu:
You can just work in your own branch and then open a new PR for it. I’m not aware of a better process on Github. Be sure to enable the checkbox “Allow edits by maintainers” (or similar), so I can add small changes later.

Moving less frequently accessed items to a submenu is the way to go. As you already did. 👍 The precise button arrangement can be decided later. Not sure whats best here. One option is to group import/export together. Another to group rename, delete, import, export as “Edit” or “More”.

phihos:
Done, see #955. Feel free to add/edit my TODO list.

PR #955, opened 2021-04-21
Import/export settings (or bootstrap with default config)

phihos:
Continuing the work of #954.

Left to do:

- [x] fix/finish export & import of profiles
- [x] fix tests
- [x] exclude archive details
- [x] add ability to provision vorta from file
- [ ] docs


phihos:
It took me longer than expected (last week was busy) but I made some progress. There is now a rudimentary config backup/restore functionality. I plan to support backup/restore of multiple profiles at once.

@m3nu It would be very welcome if you could take a look and tell me if I am going into the right direction.

m3nu:
Good stuff! Basic flow works. Great progress. Some thoughts (may overlap with your todo):

- File name/extension to use. Currently `.vortabackup`. Maybe just `.json` is clearer.
- Setting `$PROFILE.json` as default filename to save would be nice. Maybe it's possible to fix the extension and users can only enter the name.
- I'd strongly suggest to call the feature **Import/Export Profile** in classes, filenames and tests. Since Vorta is already a *Backup* app, calling it `BackupWindow` or `RestoreWindow` or `ConfigBackup` will be too ambiguous. Similar to Borg's `borg key export/import`. Export the key, export the profile. 🤓
- The main window is built around *Profiles*. Change the profile, all the content changes. The buttons next to the profile dropdown also do stuff with a single *profile*. So I would keep things simple and export the *current profile* over the export button. This will greatly simplify things because a) no profile selector and list is needed and b) one json file = one profile. If you want to export global settings, this could be an optional checkbox.
- You wanted to bootstrap from a default location. This could be triggered by looking for a specific file, if there are no repos (in srv/vorta/models.py) or with a command line argument. Looking in a default location **only if no profile exists** might be the least intrusive (doesn't add new CLI flags, a GUI shouldn't have CLI flags really). Default location could be `~/.vorta-init.json`?
- Optionally include password. Good. We may need a dialog to ask for the password if it's not included.
- I question whether events and archives should be included. Not useful and maybe outdated. Better if those are fetched when verifying the repo.

phihos:
Hi @m3nu,

beside the bootstrapping I think I ticked off all items on your list. Could you take another look?
Also I have a question about the event log: I understand, that the archives can be recovered from the repo, but that is not true for the event log right? So why should they be excluded from the export?

m3nu:
Perfect! Please allow me 1-2 days for testing and review.

Then we can merge in time for the next 0.7.6 release.

phihos:
Great, I still need to fix/implement tests and the bootstrap feature though. I try to get this done within the next two days also.

phihos:
Hi @m3nu, I am almost done. I only need to clarify two things:
1. Can you answer this question I had earlier?
>Also I have a question about the event log: I understand, that the archives can be recovered from the repo, but that is not true for the event log right? So why should they be excluded from the export?
2. I thought about my bootstrap use-case and at least in  my case I need to be able to overwrite existing profiles. I already implemented the possibility to overwrite a preexsiting profile with the same name when importing. I like the idea of using a fixed location with a json file for bootstrapping. Maybe we could extend that to overwriting a profile with the same name when this json file exists. After importing the file needs to be deleted by vorta of course to prevent resetting the profile every time. For me personally doing that silently would be fine, but I could also add a prompt for the userto accept before overwriting the profile from the export file. What do you think?

m3nu:
I'd say event logs aren't too important to keep around. Personally I never looked at them. We don't copy the log files either.

Overwriting the profile and deleting the import file after sounds good. Since you want to do it non-interactive, you probably **do not** want a confirmation? It's also very deliberate to place such a bootstrap file. It won't happen by accident.

phihos:
Ok I resolved the outstanding issues. Profiles located at `~/.vorta-init.json` will be imported on application startup. I resued some code from the interactive import for that. I also remove the event log from the backup.
@m3nu I think you can now review the final code if when you have the time. What about documentation? Do you wnat to take care of that yourself or would you like me to write it? And if so where?

m3nu:
Perfect. Will look at it first thing tomorrow.

I can do the docs. It’s a separate repo. Will also help me understand it from a user view.

phihos:
I messed up the tests with my last commit. They should pass now.

src/vorta/borg/create.py
```diff
@@ -51,13 +51,13 @@ def finished_event(self, result):
 
     @classmethod
     def pre_post_backup_cmd(cls, params, cmd='pre_backup_cmd', returncode=0):
-        cmd = getattr(params['profile'], cmd)
+        cmd = getattr(params['profile_export'], cmd)
```
m3nu, line 54: Search & Replace gone overboard here? Also affects a few other files.

src/vorta/views/main_window.py
```diff
@@ -163,6 +182,60 @@ def profile_add_action(self):
         window.profile_changed.connect(self.profile_add_edit_result)
         window.rejected.connect(lambda: self.profileSelector.setCurrentIndex(self.profileSelector.currentIndex()))
 
+    def profile_export_action(self):
+        window = ExportWindow(profile=self.current_profile)
```
m3nu, line 186: Suggest to use `self.profile()` here as reference to the profile saved in the DB. Else updated settings won't be saved during export.

src/vorta/borg/init.py
```diff
@@ -10,7 +10,7 @@ def started_event(self):
     @classmethod
     def prepare(cls, params):
 
-        # Build fake profile because we don't have it in the DB yet.
```
m3nu, line 13: Too much search & replace? 😜

tests/test_importexport.py
```diff
@@ -0,0 +1,110 @@
...
+        QFileDialog, "getOpenFileName", getOpenFileName
+    )
+
+    main = qapp.main_window
+    main.profile_import_action()
+    import_dialog: ImportWindow = main.window
+    import_dialog.overwriteExistingSettings.setChecked(True)
+
+    qtbot.mouseClick(import_dialog.buttonBox.button(QDialogButtonBox.Ok), QtCore.Qt.LeftButton)
+    qtbot.waitUntil(lambda: qapp.activeWindow() is None)
```
m3nu, line 27: This fixes the failing test on macOS. We just wait for the signal instead of the window.

```
import pytest
...
qtbot.waitSignal(import_dialog.profile_imported, **pytest._wait_defaults)
```

This waits for your import signal with a longer timeout, since some CI platforms can be a bit slow at times.

m3nu (changes requested):

m3nu:
Good stuff. 👍 Just noticed a few minor things. This will be ready to merge soon.

I was a bit skeptical on the Import/Export buttons, but after using it, it makes sense.

src/vorta/borg/create.py
phihos: Oh how embarassing. Fixing it ASAP.

m3nu: No worries. Can happen in the heat of battle.

src/vorta/views/main_window.py
phihos: For that to work in need to add BackupProfileMixin to the main window. Shall I do it?

phihos: I added it now and it works. But I can change it if you had something different in mind.

phihos:
> I was a bit skeptical on the Import/Export buttons, but after using it, it makes sense.
The choice of buttons is by no means final. If you think there is a better way to put the feature into the UI please go ahead. I am more on the backend-side of things and not very experienced with building GUIs.

I worked through all your comments and you can give the tests another go :-)

m3nu:
I don't think you can add `BackupProfileMixin` to the main window. Getting this error the second time calling *Export Profile*.

```
  File "/Users/manu/Workspace/vorta/src/vorta/views/main_window.py", line 186, in profile_export_action
    window = ExportWindow(profile=self.profile())
  File "/Users/manu/Workspace/vorta/src/vorta/models.py", line 188, in profile
    return BackupProfileModel.get(id=self.window().current_profile.id)
TypeError: 'ExportWindow' object is not callable
```

m3nu:
Removed `BackupProfileMixin` since we already have `profile.refresh()` to load new changes from the DB. Forgot about that.

m3nu:
Else I'm satisfied with this PR. The changes are relatively isolated and there are tests.

I still want to re-arrange the buttons next to profile and in *Sources*. That can be a separate PR.

m3nu:
One last bug when using `~/.vorta-init.json`

```
  File "/Users/manu/Workspace/vorta/.direnv/python-3.9.2/bin/vorta", line 33, in <module>
    sys.exit(load_entry_point('vorta', 'gui_scripts', 'vorta')())
  File "/Users/manu/Workspace/vorta/src/vorta/__main__.py", line 58, in main
    app = VortaApp(sys.argv, single_app=args.profile is None)
  File "/Users/manu/Workspace/vorta/src/vorta/application.py", line 66, in __init__
  File "/Users/manu/Workspace/vorta/src/vorta/views/main_window.py", line 103, in __init__
    self.bootstrap_from_export()
  File "/Users/manu/Workspace/vorta/src/vorta/views/main_window.py", line 219, in bootstrap_from_export
    profile = profile_export.to_db(overwrite_profile=True, overwrite_settings=True)
  File "/Users/manu/Workspace/vorta/src/vorta/profile_export.py", line 86, in to_db
    self._profile_dict['id'] = existing_model.id
AttributeError: 'NoneType' object has no attribute 'id'
```

m3nu:
Would be good to have a test to catch the above error when using `.vorta-init.json`.

phihos:
>Would be good to have a test to catch the above error when using .vorta-init.json.
Done. I also discovered a bug while doing it :-D
When importing a profile that has no existing profile with that name but you set `overwrite_profile` anyway the ID of the import profile is kept and a profile with the same ID will be overwritten instead. I changed the behavior so that in that case it behaves the same as with `overwrite_profile=False`.

m3nu:
Thanks a lot! Getting there. This is a complex feature to be honest. Kudos for tackling it!

(I may need longer to test this time, since my laptop is in the shop for a few days.)

m3nu:
Looks all good to me. Just rebased and added a small change to make sure the icon color changes with the color schema.

Bootstrapping with `~/.vorta-init.json` also works as expected. Just noticed that after bootstrapping there are 2 profiles (*Default* and the imported one). Is that OK for your use case or would you add only the imported profile and skip *Default*, since it's not used here. For me both options are OK. Just worth considering.

m3nu (dismissed):

m3nu:
@phihos , any comment on the [above](https://github.com/borgbase/vorta/pull/955#issuecomment-851018007)? Else I'll go ahead and merge it.

phihos:
Hi, had a busy week. Having an empty "Default" profile after bootstrapping is ok for my use-case but it does not feel right. I will create another commit fixing that now.

m3nu:
Agree. Will be more clear, if only the bootstrapped profile is there. Thanks!

After this is merged, I'll do the next release. And then re-arranging some buttons and bug fixes for the one after.

phihos:
The change was more extensive than anticipated. I did not want to delete the "Default" profile when bootstrapping but prevent its creation instead. Else I fear there could be some unexpected side effects.
What I did:
* I moved the creation of the defaut profile from init_db to the main window.
* I run init_db at the end of `ProfileExport.to_db` to fix another issue I encountered: When the export does not contain all the settings the missing setting will not be recreated until after application restart.
* The main window has now an optional parameter for where to look for the export file. If it does not exist bootstrapping is silently skipped as before.
* I changed the init_db pytest fixture to check for a pytest mark "profile_bootstrap_file" telling it to skip creating default data and initializing the main window with an export file path instead.
* I split the existing bootstrap test into two tests: The first only checks if the bootstrap file is deleted properly after import. The second one tells the init_db fixture to use a valid export file for bootstrapping and checks if the app imported it correctly. There is also an assert checking if the bootstrapped profile is the only existing profile.

Beside the tests I also checked the behavior manually. When starting Vorta for the first time and bootstrapping from .vorta-init.json we only have the imported profile. When starting Vorta not the first time and we bootstrap from file preexisting profiles are preserved (if the imported profile name != existing profile name of course).

m3nu:
Another option is to only do bootstrapping, if no profile exists yet? Since overwriting things may be surprising for users, if you place the file via MDM.

Reasons for moving stuff around are clear. I'd prefer to have less stuff in MainWindow, but that may make it more complicated. Will see.

phihos:
> Another option is to only do bootstrapping, if no profile exists yet?

At least in my use-case it causes more confusion: I work with a self-service app, and I plan to add an item there like "Setup Backup" or something. It should install Vorta and place the bootstrap file. The item the changes to "Update Backup" and when clicked it re-places the bootstrap file. This is for botched configurations or when the backup server's port changes or similar.
If we only do bootstrapping when no profile exists then I would either have to delet the whole database beforehand or tell users to delete the profile before clicking "Update Backup". Both is not ideal I think.
How would you solve this particular case?

m3nu:
> How would you solve this particular case?

I don't have a better solution either. Overwriting it is OK, since placing this file is very intentional.

Will continue to understand the changes and merge it now.

m3nu (approved):

m3nu:
Re-arranged a few things to isolate your code slightly better. Just so it's all in one place, if someone needs to make changes.

For the test, I just removed the pytest-mark thing in `conftest.py` and mocked `Path.unlink()` instead to achieve the same thing (test if profile is imported and file deleted).

Then for the hardcoded `~/.vorta-init.json` path, I moved it to `vorta.config`, since we already have some other path configs there. Just to have it all in one place.

Hope you agree with my changes. It's really just how I like to structure things for the future. There was nothing wrong before. Maybe I even broke something, so please check that it all works as expected still.

After that, let's merge this and do a release to get some feedback (Users don't usually test the `master` branch, so I do a release and then a bugfix release shortly after, if needed 😄)

phihos:
The changes are fine to me. And it totally makes sense that you keep an eye on the overall architecture. I wouldn't do it differently with my own projects :-)

I checked out your changes and tested the bootstrapping. Still seems to work. On bootstrap the user is not notified about the import any more. Is that on purpose?

m3nu:
> On bootstrap the user is not notified about the import any more. Is that on purpose?

Not really. Maybe I cleaned up too much. Let me put it back.

m3nu:
Put back the notification and added a log as well.

m3nu:
Seems we're done here. Next up are some docs on how to use the bootstrapped import. Not very complicated though.

m3nu (approved):

phihos:
That is great news! Can't wait for your next release :tada:

m3nu:
Docs: https://github.com/borgbase/vorta.borgbase.com/pull/24

Please let me know if you want to add anything. This will be below *Usage*

m3nu:
Release tomorrow morning or so. We have enough features by now.

m3nu:
Release done, @phihos : https://github.com/borgbase/vorta/releases/tag/v0.7.6

Stokel:
Just discovered vorta and I noticed, that when exporting my profile, that only the currently selected repository is exported into the json file. Is this working as intended? Is there a way to export the complete profile/configuration to include all repositories?

real-yfprojects:
> Just discovered vorta and I noticed, that when exporting my profile, that only the currently selected repository is exported into the json file. Is this working as intended? Is there a way to export the complete profile/configuration to include all repositories?

Currently only the current profile is exported along with the general settings. If you feel like vorta should show a different behaviour, which is very legitimate, please open a new issue.

Issue #915, opened 2021-03-04
empty files don't show size in extract dialog, and are sorted incorectly.


rblenis:
1. create an archive with some empty files (size==0)
2. select archive and click 'extract'
3. expand tree view to see files
4. notice that empty files are sorted at the top (with directories), and size field is blank

- OS: Ubuntu 20.10
- Vorta version: 0.7.5
- Installed from: snap

m3nu:
Interesting. Will try to reproduce this.

m3nu:
Was able to reproduce this as well. Empty files end up on top of folders.

I believe they should show up with 0B together with other files.

![Screen Shot 2021-04-09 at 11 38 53 PM](https://user-images.githubusercontent.com/3916435/114205410-dac8dc80-998c-11eb-9b5c-965f8db6c33c.jpg)

PR #999, opened 2021-05-28
Fix for #915 . Folder ares sorted with type.

bastiencyr:
Fix for #915 . Folders are sorted with their type ('d') and not their size. It's a little first contribution (but I have deeply studied tree_view and extrac_dialog), so maybe it's not the better way to handle this issue. I don't understand why tests on archive tab are not passed from now on...


m3nu:
Thanks a lot. Every contribution is welcome and even small changes can make a big difference. 👍

m3nu:
Merged. I wasn't aware of the `type` field in the JSON output. It's already there in 1.1.0, which is the lowest version we can support. So no problem to rely on that.

m3nu (approved):

Issue #942, opened 2021-04-08
Back up to multiple repos from one profile


m3nu:
Then let's start out with multiple repos in one profile. For implementation, the repo-dropdown could have [checkboxes](https://stackoverflow.com/questions/22775095/pyqt-how-to-set-combobox-items-be-checkable), like this:

![llcc0](https://user-images.githubusercontent.com/3916435/50468300-0739d400-09e2-11e9-81bd-4f4fbc0f6214.png)

_Originally posted by @m3nu in https://github.com/borgbase/vorta/discussions/853#discussioncomment-369636_

bastiencyr:
Hello @m3nu , do you need help for this ? I started implement this option

m3nu:
Any help is welcome to get improvements launched quicker and eventually add more active maintainers.

diivi:
Should this start with supporting only archive creation with multiple repositories, and then expand to handling other commands with multiple repositories (might need to discuss designs in that case)?
here's how Borgmatic works with rlist:
![image](https://user-images.githubusercontent.com/41837037/226944886-563224a0-2661-4ebf-9990-0a97e27233bd.png)

Vorta could handle this by adding a column to the archive table with the repo name. Or it could use multiple repositories for creating archives and a single repository for all other stuff (select repo in the archive tab to list its archive etc.)

If we end up using multiple repositories for every command Vorta runs, maybe this could also be a good long GSoC task?

m3nu:
They just run the same backup command (same source folders) against different repos. In our case it would add multiple BorgCreate jobs. We already have the internal job queue you added some time ago. Would it be enough to add multiple jobs there? May need small changes to specify the repo.

real-yfprojects:
> internal job queue you added some time ago

Actually @bastiencyr implemented it.

> Would it be enough to add multiple jobs there? 

That should work.

bastiencyr:
This feature is more difficult to tackle than it looks. 
You can see my old PR about this feature at [https://github.com/borgbase/vorta/pull/1005](https://github.com/borgbase/vorta/pull/1005). 

At this time, I didn't have the job queue and I realized that is was nearly impossible to implement this feature without it. 

But there is still some issues and considerations to have : 
 1. You need to consider the export feature and particularly the import feature. Check the conversation in the PR. 
 2. Custom command for backup are added at the profile level for the moment. You need to choose if the command will be made at the profile or repo level 
 3. For the archive, you  also need to add a combo box to select the repo
[...]

PR #1003, opened 2021-05-31
Back up to multiple repos from one profile #942

bastiencyr:
Tests are running well on my laptop... Tests don't pass since the last commit.

I added check box but did you think about archives tab ? How to select a repository to see archives ? 
Maybe something like that : 
QCombobox Archives

The unlink button remove a repository for ALL profiles. Maybe its better to print a message to the user now.

### DIFF
--- a/package/vorta.spec
+++ b/package/vorta.spec
@@ -80,7 +80,7 @@ app = BUNDLE(coll,
                  'SUFeedURL': APP_APPCAST_URL,
                  'LSEnvironment': {
                              'LC_CTYPE': 'en_US.UTF-8',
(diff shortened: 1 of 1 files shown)


bastiencyr:
I have made substantial commits As you notice, all tests have failed since the last commit... but it works well on my laptop and error on github workflow is not clear at all.

I cant change currentIndexChange signal for activated in repo_tab. Tests doesnt pass and I dont know why.

bastiencyr:
Hello @m3nu ,  I think I have made good progress and its stable. What do you think about ? 

(For the moment, it is still missing activated signal and a solution for archive. Multiple backups doesnt work with scheduler.)

m3nu:
Thanks! Will check it out tomorrow.

I didn't consider the issue of viewing the archive list. 😬

PR #1004, opened 2021-06-01
Use pipe to pass password to borg

steelman:
Depending on system configuration environment variables of a process
may be readable for other processess. Password passed through a pipe
is not visible to other processes.


m3nu:
Which problem will this solve? We actually want to pass this password to Borg only.

It also adds a good bit of complexity to the code. So there would need to be a strong reason to make this change.

steelman:
BTW. I have tested this patch only on Linux. I expect it to work properly on macos, but I've got no proof. Please make sure it works fine before merging.

m3nu:
I see the issue, appreciate the contribution and understand the security concern, even if it's minor. Concrete reservations I have:

- Add a good chunk of new code that new contributors will need to understand before making any changes. The current mechanism is easier to understand.
- Support for `BORG_PASSPHRASE_FD` was only added in Borg 1.1.8. We generally support 1.1.0, so this will also need a version check for older Borg versions and both methods implemented.
- Will this FD trick work on macOS and Windows reliably?
- All [tests](https://github.com/borgbase/vorta/pull/1004/checks?check_run_id=2720088691) are failing after this change. `AttributeError: 'BorgListRepoThread' object has no attribute 'pipe_read'`

steelman:
> Adds a good chunk of new code that new contributors will need to understand before making any changes. The current mechanism is easier to understand.

That is a matter of judgement and of course it is up to you to decide.

> Support for BORG_PASSPHRASE_FD was only added in Borg 1.1.8. We generally support >=1.1.0, so this will also need a version check for older Borg versions and both methods implemented.

Fair point. I'll see, what I can do.

> Will this FD trick work on macOS and Windows reliably?

As I wrote above, I expect it to work on macOS. I can check Windows. I'll write some short testing code and put it here.

> All tests are failing

Fixed, pushed.

bastiencyr:
I have made substantial commits As you notice, all tests have failed since the last commit... but it works well on my laptop and error on github workflow is not clear at all. 

I cant change currentIndexChange signal for activated in repo_tab. Tests doesnt pass and I dont know why.

real-yfprojects:
Are you still working on this @steelman?

ThomasWaldmann:
I agree that this needs thorough platform testing.

About minimum borg requirement being borg 1.1.8:

There was a **major** bugfix in 1.1.11 that fixed a hashtable issue, see also the advisory at top of the change log.

Thus, nobody should use anything older than 1.1.11 nowadays (only exception being patched/fixed versions like e.g. 1.1.9 in debian or ubuntu).

steelman:
I'll revisit it.

Issue #1016, opened 2021-06-15
Diff doesn't show the modified status or size of change for files anymore


Victor239:
**Describe the bug**
When I did my first two backups with Vorta doing a diff between them seemed to work fine - every file was either green, red or yellow and showed the size of the change. I've done a third one, but now that I try to diff between any two backups it only shows me the modified status and size change for one file despite there being 3.7GB of changes:

![](https://i.imgur.com/NDgPgr7.png)

This is using borgbase.com as a remote repo.

**To Reproduce**
1. Create 3 backups for a single repo to borgbase.com
2. Try to diff any of the 2 backups
3. See error

**Environment (please complete the following information):**
- OS: Kubuntu 21.04
- Vorta version: 0.7.6
- Installed from: Flatpak

**Additional context**
Let me know if there's anything else needed to troubleshoot this.

m3nu:
I can confirm that the output of this dialog changed between v0.7.5 and v0.7.6. Probably related to https://github.com/borgbase/vorta/pull/909. Maybe @rblenis can weigh in?

Hofer-Julian:
Can confirm, the details of the change are not present any more.

bastiencyr:
Im sorry, its clearly related to #999 :( I'm working on fix, I will open a PR soon.
Notice that making diff between two repo without any differences will result in bad failure : 
`
Uncaught exception, please file a report with this text at
https://github.com/borgbase/vorta/issues/new
Traceback (most recent call last):
  File "/home/bastien/Documents/Travail/Vorta/vorta/src/vorta/views/archive_tab.py", line 558, in list_diff_result
    window = DiffResult(result['data'], archive_newer, archive_older, result['params']['json_lines'])
  File "/home/bastien/Documents/Travail/Vorta/vorta/src/vorta/views/diff_result.py", line 35, in __init__
    files_with_attributes = [attrs + (type_f, ) for attrs, type_f in zip(files_with_attributes, fs_data[1])]
IndexError: string index out of range

`

m3nu:
Oh, great. Waiting for your PR.

rblenis:
sorry, been tied up at work. glad someone is taking a look. I'm available to review or help out if needed.

Issue #1011, opened 2021-06-08
Can't run v0.7.6 with latest PyQt5 version


JackDinn:
I just had a vorta update 
[2021-06-08T17:20:16+0100] [ALPM] upgraded vorta (0.7.5-1 -> 0.7.6-1)

Now it won't run and i can't figure out exactly what iv got to do?

```
❰greg❙~❱✔≻ vorta
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/pkg_resources/__init__.py", line 568, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3.9/site-packages/pkg_resources/__init__.py", line 886, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3.9/site-packages/pkg_resources/__init__.py", line 777, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (PyQt5 5.15.4 (/usr/lib/python3.9/site-packages), Requirement.parse('pyqt5<=5.15.3'), {'vorta'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/vorta", line 33, in <module>
    sys.exit(load_entry_point('vorta==0.7.5', 'gui_scripts', 'vorta')())
  File "/usr/bin/vorta", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.9/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 855, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/usr/lib/python3.9/site-packages/vorta/__main__.py", line 10, in <module>
    from vorta.models import init_db
  File "/usr/lib/python3.9/site-packages/vorta/models.py", line 17, in <module>
    from vorta.utils import slugify
[...]

PR #1019, opened 2021-06-18
Fix for  #1016

bastiencyr:
Hello, 
Folders and files are correctly sorted without error and regression :)


m3nu:
Thanks for the quick fix!

Hofer-Julian:
Can we make a bugfix release for that? 
Alternatively I can just release the latest master on flathub.

m3nu:
Yes, I was planning that. I also want to undo the PyQt version limit from #1011 , since it only affects Pyinstaller and isn't needed for other install options.

Hofer-Julian:
Thanks!

PR #1028, opened 2021-07-10
Fix context menu not accessible on some setup

olivierlemoal:
On some setup (i3 wm), there is a bug preventing user to access the context menu from the tray icon. The context menu is hidden from user after a right click : 

https://user-images.githubusercontent.com/1941583/125169047-c1f8de80-e1a8-11eb-905e-03cfc770b540.mp4

The `QMenu::clear()` method seems to be responsible for this behavior. Calling `clear()` on the menu widget only when menu is hidden fix the issue.

### DIFF
--- a/src/vorta/tray_menu.py
+++ b/src/vorta/tray_menu.py
@@ -16,6 +16,7 @@ class TrayMenu(QSystemTrayIcon):
 
         # Workaround to get `activated` signal on Unity: https://stackoverflow.com/a/43683895/3983708
         menu.aboutToShow.connect(self.on_user_click)
+        menu.aboutToHide.connect(self.on_menu_hide)
 
         self.setContextMenu(menu)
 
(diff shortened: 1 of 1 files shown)


bastiencyr:
Hey, 
It seems that clear doesnt work anymore. Do you have the same bug ? 

https://user-images.githubusercontent.com/63359575/125177359-64ca5080-e1db-11eb-82d9-d7ca1d2c76f3.mp4

olivierlemoal:
Not really, I my case the actions items are indeed cleaned up and does not stack up like this. I just have this issue of the context menu that won't stay displayed.

olivierlemoal:
What is your desktop environment ?

bastiencyr:
> What is your desktop environment ?

Yes I checked the link in the code too : https://stackoverflow.com/a/43683895/3983708 and https://bugreports.qt.io/browse/QTBUG-55911
The link suggest that its an issue with unity. But I dont have unity, I have linux mint (so its cinnamon = fork of gnome 2). I think  we can add Mint to the list of OS where aboutToHide is not emitted.

olivierlemoal:
Indeed, that's a shame :octopus: 
I'll think of something else...

PR #1031, opened 2021-07-12
Fix context menu not accessible on some setup

### LINKED ISSUE #1030: Context menu not accessible on some setup
**Describe the bug**
On some setup (i3 wm), there is a bug preventing user to access the context menu from the tray icon. The context menu is hidden from user after a right click :

https://user-images.githubusercontent.com/1941583/125209925-b090fe80-e29c-11eb-9074-cb4e9a777c18.mp4

I tried to push some fix (#1028), but it seems that the `aboutToHide` signal is not reliable on cross-platform environments.

I found that the `on_user_click` function is run twice when right clicking the systray on my setup : 
* Via the `aboutToShow` signal: https://github.com/borgbase/vorta/blob/master/src/vorta/tray_menu.py#L18
* Via the `on_activation` function: https://github.com/borgbase/vorta/blob/master/src/vorta/tray_menu.py#L37

Only the second call (in the `on_activation` function), seems to cause this behavior. It seems that the `aboutToShow` is reliable on cross-platform, is there any benefits in keeping the second call ?

**Environment:**
- OS: Linux 5.12.15-arch1-1
- Vorta version: 0.7.7
- Installed from: Arch PKGBUILD

olivierlemoal:
This PR fix #1030.

It was tested on : 
- i3 (Arch)
- sway / Wayland (Arch)
- Cinnamon (Mint)
- Unity (Ubuntu 20.04)
- Windows 10

The `aboutToShow` signal should be reliable on Mac as well (not tested).

### DIFF
--- a/src/vorta/tray_menu.py
+++ b/src/vorta/tray_menu.py
@@ -33,8 +33,6 @@ class TrayMenu(QSystemTrayIcon):
         if reason in [QSystemTrayIcon.Trigger, QSystemTrayIcon.DoubleClick] and \
                 os.environ.get('XDG_CURRENT_DESKTOP'):
             self.app.toggle_main_window_visibility()
-        else:
-            self.on_user_click()
 
     def on_user_click(self):
         """Build system tray menu based on current state."""


m3nu:
Still works on macOS. So guess we can merge it, since you tested on Linux already.

PR #1008, opened 2021-06-05
UI Improvements

bastiencyr:
Some variables have to be rename (qtoolbox) but it's ok. Do you like ? 

(edit) Related discussion: https://github.com/borgbase/vorta/discussions/993

PR #1045, opened 2021-07-27
Implement multiple queues.

bastiencyr:
Hello, 
This PR will help me for #1005 . It is not ready yet since cancel button doesn't work anymore. Scheduler now has the ability to run jobs on multiple repositories concurrently and run multiple jobs on one repo (by queuing them). 

In the future, all jobs will be copied into the db and run later if needed. It does not solve missed backup. It solves a fail backup (Power failure for example). I wait #1005 for this because #1005 change the db. 

For each repository, there is one queue. I have represented a queue by a 'site'. Between sites (ie repository), tasks run concurrently. On one site, tasks run one by one. 
The user also run tasks by adding them to the queue but he can't run multiple backups because start backup button is disabled when a job is running.  

Related :  #830 

- [x]  support for cancel jobs 
- [x] check concurrent access (add mutex where necessary)
- [x] clean code 
- [x] docs
- [x] pass tests 
- [x] use simple queue instead of priority queue 
- [x] rewrite enabled/disabled cancel button logic maybe with semaphore


m3nu:
Thanks for taking care! So this should be merged before or after #1005?

Since this is an essential feature, we should discuss it in detail before writing code. But testing different options is often needed of course.

My travels are almost done and I hope to get more involved again in the next days.

bastiencyr:
> Thanks for taking care! So this should be merged before or after #1005?

Before its better.

> Since this is an essential feature, we should discuss it in detail before writing code. But testing different options is often needed of course.

I have no problem rewriting this code :)

> My travels are almost done and I hope to get more involved again in the next days.

Enjoy your travel, this gives me time to better understand the code ;)

bastiencyr:
I added cancel support. There are some tests that don't pass anymore, it will be fix soon.

This is what the VortaQueue looks now : 
image

bastiencyr:
Hello @m3nu , this PR is almost done and I dont expect any major implementation changes. The only work left is to use VortaQueue everywhere in the code (mount, prune, rename...)

m3nu:
Looks good at first sight. I'll run it in different situations tomorrow and get back.

Thanks for the contribution!

m3nu:
When a background job runs during a user-initiated backups, the "Start" button doesn't go back to enabled state. Probably the reason why one test is failing on the same check.

Screen Shot 2021-09-20 at 08 50 46

m3nu:
> This new implementation makes BorgThread somewhat useless. The JobsManager already launch thread, so its a bit repetitive to launch a thread.

You can remove the inheritance from QThread for `BorgThread` and call it e.g. `BorgJob`, if threading is now done elsewhere. The main purpose of `BorgThread` is to prepare, run and post-process different Borg commands.

So the user can only start one job at a time, but if there are multiple background jobs, they won't fail, but run after?

src/vorta/scheduler.py
```diff
@@ -96,21 +114,139 @@ def next_job_for_profile(self, profile_id):
...
+    def set_status(self, status):
+        self.__status = status
+
+    # Keep default
+    def get_status(self):
+        return self.__status
+
+    # Must return the site id. In borg case, it is the id of the repository.
+    @abstractmethod
+    def get_site_id(self):
```
m3nu, line 147: How about calling it repo_id to avoid confusion? Since we don't use the concept of a `site` anywhere else.

src/vorta/scheduler.py
```diff
@@ -96,21 +114,139 @@ def next_job_for_profile(self, profile_id):
         else:
             return job.next_run_time.strftime('%Y-%m-%d %H:%M')
 
-    def create_backup(self, profile_id):
+    def enqueue_create_backup(self, profile_id, repo_id, vorta_queue):
+        vorta_queue.add_job(CreateJobSched(profile_id=profile_id, site=repo_id))
+
+
+class JobStatus(Enum):
```
m3nu, line 121: I wonder if all your job scheduling classes would better fit in the `borg` package, since they deal with running Borg commands? The `scheduler.py` module is just one more way to add jobs, but not much different from pressing the *Start* button. Maybe add your classes in `borg/jobs_manager.py`?

src/vorta/views/archive_tab.py
```diff
@@ -216,42 +221,55 @@ def check_action(self):
...
             self._set_status(self.tr('Pruning finished.'))
-            self.list_action()
+            self.job_list_action()
         else:
             self._toggle_all_buttons(True)
 
+    def job_list_action(self):
+        profile = self.profile()
+        QApplication.instance().scheduler.jobs_manager.add_job(
+            FuncJob(self.list_action, site=profile.repo.id))
```
m3nu, line 262: So this is the new way to run a Borg command? I'm surprised this still works. Isn't the function running in another context or is it the same in another Thread? Where did you learn or see this pattern? Looks really useful for GUI apps.

m3nu: What we really want is async/await with QThreads, right? This lib here is kinda similar: https://github.com/CabbageDevelopment/qasync/blob/master/qasync/__init__.py#L166

bastiencyr:
> > This new implementation makes BorgThread somewhat useless. The JobsManager already launch thread, so its a bit repetitive to launch a thread.
> 
> You can remove the inheritance from QThread for `BorgThread` and call it e.g. `BorgJob`, if threading is now done elsewhere. The main purpose of `BorgThread` is to prepare, run and post-process different Borg commands.

Fine I will try to the thread dependence. 

> So the user can only start one job at a time, but if there are multiple background jobs, they won't fail, but run after?

Yes The user have to wait that every jobs are finished to run a job and can run one job at a time because the manager tracks running job. The scheduler do whatever it wants, so if on 2 profiles you run backups on repo 1 and repo 2 at same time, the backup will be parallelized. If at the same time, a third profile run a backup on repo 1, the manager will just add this job to a queue to be run later.  
You can make some performance tests, If you schedule backups on different profile on different backups, it will be faster than running these backups one by one (from the user for example).

src/vorta/scheduler.py
bastiencyr: Yes it's true we can. get_site_id could return other thing that repo id. For example, if in the code I want to run borg version, we can use the manager. Since there is no repo associated with a borg version, get_site_id will just return "borgversion" as repo.

src/vorta/scheduler.py
bastiencyr: I can because there are too much code in scheduler.py and no link between implementation. 
But maybe I can let the jobsManager initialization in the scheduler. In some way, the jobsManager schedule jobs too.

src/vorta/views/archive_tab.py
bastiencyr: I'm not sure to understand. 
It's the same function run in another thread. FuncJob is a convenient way to run a func in  a job. Usually, you have to inherits Job and override run method. Job is defined in scheduler.py 
Its inspired from the command pattern which is indeed useful for gui.

bastiencyr:
> When a background job runs during a user-initiated backups, the "Start" button doesn't go back to enabled state. Probably the reason why one test is failing on the same check.
> 
> Screen Shot 2021-09-20 at 08 50 46

Is it happens with my last commit ? If so, you launch a backup on a different repo than the scheduler ?

m3nu:
> Is it happens with my last commit ? If so, you launch a backup on a different repo than the scheduler ?

Yes, the latest commit https://github.com/borgbase/vorta/pull/1045/commits/2bff7e82b8830c46a08a79706793491d17576d6a

Happened, when I had a backup from one profile running and another profile was scheduled, which ran after (or at the same time).

bastiencyr:
> When a background job runs during a user-initiated backups, the "Start" button doesn't go back to enabled state. Probably the reason why one test is failing on the same check.
> 
> Screen Shot 2021-09-20 at 08 50 46

bastiencyr:
> > Is it happens with my last commit ? If so, you launch a backup on a different repo than the scheduler ?
> 
> Yes, the latest commit [2bff7e8](https://github.com/borgbase/vorta/commit/2bff7e82b8830c46a08a79706793491d17576d6a)
> 
> Happened, when I had a backup from one profile running and another profile was scheduled, which ran after (or at the same time).

Sorry I missclicked and closed the PR.

I cannot reproduce this. What I tried :
profile1:  repo1 , 4GB to backup
profile2 : repo2 , 8GB to backup

1. Schedule profile2 on repo2 for 19:10. Start profile1 on repo1 at 19:09. Works well. 
2. Remove archives. Schedule profile1 on repo1 for 19:10. Start profile2 on repo2 at 19:09. Works well.

src/vorta/views/archive_tab.py
m3nu: True. Async/await wouldn't have a queue. It just runs the function without blocking the main thread (or UI).

src/vorta/scheduler.py
m3nu: Yes, that would be great! I know the name and concept of `scheduler.py` is similar, but what you did for the execution of Borg threads is more fundamental.

I also plan on making some changes to the actual Scheduler later. So the code there needs to focus on scheduling the next job, scheduling missed jobs, etc: https://github.com/borgbase/vorta/pull/908

src/vorta/scheduler.py
bastiencyr: Done

src/vorta/scheduler.py
bastiencyr: Done

m3nu:
Activity display in the taskbar also stops working with this PR. Should be orange, when a backup is active.

Screen Shot 2021-09-22 at 18 23 21

bastiencyr:
> Activity display in the taskbar also stops working with this PR. Should be orange, when a backup is active.
> 
> Screen Shot 2021-09-22 at 18 23 21

I dont have this bug too :( The method set_tray_icon in https://github.com/borgbase/vorta/blob/901992f026e2ddc968ec65430c2910fc6d657452/src/vorta/tray_menu.py is called with active=true on my laptop when I start  a backup.

src/vorta/borg/borg_thread.py
```diff
@@ -46,15 +55,15 @@ def __init__(self, cmd, params, parent=None):
...
+        self.site_id = site
         self.app = QApplication.instance()
-        self.app.backup_cancelled_event.connect(self.cancel)
 
         # Declare labels here for translation
-        self.category_label = {"files": trans_late("BorgThread", "Files"),
-                               "original": trans_late("BorgThread", "Original"),
-                               "deduplicated": trans_late("BorgThread", "Deduplicated"),
-                               "compressed": trans_late("BorgThread", "Compressed"), }
+        self.category_label = {"files": trans_late("BorgJob", "Files"),
```
m3nu, line 63: Note to self: Update translations after renaming this.

src/vorta/borg/borg_thread.py
```diff
@@ -13,30 +13,39 @@
...
+All methods in this class must be thread safe. Particularly,
+I strongly unadvised global variable and class variables.
+Sqlite access are thread-safe because peewee is thread-safe.
+The method prepare is not thread-safe because of keyring and I don't know why. That's why I added a
+temporary mutex.
+"""
+
 
-class BorgThread(QtCore.QThread, BackupProfileMixin):
+class BorgJob(Job, BackupProfileMixin):
```
m3nu, line 38: The file should be called `borg_job.py` now?

src/vorta/scheduler.py
```diff
@@ -30,6 +32,11 @@ def __init__(self, parent):
         self.qt_timer.setInterval(45 * 60 * 1000)
         self.qt_timer.start()
 
+    def cancel_all_jobs(self):
```
m3nu, line 35: Depending on where this is used, the caller could just use `self.jobs_manager.cancel_all_jobs()` directly, since it's not really related to the scheduler.

m3nu:
Good progress. I'm liking this more and more.

There is still this random segfault in the test. `FAILED tests/test_repo.py::test_repo_add_success`. Probably an orphaned Qt object or something. It's a pain to get this right for the tests. I can look at that.

src/vorta/borg/borg_thread.py
bastiencyr: It makes sense. Its true that BorgJob class inherits from Job but Job is just an object, not a thread. You have to add it to the JobsManager to be run.

src/vorta/scheduler.py
bastiencyr: I am not sure about this. This method simply cancels all jobs without knowing the scheduler in depth. Calling cancel_all_jobs directly, some method will call "this_instance.scheduler.jobs_manager.cancel_all_jobs()" which is long. If we do that, maybe the JobsManager can be initialized in the VortaApp and the previous call becomes "this_instance.jobs_manager.cancel_all_jobs()".

src/vorta/borg/borg_thread.py
bastiencyr: Done locally.

bastiencyr:
> Good progress. I'm liking this more and more.
> 

<3

> There is still this random segfault in the test. `FAILED tests/test_repo.py::test_repo_add_success`. Probably an orphaned Qt object or something. It's a pain to get this right for the tests. I can look at that.

I forget to add a job to the jobs_manager in repo_add_success and the job.run() method was directly run. But Im not sure to understand why it solves this issue. If I job.run() is called, there is no thread started like before. So the program block on this call. 
Maybe, its linked to this line I removed : "self.thread = thread". 
The commit that solved this is a78a615 , "Probably pass tests"

m3nu:
Nice! Seems to have fixed it.

I'll play with it a bit more and possibly fix minor bugs, like the icon that we saw.

Then we make a new branch v0.8.0, since this is a bigger change and we need to keep a maintenance branch of v0.7

bastiencyr:
> Nice! Seems to have fixed it.
> 
> I'll play with it a bit more and possibly fix minor bugs, like the icon that we saw.
> 
> Then we make a new branch v0.8.0, since this is a bigger change and we need to keep a maintenance branch of v0.7

Yes ! After this I can back to #1005, the rebase will be a pain hh. 
Will #1005 and #1008 also be in v0.8.0 ?

m3nu:
> Will #1005 and #1008 also be in v0.8.0 ?

We should keep all large changes from here in a new branch. Mostly because Vorta is in Debian and they need to cherry-pick minor patches sometimes, but can't have too many changes. It also gives us time to properly develop a new major release without thinking about breaking stuff (as much).

m3nu (approved):

m3nu:
Thanks for this massive contribution! 👏 And sorry it took longer to merge it. I'm just wary of large PRs. I see you simplified a lot fo things and addressed all comments.

I've added a maintenance branch for the previous version `v0.7` for adding small and urgent fixes. So we can keep doing larger changes in `master`. I'll also update the issue milestones to reflect this.

bastiencyr:
> Thanks for this massive contribution! clap And sorry it took longer to merge it. I'm just wary of large PRs. I see you simplified a > lot fo things and addressed all comments.

You're welcome. 
No problem, it was good to take time. But there will be bugs and I will be there to fix it.

Issue #1050, opened 2021-08-02
Follow an official design guideline HIG


bastiencyr:
Hello, 

I started to read some guidelines and docs from official design guidelines 💪 . According to me,  vorta UI lacked consistency and didn't look modern but I couldn't say why. So after some research, I discovered Flat and Material design. Flat design is everywhere but I didn't realize and because I have always seen this theme (I'm rather young 😅 ), vorta UI interface seems really old to me.

Consistency is another notion and design guidelines are very helpful for that. If we want to make an interface that looks native, user should  find the elements in the same place and in the same way. For example, vorta doesnt follow everytime the header capitalization ( [Gnome](https://developer.gnome.org/hig/stable/guidelines/writing-style.html) , [Apple](https://developer.apple.com/design/human-interface-guidelines/macos/fields-and-labels/text-fields/) ). "One folder or file per line" should be "One Folder or File per Line" or "Can't delete the last profile." should be "Can't Delete the Last Profile". 

Since vorta users come from Apple, Gnome or KDE, it's more complicated. But lets be honest, all UI interfaces and guidelines converge towards the same ideas 😝 and looks the same !

So, my proposition is to follow KDE, Gnome and/or Apple guidelines as guide and to follow flat design. Note that Apple, Gnome and KDE have officially chosen the flat design. 

This is the three guidelines ❤️ : 
https://develop.kde.org/hig/introduction/
https://developer.gnome.org/hig/stable/principles.html
https://developer.apple.com/design/human-interface-guidelines/macos/app-architecture/onboarding/

There are many common points between these guidelines.

Thanks for reading 🥰

m3nu:
Do you have some concrete examples that could be changed to follow such guidelines?

bastiencyr:
> Do you have some concrete examples that could be changed to follow such guidelines?

I don't read the whole guidelines but this is some changes : 
For text : 

 - header capitalization as I said
 - limit the tooltip because touch screen can't print the tooltip
 -  keep text short (translation can expand the text to 30% more in some languges. For example translation of "Repositories and SSH keys are shared between profiles. Source folders, active destination repo, allowed networks, pruning, validation and scheduling are per-profile. (166 chars)" in french is "Les dépôt et les clés SSH sont partagés entre les profils. Les dossiers sources, le dépôt de destination actif, les réseaux autorisés, l'élagage, la validation et la planification sont propres à chaque profil. (210 char)". 
 - text must be selectable 
[...]

PR #1072, opened 2021-10-06
Add tooltip containing path name

XXXBold:
Minor improvement, avoiding situations which might lead to path names not readable because they're too long (Or vorta-window too small):
added-tooltip

### DIFF
--- a/src/vorta/views/source_tab.py
+++ b/src/vorta/views/source_tab.py
@@ -112,8 +112,10 @@ class SourceTab(SourceBase, SourceUI, BackupProfileMixin):
 
         index_row = self.sourceFilesWidget.rowCount()
         self.sourceFilesWidget.insertRow(index_row)
-        # Insert all items on current row
-        self.sourceFilesWidget.setItem(index_row, SourceColumn.Path, QTableWidgetItem(source.dir))
+        # Insert all items on current row, add tooltip containg the path name
+        new_item = QTableWidgetItem(source.dir)
+        new_item.setToolTip(source.dir)
+        self.sourceFilesWidget.setItem(index_row, SourceColumn.Path, new_item)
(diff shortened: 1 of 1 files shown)


bastiencyr:
Hello @XXXBold , 
Thanks for the contribution <3
Tooltips should be avoid if possible (tooltip not printed in touchscreen, see #1050). However I don't see other simple solution and this tooltip doesn't give essential info.  Another solution is to scroll the text when item is selected. 
So its ok for me. Lets wait the maintainer 😁

XXXBold:
Hi @bastiencyr 
Thanks for the feedback.
It might be true that tooltips are not helpful on a touchscreen, but I implemented this behaviour because it's common practice in  most software I know.
I don't think there are much people out there using vorta on a touch screen either, because most parts of the UI aren't suitable for that anyway.

bastiencyr:
> Hi @bastiencyr Thanks for the feedback. It might be true that tooltips are not helpful on a touchscreen, but I implemented this behaviour because it's common practice in most software I know. I don't think there are much people out there using vorta on a touch screen either, because most parts of the UI aren't suitable for that anyway.

What is important is to not abuse them and not give essential information and it's fine for this tooltip 🤗. I just quoted the HIG of Gnome.

Most part ?
Even if Vorta is not suitable for touchscreen, It's not a reason to continue. It's just that nobody has considered this before.

XXXBold:
> Even if Vorta is not suitable for touchscreen, It's not a reason to continue.
Sure, don't get me wrong, you may ofc use it this way. I'm just pretty sure that the current interest in touch screens is low (aside from mobile devices like smart phones).

> It's just that nobody has considered this before
Well, I think this might gain attention if for example Android or other mobile systems get support from borgbackup :)

PR #1093, opened 2021-11-01
Always handle autosave setting when saving SettingsModel (fixes #1092)

### LINKED ISSUE #1092: Autostart setting not set after importing settings from .vorta-init.json
**Describe the bug**

After importing a profile from .vorta-init.json the autostart setting is not set in the system. 
This makes it impossible to provision Vorta with autostart without user interaction. 

**To Reproduce**
1. Populate a .vorta-init.json with autostart setting enabled.
2. Start Vorta.
3. See that the setting is enabled, but it is not actually in autostart.
4. As workaround disable and reenable the setting.

- OS: Linux
- Vorta version: 0.7.8
- Installed from: PyPI

A PR will follow soon.

phihos:
Fixes #1092 - This PR attempts to fix the autostart setting not being set correctly when importing. I deliberately chose not to duplicate the code from the misc tab to the init_db function. Instead I tried to keep it DRY and use signals from Pewee. This has the advantage that is does not matter where the SettingsModel is saved - the autostart is always set up afterwards.
This is only a suggestion how to fix it and I know that this is a bit different from an architectural point of view. We basically pull logic from a view to the models. I am always open to a different approach.

### DIFF
--- a/src/vorta/models.py
+++ b/src/vorta/models.py
@@ -10,8 +10,10 @@ import sys
 from datetime import datetime, timedelta
 
 import peewee as pw
+from playhouse import signals
 from playhouse.migrate import SqliteMigrator, migrate
 
+from vorta.autostart import open_app_at_startup
 from vorta.i18n import trans_late
 from vorta.utils import slugify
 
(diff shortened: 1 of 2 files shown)


m3nu (approved):

m3nu:
Looks fine. Maybe we will find other uses for Peewee Signals.

https://docs.peewee-orm.com/en/latest/peewee/playhouse.html#connecting-handlers

PR #1105, opened 2021-11-12
Remove jobs if scheduler setting is changed.

m3nu:
This fixes a situation where the schedule for a profile is disabled, but there are still jobs scheduled. Those need to be removed.


m3nu:
This fixes another (common) scheduler case. What do you think, @jayeclark? Did you observe this other issue too? Does this PR resolve the issue and can be merged?

Issue #1026, opened 2021-06-29
Feature: Ability to sort by size when diffing


Victor239:
Current when using Diff you can only see the file changes sorted by directories, it would be helpful to also be able to instead sort by size so that the user can easily see which particular directories or files contributed the most megabytes in a particular diff. 

It'd be helpful especially when I for example see an unexpected large diff of multiple gigabytes, but I can't easily tell which directories are contributing all of that data because many directories each contribute hundreds of little small changed files.

Golddouble:
I have already suggested this feature here:
https://github.com/borgbase/vorta/discussions/973

> a) you could add the possibility to sort the list according to file-size

Maybe you want to vote for this there.

m3nu:
Probably easy to add. Others have added sorting to Qt tables before. Just needs some settings added when setting up the Qt table. Most likely here: https://github.com/borgbase/vorta/blob/master/src/vorta/views/diff_result.py#L40

real-yfprojects:
This could be done similiar to the KDE Taskmanager *KSysGuard*:

![Screenshot_20220106_194910](https://user-images.githubusercontent.com/62463991/148435254-c32fd5df-706d-40b8-9668-29fe017933bf.png)

It supports a tree view and a list/table view in which the items can be sorted by every single column.

m3nu:
Sorting would happen by clicking the column title. Same as in the "CPU" column in your screenshot.

Your dropdown does filtering, which is a different operation.

real-yfprojects:
The important to "filters" are `All Processes` and `All Process, Tree`. The first one is a simple list of process (in our case changes) and the second one displays the processes in a tree (like a directory tree in the diff view) and only sorts within an hierarchy.
Also a search bar might be useful in the diff view.

Victor239:
Feature is added now https://github.com/borgbase/vorta/pull/1156

real-yfprojects:
I don't think that one is able to sort by size at the moment.

Victor239:
An additional UX issue is that folders aren't always colour-coded, and are usually missing a value in the `Modified` column. In this example below I had to open two subfolders in order to find out if the given size for `.code42` had been removed or added:

![](https://i.imgur.com/b4cpRRY.png)

real-yfprojects:
I would suggest to show a signed size for indicating that and leave the modified column for the folders as is since they are not modified in this example.

PR #1138, opened 2021-12-15
Adds ability to sort by size when diffing (#1026)

moleskined:
I have added the ability to sort the diff tree, including by size. I've maintained the original behaviour of keeping folders at the top when sorting by name.

### DIFF
--- a/src/vorta/views/diff_result.py
+++ b/src/vorta/views/diff_result.py
@@ -3,9 +3,9 @@ import os
 import re
 
 from PyQt5 import uic
-from PyQt5.QtCore import Qt, QVariant
+from PyQt5.QtCore import QModelIndex, QSortFilterProxyModel, Qt, QVariant
 from PyQt5.QtGui import QColor
-from PyQt5.QtWidgets import QHeaderView
+from PyQt5.QtWidgets import QHeaderView, QTreeView
 
 from vorta.utils import (get_asset, get_dict_from_list, nested_dict, uses_dark_mode)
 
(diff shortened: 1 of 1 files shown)


m3nu:
Thanks for the contribution! Works as advertised.

Please just make sure to fix the linting and [failing tests](https://github.com/borgbase/vorta/runs/4545427971?check_suite_focus=true). And maybe add a relevant test to ensure the feature you added keeps working after future updates.

```
./src/vorta/views/diff_result.py:252:30: E713 test for membership should be 'not in'
./src/vorta/views/diff_result.py:253:31: E713 test for membership should be 'not in'
```

src/vorta/views/diff_result.py
```diff
@@ -219,3 +233,49 @@ def flags(self, index):
...
+            try:
+                left_value = left_data["bytes"]
+            except KeyError:
+                left_value = left_total_bytes
+            try:
+                right_value = right_data["bytes"]
+            except KeyError:
+                right_value = right_total_bytes
+
+        return left_value < right_value
```
real-yfprojects, line 281: This seems way to complicated for implementing a *lower than*. I especially think that the use of the sort order and sort column attribute might even be wrong inside this method. The use of the `UserRole` also seems unnecessary since the internal pointer of the indices will reference the corresponding `FolderItem`. 

This is how simple I did it:
https://github.com/real-yfprojects/vorta/blob/sortproxymodel/src/vorta/views/diff_result.py#L240-L262

PR #1145, opened 2021-12-21
Upgrade flatpak runtime org.kde.Platform to v5.15

### LINKED ISSUE #1023: Flatpak vorta doesn't take the system theme
Hello, 

Flatpak vorta doesn't take the system theme which result in a bad look for cinnamon users.  The theme is the native theme of qt I guess. 
With the ppa, it works well.

- OS: Linux Mint cinnamon
- Vorta version: >=0.7.6
- Installed from:flatpak and ppa 

Thanks :)

phihos:
This MR partially fixes #1023. The theming looks closer to the system theme, but dialogs are still missing window decoration.
Here are some screenshots taken with Ubuntu 21.10:
grafik
grafik
grafik

The verbose flag when compiling PyQt is not strictly necessary but it helps detecting issues like I had when I did not know that the new runtime is based on Python 3.8 instead of 3.7.

### DIFF
--- a/flatpak/com.borgbase.Vorta.json
+++ b/flatpak/com.borgbase.Vorta.json
@@ -1,7 +1,7 @@
 {
     "app-id": "com.borgbase.Vorta",
     "runtime": "org.kde.Platform",
-    "runtime-version": "5.14",
+    "runtime-version": "5.15",
     "sdk": "org.kde.Sdk",
     "command": "vorta",
     "branch": "devel",
--- a/flatpak/dependencies/pyqt5.json
+++ b/flatpak/dependencies/pyqt5.json
@@ -13,7 +13,7 @@
         {
             "type": "script",
             "commands": [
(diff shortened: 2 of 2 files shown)


Hofer-Julian:
Please open PRs towards the flatpak [here](https://github.com/flathub/com.borgbase.Vorta).


Unfortunately this specific one will not be merged since it breaks the directory chooser on Kubuntu 20.04

phihos:
@Hofer-Julian Ah I did not know about that repo. Are the [docs](Upgrade flatpak runtime org.kde.Platform to v5.15) outdated on that matter?
And can you give me more insight about the broken directory chooser? Is this a know bug in the runtime or does Vorta's Qt code need some changing?

m3nu:
Maybe there is a workaround or newer version to get around the chooser bug?

Hofer-Julian:
> Are the docs outdated on that matter?

It is still valid for local development buids (but I should do a better job to keep it up to date)

> Is this a know bug in the runtime or does Vorta's Qt code need some changing?

While researching this, I found other flatpaks based on Qt with the same problem on Kubuntu 20.04.
Vorta's code is correct as far as I can tell

> Maybe there is a workaround or newer version to get around the chooser bug?

That would be nice, but couldn't find one.
Keep in mind that this is not a general KDE problem but only occurs on Kubuntu 20.04.

Hofer-Julian:
I had the hope that we could just update as soon as Kubuntu 22.04 is out.

Issue #1140, opened 2021-12-17
Feature: when diffing 2 archives, show size at the directory level too


sandrotosi:
Hello,
when i diff 2 archive, the size difference is only shown at the file level; that makes it hard to know which directories contain the most changed files/size, which is a valuable information.

Can you extend the diff tree to show the diff size also at the directory level, and not only at the file level?

thanks!

m3nu:
We don't get this data from Borg, so it would need summing up the files while building the tree. Could work because the current data model is relatively well suited to do this. Would just be one more step when adding more items to the parent tree object.

m3nu:
Fixed via #1156

PR #1156, opened 2022-01-03
Diff view now shows folder sizes as sum of sizes of contained items

skrap:
I found that I wanted a diff view which would allow me to investigate unexpectedly large backups.  Since the size column was unused on the current diff view, and it seemed natural to me to have it be the sum of the size of the contained items, I made that change.

String compare is used when filtering because large diffs were taking too long with os.path.commonpath, which is more general than what we need here.

### DIFF
--- a/src/vorta/views/partials/tree_view.py
+++ b/src/vorta/views/partials/tree_view.py
@@ -20,6 +20,7 @@ class FolderItem:
         self.parentItem = parent
         self.path = path
         self.itemData = [name, modified]
+        self.contained_size = 0
         self.childItems = []
         self.checkedState = False
         self.files_with_attributes = files_with_attributes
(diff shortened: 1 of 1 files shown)


m3nu:
Awesome! We had an open PR for this actually.

m3nu:
Fixes #1140

skrap:
> Awesome! We had an open issue for this actually.

Oh, well feel free to disregard this one, then! Or take whichever you like better.

m3nu (approved):
This is a useful addition and I didn't notice any performance problems. Might start using the diff feature actually. 😄

Any objections or comments, @Hofer-Julian ? Since this feature is your baby.

m3nu:
We also have #1138 to add sorting by size for diffs. It adds more code, but isolated in one class. And the contributor didn't adjust the tests yet.

Hofer-Julian (approved):
I have no machine available to test it at the moment.
However, the code change is simple enough and the use case legit.
Thanks!

PR #1166, opened 2022-01-09
Fix showing first log message when `borg check` fails.

real-yfprojects:
Revert 19544050da7dc3fe14d11f084b57906a2a3452cb.

### DIFF
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -305,7 +305,7 @@ class VortaApp(QtSingleApplication):
         repo_url = result['params']['repo_url']
         returncode = result['returncode']
         errors: List[Tuple[int, str]] = result['errors']
-        error_message = errors[0][1] if errors else ''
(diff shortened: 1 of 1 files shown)


m3nu:
Actually both options are fine: with and without this patch:

- Without this patch: first error message = "Index object count mismatch."
- With this patch: last error message = "Completed repository check, errors found."

I find the first more precise. Probably hard to say which is better for each situation.

Should we still merge this or just leave it?

real-yfprojects:
That is a good point. My testing of borg error message leads to the same conclusion: Most of the time the last error message will be less helpful than the first one. Still warnings will be shown instead of error messages if they occur earlier. I am not sure whether that has to be changed since warning messages are basically as critical as error messages in case of `borg check` and the message box always points to the logs for more details.

real-yfprojects:
Should I close this? @m3nu

Issue #930, opened 2021-04-01
cancelling the archive check doesnt kill the underlying borg command


sandrotosi:
**Describe the bug**
cancelling the archive check doesnt kill the underlying borg command

**To Reproduce**
1. have a backup enabled with the option "Validate repository data every ..." enabled
2. when the check step is running, click cancel at the bottom right
3. the action is marked as cancelled but the underlying `borg`  command is not killed and keeps running.

**Environment (please complete the following information):**
- OS: Debian unstable
- Vorta version: 0.7.5
- Installed from: debian package

m3nu:
Wow. We are already *very* aggressive in killing that process. Basically if it doesn't react to SIGINT within X seconds, the whole process group is terminated. So a bit surprising.

m3nu:
Is this a remote or local repo?

sandrotosi:
this is a "local" repo, as in it's an NFS share mounted on a local directory

m3nu:
Tried to reproduce this on a VMWare-mounted share, but failed. Maybe it needs a slow network share to reproduce.

<img src="https://user-images.githubusercontent.com/3916435/114202414-bcadad00-9989-11eb-8e0b-b29333670b18.jpg" width="250" />

sandrotosi:
i wouldnt call this a slow network share: it's a 1gbps ethernet cabled connection from my laptop, to a switch, to a nas.

FTR it happened again a couple of days ago (that's the weekly check running *not* the backup): i decided to cancel the check but the process didnt die, so i had to kill it by hand

m3nu:
I also tried cancelling, when running the `borg check` command to a remote repo over the internet.

Will cancelling also fail for you for remote repos or just network shares?

m3nu:
Here the "kill code", if you want to play with it: https://github.com/borgbase/vorta/blob/master/src/vorta/borg/borg_thread.py#L277

sandrotosi:
> I also tried cancelling, when running the `borg check` command to a remote repo over the internet.

ok this is interesting: did you go to the "Archives" tab and press "Check" at the bottom of it? i was able to cancel the check by following that sequence of steps.

The situation for which i reported this bug is slightly different: i cannot kill the check when executed automatically after a backup, as scheduled by enabling "Validate repository data ever X weeks" in the "Schedule" tab.

m3nu:
Oh. I started it manually from *Archives*.

[...]

PR #1179, opened 2022-01-26
Upgrade llfuse.

m3nu:
See also https://github.com/borgbase/vorta/issues/930#issuecomment-1021717045

@Hofer-Julian

### DIFF
--- a/flatpak/dependencies/python3-llfuse.json
+++ b/flatpak/dependencies/python3-llfuse.json
@@ -8,8 +8,8 @@
     "sources": [
         {
             "type": "archive",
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
At first glance I don't see how that relates to the comment, but upgrading llfuse is fine of course.
Regarding flatpak, the relevant part for flathub lives [here](https://github.com/flathub/com.borgbase.Vorta) with an updated branch [here](https://github.com/flathub/com.borgbase.Vorta/pull/51). I will merge the linked PR as soon as Ubuntu 22.04 is out so we can ignore the problems on Kubuntu 20.04.

Hofer-Julian (approved):

PR #1180, opened 2022-01-26
Update label displaying next backup time.

### LINKED ISSUE #1116: Next Backup time under the schedule tab is never updated
**Describe the bug**

When using scheduled backups the value for 'Next Backup' is never updated after the first run.

70vorta

The current date/time is `2021-11-18 10:17`

The next backup is **actually** scheduled to run: `2021-11-18 11:00`. This runs as expected.

The `Next Backup` field under the `Schedule` tab is displaying: `2021-11-17 14:00` which is clearly incorrect.

The value is never updated after the first time a scheduled backup is set.

The next scheduled backup **is** shown correctly when right clicking the Vorta icon in the panel - I believe this was fixed by #1099 

**To Reproduce**

1. Set up a scheduled backup
2. Let the scheduled backup run
3. Check the value of the `Next Backup` field under the `Schedule` tab - this will be incorrect and will continue to display the first scheduled backup time indefinitely.

**Environment:**

- OS: Arch Linux
- Vorta version: 0.8.1
- Installed from: Arch AUR. `pacman -Q vorta-git` outputs: `vorta-git r578.0dafa9d-1` (Built from 0dafa9d)
- Borg version: 1.1.17

real-yfprojects:
Always update the label in the schedule tab that displays the next time to start a backup when the scheduler updates its timer.
Fixes #1116.

* src/vorta/scheduler.py (VortaScheduler): Add `schedule_changed` signal.

* src/vorta/scheduler.py (VortaScheduler.set_timer_for_profile): Emit `schedule_changed` signal.

* src/vorta/views/schedule_tab.py (ScheduleTab): Connect `draw_next_scheduled_backup` to `schedule_changed` signal.

### DIFF
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -19,6 +19,10 @@ logger = logging.getLogger(__name__)
 
 
 class VortaScheduler(QtCore.QObject):
+
+    #: The schedule for the profile with the given id changed.
+    schedule_changed = QtCore.pyqtSignal(int)
+
     def __init__(self):
         super().__init__()
         self.timers = dict()  # keep mapping of profiles to timers
@@ -117,6 +121,9 @@ class VortaScheduler(QtCore.QObject):
         self.timers[profile_id] = {'qtt': timer, 'dt': next_run}
 
         self.lock.release()
+
+        # Emit signal so that e.g. the GUI can react to the new schedule
+        self.schedule_changed.emit(profile_id)
 
     def reload_all_timers(self):
         logger.debug('Refreshing all scheduler timers')
--- a/src/vorta/views/schedule_tab.py
+++ b/src/vorta/views/schedule_tab.py
@@ -77,6 +77,9 @@ class ScheduleTab(ScheduleBase, ScheduleUI, BackupProfileMixin):
         self.validationWeeksCount.valueChanged.connect(
             lambda new_val, attr='validation_weeks': self.save_profile_attr(attr, new_val))
 
+        # Connect to schedule update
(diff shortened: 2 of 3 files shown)


real-yfprojects:
I was able to track the error down:

When `VortaApp` is initialized it creates a main window which in turn creates a `ScheduleTab` connecting to the scheduler. In the tests (conftest.init_db) the main window is deleted while the schedule tab is not while remaining connected to the signal. This instance will then produce the error seen because it isn't able to access its parent window.

This is clearly an issue with the testing setup or maybe with `VortaApp` creating the main window. Could you have a look at the configuration of the test and try to fix this, @m3nu?

m3nu:
Redoing all the tests sounds like a big task just to preview the next scheduler run more often.

I keep the app around, but reset the database and windows during tests. Destroying the whole app would take much longer and caused segfaults from Qt, as it expects only one Qt app per process. 😬

real-yfprojects:
I implemented a work-around for the failing test.

m3nu:
That's a clever implementation and good workaround for the test. Did something like this myself in a few places. Good job!

Let me run this locally for a while and then merge. Thanks!

real-yfprojects:
Do you know that you don't have to merge the master into the feature branch if you merge it into master afterwards? In case you want to check the tests you can also rebase the branch from within the github ui or using `git rebase master`. That won't add a new commit to the history.

m3nu (approved):

m3nu:
Works as promised. Very elegant solution. Thanks for taking care! 💃

PR #1193, opened 2022-02-04
Change default returncode for log entries to -1.

### LINKED ISSUE #1183: When execution of a (create) job is terminated the logs do not contain `returncode=2`.
**Describe the bug**
When the job starts, an event log is created with `start_time = end_time = now` and `returncode=1`. 
If the computer is shut down for example the job won't be able to finish and its abortion can't be handled by the vorta process.
Still the logs will keep a returncode of `1`. This is a problem for the *create* command since the scheduler will think it was a successful backup.
Instead `returncode` should be set to `2` (=abnormal termination) and only updated to 0 or 1 (=no errors or minor permission errors) after the backup task is complete.

Furthermore, the scheduler should filter by `returncode` of the event and only consider events with 0 or 1 return codes.

I don't know whether this behaviour is a problem for other kinds of jobs as well. Additionally the title of this issue might not be the most fitting.

Related to #1130.

real-yfprojects:
Fixes #1183.

### DIFF
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -1,17 +1,19 @@
 import logging
 import threading
-from datetime import datetime as dt, date, timedelta
+from datetime import date
+from datetime import datetime as dt
+from datetime import timedelta
 
 from PyQt5 import QtCore
 from PyQt5.QtWidgets import QApplication
+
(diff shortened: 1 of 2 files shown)


m3nu (approved):
Works as advertised 👏

PR #1194, opened 2022-02-05
Remove installing python

tiziodcaio:
Python 3.9 is included in latest images of macos-10.15
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md


m3nu:
I'll take it. And it seems we need some more cleanup tasks with regards to pyinstaller and PyQt pinning. Good excuse to add that too.

Issue #238, opened 2019-03-24
Adjustments for Borg 1.2.0


ThomasWaldmann:
Not sure if anyone ever has tested it, but it would be good to do so in case anything in borg or vorta needs changes.

**Do NOT use borg 1.2.0 alpha on your production repos, always use copies or fresh repos.**

https://github.com/borgbackup/borg/releases

m3nu:
Nice changelog. I'll try it out. We use the `--list` option with `borg extract`. This may need attention (find out which Borg version is used).

m3nu:
Did a rough test. Extract command still works, as does the rest. Had to use this to install on macOS:

`$ BORG_OPENSSL_PREFIX=/usr/local/Cellar/openssl/1.0.2q pip install "borgbackup==1.2.0a5"`

- client: `1.2.0a5`
- server: `1.1.8`

ThomasWaldmann:
Yup, this is expected if it can't be discovered by pkg-config.

And on macOS (with brew?) it can't do the latter, if one does not manually adjust PKG_CONFIG_PATH.

If your python was compiled with openssl 1.1, you can also point borg to 1.1.

m3nu:
I only have openssl 1.0.2q. Homebrew offers 1.1. as a kind of special version. I can try that next.

Setting `PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig` didn't help, so I assume pkg-config doesn't know about it, as you say in https://github.com/borgbackup/borg/issues/4444.

ThomasWaldmann:
did you use ` export PKG_CONFIG_PATH=...`?

m3nu:
Yes. Wasn't enough.

I should add that I recently installed a second/older XCode because I needed it for some old SWIFT code. This may confuse it. I do use `xcode-select` to switch CLI tools between those versions. So maybe I'm not the most representative example.

m3nu:
Alpha6 still working. Used the pre-compiled binary this time. Will test installing from PIP next.

ThomasWaldmann:
Just coming to my mind: does vorta ever call `borg compact` if it detects borg 1.2?

m3nu:
Not currently, but I guess it should. It could run at the same time as the `prune` command. So the option name would change and Vorta would run an additional command if `borg >= 1.2.0`.

Added a checklist item above.

ktosiek:
[...]

PR #1196, opened 2022-02-07
Work around slow --stats command in Borg 1.2

m3nu:
We can't use `borg create --stats` any more, since it can be very slow under certain conditions.

This PR removes the related `snapshot_id`, which wasn't used much anyways and uses archive name and repo id as primary keys to update existing archives.

See also:
- #238 
- https://github.com/borgbackup/borg/discussions/6166#discussioncomment-2040866
- https://github.com/borgbackup/borg/issues/4360#issuecomment-1031500826


real-yfprojects:
Vorta could record the backup duration itself.

m3nu:
Closing since this is already fixed upstream.

PR #1197, opened 2022-02-10
Add reasonable, default ignores to flake8.

real-yfprojects:
The tool *flake8* used in the tests runs on github allows for ignoring specific codes. The default named in the [docs](https://flake8.pycqa.org/en/latest/user/options.html?highlight=ignore#cmdoption-flake8-ignore) are `E121,E123,E126,E226,E24,E704,W503,W504`, although these default reasonable it seems as if flake actually doesn't use them by default. I would therefore like to add them to the test runs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,6 +92,6 @@ jobs:
         pip install .
         pip install -r requirements.d/dev.txt
     - name: Run Flake8
-      run: flake8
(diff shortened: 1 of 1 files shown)


real-yfprojects:
> Did you run into a specific issue that prompted this PR?

I ran into the issue multiple times that I wanted to wrap an if-statement since it exceeded 79 characters but `W503` and `W504` prevent this because they forbid line breaks around binary operators.

PR #1204, opened 2022-02-19
Rewrite scheduler.

### LINKED ISSUE #1182: Missed backups aren't catched up if no previously scheduled backup ran.
# Issue
When vorta checks for a missed backup to catch up it queries for the last scheduled backup and will calculate whether this is too long in the past. But if there aren't any scheduled backups it will schedule a new one anyway e.g. in an hour from the current time. But if vorta is closed and this scheduled backup is missed in conclusion it won't be run on startup.

Related to #1130.

# Proposal

Vorta always calculates the next backup time from the last backup made whether it is scheduled or not. I think this is the wanted/expected behaviour. When choosing *run every hour* the result should be backups that are an hour apart from each other.

This will fix the issue.

### LINKED ISSUE #1184: Changing time in *fixed* schedule mode will break catching up.
If the fixed schedule mode is changed e.g. from 3 pm to 3 am at 2 am in the morning and vorta isn't running at 3 am, the backup will not be run when starting up vorta shortly after. It will be run when starting vorta at 3 pm or later.

Relates to #1130.

real-yfprojects:
The scheduler will now react to a new fixed time also if 24h haven't passed yet.
All in all the code is supposed to be cleaner and easier understandable.
The logging was slightly modified.

I wrote some tests for the schedule routine itself, but there are still some test cases for the feature that catches up missed backups missing.

Fixes #1184 and fixes #1182.

### DIFF
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -3,6 +3,7 @@ import threading
 from datetime import date
 from datetime import datetime as dt
 from datetime import timedelta
+from typing import Dict, Union
 
 from PyQt5 import QtCore
 from PyQt5.QtWidgets import QApplication
@@ -25,7 +26,10 @@ class VortaScheduler(QtCore.QObject):
 
     def __init__(self):
         super().__init__()
-        self.timers = dict()  # keep mapping of profiles to timers
+
+        #: mapping of profiles to timers
+        self.timers: Dict[int, Dict[str, Union[QtCore.QTimer, dt]]] = dict()
+
         self.app = QApplication.instance()
         self.lock = threading.Lock()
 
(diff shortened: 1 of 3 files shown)


m3nu:
I'm happy that someone was able to improve my initial implementation. Seems clearer and easier to follow. So we should definitely use it. Also great to have some tests at last to make sure future changes improve in the same way.

You're still planning the missing tests you mention to this PR?

real-yfprojects:
> You're still planning the missing tests you mention to this PR?

Currently not. We could open a *good first issue* for that.

m3nu (approved):

m3nu:
The PR makes the scheduler easier to follow and addresses some concerns we got over the old implementation. It also adds some much-needed tests. Good job, @real-yfprojects! 👍

PR #1205, opened 2022-02-20
Add new `borg compact` command

m3nu:
Adds support for the new `borg compact` command which cleans up segments marked for deletion. See #238 

After this feature works for everyone, there should be a checkbox to run it automatically every x weeks, like we do for `borg check`.

Screen Shot 2022-02-20 at 10 18 28


real-yfprojects:
Do you plan to merge this before #1176? In this case #1176 will need some additional changes for this new button.

m3nu:
> Do you plan to merge this before #1176? In this case #1176 will need some additional changes for this new button.

Totally missed #1176. Since it's more complex it should go first. Checking now..

m3nu:
With Borg 1.2 due to be released tomorrow, I'll have to merge this before #1176, I'm afraid.

real-yfprojects:
Not a problem.

src/vorta/borg/compact.py
```diff
@@ -0,0 +1,48 @@
...
+            return ret
+        else:
+            ret['ok'] = False  # Set back to false, so we can do our own checks here.
+
+        if not borg_compat.check('COMPACT_SUBCOMMAND'):
+            ret['ok'] = False
+            ret['message'] = trans_late('messages', 'This feature needs Borg 1.2.0 or higher.')
+            return ret
+
+        cmd = ['borg', '--info', '--log-json', 'compact', '--cleanup-commits']
```
ThomasWaldmann, line 42: `--cleanup-commits` was intended for one time use (per repo), afterwards i guess it is not really needed any more.

m3nu: We can keep it for a while until most repos are managed with 1.2 or cleaned up. This command won't run too often. Currently it's manually. Later we can run it together with prune every few weeks.

ThomasWaldmann: Well, my point is that existence of 17byte files are not a bug in general. These are just separated commits.

In borg 1.1, there is a bug creating or leaving behind too many of them.
In borg 1.2 this bug was resolved, but separated commits are still created intentionally.

So I guess I rather would not remove them after the switch from 1.1 to 1.2.

real-yfprojects: So this parameter option is only intended for very specific repositories that are known to be spammed with commits.

ThomasWaldmann: it's intended to be run once (per repo) when borg is upgraded to 1.2.

m3nu: So would you remove it or keep it? What I’m probably asking: is there any harm in running it too often?

real-yfprojects: Maybe vorta could detect whether a repository is still from 1.1 and run compact in this case.

ThomasWaldmann: Well, that cleanup operation is rather simplistic: it just removes all separated commits (17b files). The compact_segments running afterwards will then add a commit at the end.

If the repo is ok before, it will also be ok afterwards, but IIRC I didn't analyze all error conditions that could also exist.

ThomasWaldmann: besides that, if something goes wrong in a repo (like losing some files), it is nice to have multiple commits present.
if there is some commit before the lost files, one can just roll back to that commit and all is fine again.
you don't have that if you regularly remove all commits except the final one.

m3nu: Then I'll probably remove the `--cleanup-commits` option when touching the code next time. No need to overdo it.

> besides that, if something goes wrong in a repo (like losing some files), it is nice to have multiple commits present.

This is for `borg compact` in general? And it shouldn't run too often? I was planning to run it every few weeks, similar to our `borg check` option.

ThomasWaldmann: matter of taste / requirements:
- more safety: run less often
- more space savings: run more often

real-yfprojects: In that case I would add an additional option to the schedule tab like the one for validation but for compacting.

PR #1219, opened 2022-03-12
Improve the extract dialog and the dialog showing diff result.

real-yfprojects:
I went over board a little bit and implemented a new, more abstract `FileTreeModel` for use with tree views in general and especially with the ones in the diff result dialog and the extract dialog. 

I have already made the transition of the diff result dialog to the new model.
The dialog now features:

* Three view modes: Tree,  Simplified Tree and Flat
* A tooltip for items with all the diff information
* Sorting
* A button to keep folders on top while sorting
* A context menu to copy an item's path and for expanding all children of the item recursively
* A button to collapse all items
* The dialog now doesn't stay on top of the main window. This allows to compare multiple diff results easily.

Closes #1138.


m3nu:
Sounds promising! Any screenshots so people can quickly see it in action without pulling your branch?

real-yfprojects:
screen
size-folders-not-ontop
size-folders-ontop

context-menu

tooltip-path
tooltip-info
context-menu-2
context-menu
flat
simplified-tree

m3nu:
🤗 Looks great! @Hofer-Julian might be interested, as this feature is his baby.

Hofer-Julian:
The screenshots look good, definitely an improvement over what we have now.

The current implementation takes a bit of time to parse and display borg's output. @real-yfprojects did you check how your version copes with big repos?

real-yfprojects:
> The screenshots look good, definitely an improvement over what we have now.
> 
> The current implementation takes a bit of time to parse and display borg's output. @real-yfprojects did you check how your version copes with big repos?

No I don't have any big ones. Do you know an easy way of creating one?

Hofer-Julian:
> > The screenshots look good, definitely an improvement over what we have now.
> > The current implementation takes a bit of time to parse and display borg's output. @real-yfprojects did you check how your version copes with big repos?
> 
> No I don't have any big ones. Do you know an easy way of creating one? And how to perform benchmarks with python on it?

I don't have good answers on either of your questions. When your PR is in a reviewable state I can try it on one of my personal repos.

m3nu:
‘borg benchmark crud’ will fill a repo with dummy files. Should be tested with at least 250k files, which can take some profiling to remove bottlenecks.

real-yfprojects:
What do you prefer in the simplified tree mode when a parent has a child that was changed:

simplify-twig-parent
dont-simplify-twig-parent

real-yfprojects:
I optimised the new implementation by replacing the use of `PurePath` with `Tuple[str]` and by removing the code updating the persistent model indices. I also changed the way children are stored. I profiled two different approaches for that. The first one stores the children in a sorted list and uses bisect to index and insert items. The second one uses a python `dict`. The table below allows comparing the two approaches and the old implementation in different disciplines. The profiled time was measured in seconds with 16033 changes.

The branch currently uses the dict approach. Which approach should I use? Personally I prefer the bisect approach because it results in noticeably cleaner code.

#### Profile results

Task    | bisect and list | dict | old implementation
------ | ----- |  ----- | -----
parse diff results |0.951 | 0.566 | 0.222 and <br> 0.107 on json
walk model and access all indices in flat mode| 0.167 | 0.205 | -
walk model and access all indices in tree mode | 0.844 | 1.248 | 2.771
walk model and access all indices in tree mode | 0.879 |  1.426 | -

#### Ratios
Task    | bisect - dict | dict  - old| old - bisect
------ | ----- |  ----- | -----
parse diff results |1.680 | 2.550 <br> 5.290 | 0.233 <br> 0.113
walk model and access all indices in flat mode| 0.815 | - | -
walk model and access all indices in tree mode | 0.676 | 0.450 | 3.283
walk model and access all indices in tree mode | 0.616 | - | -

m3nu:
What do we learn from this? Am I reading this correctly that parsing got slower, but the rest faster?

The tree model is also used for extract? There we have more data.

real-yfprojects:
> What do we learn from this? Am I reading this correctly that parsing got slower, but the rest faster?

Yes.

> The tree model is also used for extract? There we have more data.

That's what I am currently planning to do.

real-yfprojects:
> What do you prefer in the simplified tree mode when a parent has a child that was changed:
> 
> simplify-twig-parent dont-simplify-twig-parent

? @Hofer-Julian @m3nu

Hofer-Julian:
> > What do you prefer in the simplified tree mode when a parent has a child that was changed:
> > simplify-twig-parent dont-simplify-twig-parent
> 
> ? @Hofer-Julian @m3nu

How would you display the first case if multiple children of the same folder change?

real-yfprojects:
> How would you display the first case if multiple children of the same folder change?

The same as for `/home/theuser/Documents`.

Hofer-Julian:
> > How would you display the first case if multiple children of the same folder change?
> 
> The same as for `/home/theuser/Documents`.

Ah yes, then the first. But no strong feelings there.

real-yfprojects:
I tried the second because I found it difficult to differentiate between the path and the file name when looking at the first approach.

real-yfprojects:
> What do you prefer in the simplified tree mode when a parent has a child that was changed:
> 
> simplify-twig-parent dont-simplify-twig-parent

@m3nu I prefer the second one, Hofer-Julian the first. What about you?

m3nu:
Second seems more consistent, but no strong opinion. You build it, so you break the tie.

real-yfprojects:
> The current implementation takes a bit of time to parse and display borg's output.

The parsing now runs in a separate thread and the status label is updated accordingly.

m3nu:
Some changes to Borg's diff handling. Could be added in this PR: https://github.com/borgbackup/borg/issues/6483#issuecomment-1102577767

real-yfprojects:
The last two commits where added and implement the new treemodel for the extract dialog. The behaviour of the checkboxes now matches the extraction result.

real-yfprojects:
The extract dialog is completely implemented now. The only task left is adding some more icons to the controls in the diff and extract dialog.

real-yfprojects:
This PR is ready for review although the tests currently fail. The tests do succeed when run locally but fail on github also when run on the master branch.

real-yfprojects:
> How do you know that? They seem to be green on current master :)

https://github.com/borgbase/vorta/actions/runs/2678440749

The check sucess shown for each commit only reflect the results from CI runs targetting the commit.

m3nu:
Looks like it just needs to be set to use a different Qt plugin for headless tests. Didn't test which one works best.

Hofer-Julian (dismissed):
I finally had time to look into that PR.
The changes are really big, so I wasn't able to fully verify all changes, but apart from that commented code everything looked fine.

I've also tested the behaviour and everything worked as advertised.
From my side this is ready to be merged.
Thanks @real-yfprojects

real-yfprojects:
> From my side this is ready to be merged.

Great. I think squash merge would be the wrong strategy for this PR because it covers a lot of changes and a big variety of changes. I already squashed and rearranged a lot of commits and cleaned the git history of this branch.

XXXBold:
Wow, I was recently thinking about improving that dialog, your version looks absolutely gorgeous, thanks for your work!

Nitpick: Shouldn't "Copy" from context menu probably be renamed to e.g. "Copy Path"? Otherwise the user might think he's copying the file itself.

real-yfprojects:
> Nitpick: Shouldn't "Copy" from context menu probably be renamed to e.g. "Copy Path"? Otherwise the user might think he's copying the file itself.

Well, he is. The path gets copied to the clipboard as text and as an URI.

m3nu:
> Should we disable tests on Linux for now, or do you already see how this can be fixed @m3nu?

Fixed this in the `master` branch, but couldn't rebase this PR. Change is just one dependency. https://github.com/borgbase/vorta/commit/33639aeaed5f0c3255b58925d5e9a52a2f0edd77

real-yfprojects:
Rebased

m3nu (approved):

m3nu:
2 approvals. Just merge it when you think it's ready, @real-yfprojects

real-yfprojects:
Usually I prefer rebase merges but in this case I think that a standard merge would fit the best. It appears like a squashed commit in the git history but still contains the information about the grouping of the changes. Though merge commits are forbidden in the master branch currently.

m3nu:
We have those 3 available. Since this PR is pretty clean, we don't need to squash, if you prefer that.

Screen Shot 2022-07-25 at 20 26 54

real-yfprojects:
The first one is only available for you since you have the ability to bypass all branch protection rules and so on.
grafik

m3nu:
Done. Hope this went as expected.

real-yfprojects:
> Done. Hope this went as expected.

Yes, it did.

PR #1229, opened 2022-03-22
Issue 1223 - Add missing translated strings on archive tab

delfino-nunez:
These translations will allow the strings to be translated to different languages, instead of showing them always in English.

missing-translations


m3nu (approved):

m3nu:
Great! We must have overlooked this when adding the other strings. Thanks for adding!

Fixes #1229

PR #1235, opened 2022-03-25
More comfortable diff selection without extra dialog.

real-yfprojects:
This removes the dialog to select two archives for comparing and instead allows to select up to two archives at once. If two archives are selected the diff button is enabled. Clicking it will then directly start the diff action for the selected archives.
A screenshot won't depict the changes. Just check it out.


Hofer-Julian:
> A screenshot won't depict the changes. Just check it out.

Github supports videos

real-yfprojects:
> Github supports videos

I don't. I always find it a hassle to record a single window on linux and setting that up in a VM is more work than writing the PR itself.

real-yfprojects:
Rebase triggered the tests. All of them have passed.

m3nu (approved):

PR #1236, opened 2022-03-25
Fix use of `self.tr` for translation.

real-yfprojects:
* src/vorta/views/extract_dialog.py
* src/vorta/views/main_window.py
* src/vorta/views/profile_add_edit_dialog.py
* src/vorta/views/repo_add_dialog.py


m3nu (approved):

real-yfprojects:
Today was a maintenance. Maybe that's the cause. You can run the workflow manually as described here:
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

PR #1244, opened 2022-03-30
Resolve #1242 Add translations to treeView columns

### LINKED ISSUE #1242: Translation missing on "Extract" dialog for grid columns and "Extract" button.
**Describe the bug**
The "Extract" dialog grid columns are not translated and the "Extract" button is not translated also.

**To Reproduce**
1. Go to 'Archives'
2. Select an archive from the list
3. Click on 'Extract'
4. See untranslated strings

Selection_003

- OS: Linux 5.13.0-37-generic # 42~20.04.1-Ubuntu SMP Tue Mar 15 15:44:28 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
- Vorta version: 0.8.3
- Installed from: GitHub repo

delfino-nunez:
I'm not sure if this is the right approach for these dialogs since the translate functions can't be used on the partial tree_view.py file.

Let me know if this is not the right way.

Selection_002

### DIFF
--- a/src/vorta/views/extract_dialog.py
+++ b/src/vorta/views/extract_dialog.py
@@ -87,6 +87,7 @@ class ExtractTree(TreeModel):
         super().__init__(
             files_with_attributes, nested_file_list, selected_files_folders, parent
         )
+        self.column_names = [self.tr("Name"), self.tr("Modified"), self.tr("Size")]
 
     def data(self, index, role):
         if not index.isValid():


real-yfprojects (changes requested):
You can't use `self.tr` in `partials/tree_view.py` because self is not available in a class context. Instead you can use `translate` which you can import from `vorta.i18n`. It is good practice to translate strings where they are defined. Applied to this PR, you should change `tree_view.py` instead of `extract_dialog.py`.

delfino-nunez:
>  Instead you can use `translate` which you can import from `vorta.i18n`. It is good practice to translate strings where they are defined. Applied to this PR, you should change `tree_view.py` instead of `extract_dialog.py`.

But you can't call translate without `init_translate` right? and also you can't `init_translate` because you can't pass the app instance.
so how do you call translate from this class?

Shouldn't the columns be sent when the `TreeModel` is used, like `LoadingButton`?
`self.createStartBtn = LoadingButton(self.tr("Start Backup"))`

 ```
 File "/home/delfino/vorta/src/vorta/views/diff_result.py", line 12, in <module>
    from vorta.views.partials.tree_view import TreeModel
  File "/home/delfino/vorta/src/vorta/views/partials/tree_view.py", line 192, in <module>
    class TreeModel(QAbstractItemModel):
  File "/home/delfino/vorta/src/vorta/views/partials/tree_view.py", line 195, in TreeModel
    column_names = [translate("test","Name"),"Modified","Size"]
  File "/home/delfino/vorta/src/vorta/i18n/__init__.py", line 77, in translate
    return application.translate(*args, **kwargs)
NameError: name 'application' is not defined
```

**Comment by real-yfprojects** [comment]
I see.
Then this is one of the rare cases where `trans_late` has to be used where the strings are defined. Later when passing the header strings to the table view, `self.tr` has to be called on them.

**Comment by delfino-nunez** [comment]
I can use ```trans_late``` but in the end it will not be translated...

so the question here is if the columns names should be defined in the ```tree_view.py``` file or they should be defined for each call of the ```TreeModel```?

I know that's not the best way, but the way ```trans_late``` is defined will not do anything for the translation.

what do you think?

**Comment by real-yfprojects** [comment]
You wrap the strings in the list `column_names` with `trans_late`. In `TreeModel.headerData` you wrap `self.column_names[section]` with `self.tr`.

**Comment by delfino-nunez** [comment]
> You wrap the strings in the list `column_names` with `trans_late`. In `TreeModel.headerData` you wrap `self.column_names[section]` with `self.tr`.

I tried to do it this way but doesn't work.

` column_names = [trans_late("TreeModel","Name"), trans_late("TreeModel","Modified"), trans_late("TreeModel","Size")] `

and then 

`return self.tr(self.column_names[section])`

doesn't fail, but doesn't translate.

Also it doesn't generate new translations for the columns when running `make translations-from-source`...

So I wonder if I'm doing something wrong or the translation doesn't work in this case.

**Comment by real-yfprojects** [comment]
Patch the following:

```diff
diff --git a/Makefile b/Makefile
index 7cae6d0..62f258f 100644
--- a/Makefile
+++ b/Makefile
@@ -37,8 +37,7 @@ bump-version:  ## Tag new version. First set new version number in src/vorta/_ve
 
 translations-from-source:  ## Extract strings from source code / UI files, merge into .ts.
        pylupdate5 -verbose -translate-function trans_late \
-                          ${VORTA_SRC}/*.py ${VORTA_SRC}/views/*.py ${VORTA_SRC}/borg/*.py \
-                          ${VORTA_SRC}/store/*.py ${VORTA_SRC}/assets/UI/*.ui \
+                          $$(find ${VORTA_SRC} -regex ".*[.]\(py\|ui\)") \
                           -ts ${VORTA_SRC}/i18n/ts/vorta.en.ts
 
 translations-push: translations-from-source  ## Upload .ts to Transifex.
```

PR #1253, opened 2022-04-05
Use fnmatch for exclude patterns.

real-yfprojects:
The standard exclude pattern used by borg is *fnmatch*. 

* src/vorta/assets/UI/sourcetab.ui : Change `E.g. **/.cache` to `E.g. */.cache`.

* Use fnmatch to exclude files when calculating source stats.

### DIFF
--- a/src/vorta/assets/UI/sourcetab.ui
+++ b/src/vorta/assets/UI/sourcetab.ui
@@ -184,7 +184,7 @@
         <string/>
        </property>
        <property name="placeholderText">
-        <string>E.g. **/.cache</string>
+        <string>E.g. */.cache</string>
        </property>
       </widget>
      </item>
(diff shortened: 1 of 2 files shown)


m3nu:
Thats fantastic and gets rid of them copied Borg code. Results match up too?

real-yfprojects:
> Thats fantastic and gets rid of them copied Borg code. Results match up too?

Yes, but I had to copy some code from borg to make fnmatch matching coherent. And I had to make vorta check whether to include directories which it didn't do before.

real-yfprojects:
Introduced some optimisations.

m3nu (approved):

vraimonds:
Hello! 👋🏻 

How do I do an inverted exclude now? I'd like to exclude all the files and folders in **~/Library/** folder except for **Keychains** and let's say **Containers**  now. I tried to write a glob pattern ~/Library/!(Keychains|Containers) but it seems to not be working. Previously I had a regex for this, which was working and is not working anymore since this change got introduced.

Issue #1233, opened 2022-03-24
Unhandled exception windows when password manager is locked


supermihi:
**Describe the bug**
I have a scheduled backup with Vorta. Today it failed because apparently the password manager was locked, leading to the screenshot below that asks me to report this issue on GitHub. Imho there should be a helpful error message instead of this exception; also Vorta was completely unresponsive after that, i.e., I could neither click the "ok" button to close the exception window nor do anything else. 

**Environment:**
- OS: Arch Linux
- Vorta version: 0.8.3
- Installed from: Arch Linux AUR

Vorta and Borg versions can be found in Main Window > Misc Tab.

**Additional context**
![screenshot](https://user-images.githubusercontent.com/102419/159922730-d70aa9ac-488c-4436-ae22-00e73f379add.png)

m3nu:
This happens in the `secretstorage` package, but we should probably deal with it somehow.

supermihi:
I investigated a little more, and the problems seems to be related with me upgrading to KeePassXC 2.7.0, which implements a  [dialog for unlocking secrets](https://github.com/keepassxreboot/keepassxc/pull/6943). For some reason, this doesn't seem to work well with `secretstorage`, because I never see that dialog (works well with other clients).

supermihi:
As there's a call to `ensure_not_locked()` in the stack: perhaps Vorta should call something like `unlock_if_required()` or the like (just guessing)?

supermihi:
`collection.unlock()` seems to be the correct method (https://secretstorage.readthedocs.io/en/latest/#locking-and-unlocking-collections)

Aeysol:
It seems that KeePassXC with v2.7.0 has changed the initial status of an item when fetched using the SecretStorage API. The new dialog that is presented to users now sets the state of the item initially to _locked_ and after manual confirmation by the user to _unlocked_.

I've patched the `get_password()` method in the file [secretstorage.py](https://github.com/borgbase/vorta/blob/master/src/vorta/keyring/secretstorage.py#L43):
```py
def get_password(self, service, repo_url):
    if self.is_unlocked:
        asyncio.set_event_loop(asyncio.new_event_loop())
        collection = secretstorage.get_default_collection(self.connection)
        attributes = {'application': 'Vorta', 'service': service, 'repo_url': repo_url}
        items = list(collection.search_items(attributes))
        logger.debug('Found %i passwords matching repo URL.', len(items))
        if len(items) > 0:
            item = items[0]
[...]

PR #1255, opened 2022-04-09
SecretStorage: Unlock item if it is locked. Fixes #1232 and #1233.

### LINKED ISSUE #1232: Vorta pour Borg
`Erreur non prise en charge, veuillez remplir un rapport de bug avec ce texte à l'adresse suivante
https://github.com/borgbase/vorta/issues/new
  File "/app/lib/python3.7/site-packages/vorta/application.py", line 124, in create_backup_action
    msg = BorgCreateJob.prepare(profile)
  File "/app/lib/python3.7/site-packages/vorta/borg/create.py", line 74, in prepare
    ret = super().prepare(profile)
  File "/app/lib/python3.7/site-packages/vorta/borg/borg_job.py", line 151, in prepare
    ret['password'] = cls.keyring.get_password('vorta-repo', profile.repo.url)
  File "/app/lib/python3.7/site-packages/vorta/keyring/secretstorage.py", line 42, in get_password
    return items[0].get_secret().decode("utf-8")
  File "/app/lib/python3.7/site-packages/secretstorage/item.py", line 96, in get_secret
    self.ensure_not_locked()
  File "/app/lib/python3.7/site-packages/secretstorage/item.py", line 48, in ensure_not_locked
    raise LockedException('Item is locked!')
secretstorage.exceptions.LockedException: Item is locked!`

KeePassXC

Aeysol:
Some password managers (like KeePassXC) will lock an SecretStorage item behind a manual user approval dialog even when the collection is already unlocked. This prevents Vorta from using SecretStorage compatible password managers that exhibit this behavior.

This pull request fixes the way an item is obtained so any manual user approval dialog is triggered.

### DIFF
--- a/src/vorta/keyring/secretstorage.py
+++ b/src/vorta/keyring/secretstorage.py
@@ -21,9 +21,8 @@ class VortaSecretStorageKeyring(VortaKeyring):
         except secretstorage.exceptions.SecretServiceNotAvailableException as e:
             logger.debug("SecretStorage provider or DBus daemon is not available.")
             raise e
-        else:
-            asyncio.set_event_loop(asyncio.new_event_loop())
-            self.collection = secretstorage.get_default_collection(self.connection)
+        asyncio.set_event_loop(asyncio.new_event_loop())
+        self.collection = secretstorage.get_default_collection(self.connection)
 
     def set_password(self, service, repo_url, password):
         """
@@ -60,9 +59,7 @@ class VortaSecretStorageKeyring(VortaKeyring):
     @property
     def is_unlocked(self):
         # unlock() will return True if the unlock prompt is dismissed
-        if self.collection.is_locked() and self.collection.unlock():
-            return False
-        return True
+        return not (self.collection.is_locked() and self.collection.unlock())
 
     @classmethod
     def get_priority(cls):
--- a/src/vorta/keyring/secretstorage.py
+++ b/src/vorta/keyring/secretstorage.py
@@ -51,7 +51,10 @@ class VortaSecretStorageKeyring(VortaKeyring):
             items = list(collection.search_items(attributes))
             logger.debug('Found %i passwords matching repo URL.', len(items))
             if len(items) > 0:
-                return items[0].get_secret().decode("utf-8")
+                item = items[0]
+                if item.is_locked():  # Some providers lock items until the user manually approves it
+                    item.unlock()
+                return item.get_secret().decode("utf-8")
         return None
 
     @property
--- a/src/vorta/keyring/secretstorage.py
+++ b/src/vorta/keyring/secretstorage.py
@@ -1,6 +1,5 @@
 import asyncio
 import os
-import sys
 
 import secretstorage
 
@@ -19,26 +18,23 @@ class VortaSecretStorageKeyring(VortaKeyring):
         """
         self.connection = secretstorage.dbus_init()
         asyncio.set_event_loop(asyncio.new_event_loop())
-        secretstorage.get_default_collection(self.connection)
+        self.collection = secretstorage.get_default_collection(self.connection)
 
     def set_password(self, service, repo_url, password):
         """
         Writes a password to the underlying store.
         """
-        try:
+        if self.is_unlocked:
             asyncio.set_event_loop(asyncio.new_event_loop())
-            collection = secretstorage.get_default_collection(self.connection)
             attributes = {
                 'application': 'Vorta',
                 'service': service,
                 'repo_url': repo_url,
                 'xdg:schema': 'org.freedesktop.Secret.Generic'}
-            collection.create_item(LABEL_TEMPLATE.format(repo_url=repo_url),
-                                   attributes,
-                                   password,
-                                   replace=True)
-        except secretstorage.exceptions.ItemNotFoundException:
-            logger.error("SecretStorage writing failed", exc_info=sys.exc_info())
+            self.collection.create_item(LABEL_TEMPLATE.format(repo_url=repo_url),
+                                        attributes,
+                                        password,
+                                        replace=True)
 
     def get_password(self, service, repo_url):
         """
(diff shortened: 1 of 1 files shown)


m3nu:
Awesome! This was overdue. @real-yfprojects @bastiencyr @Hofer-Julian please test on your Linux desktop, if you can. 🙏

src/vorta/keyring/secretstorage.py
```diff
@@ -51,7 +51,10 @@ def get_password(self, service, repo_url):
             items = list(collection.search_items(attributes))
             logger.debug('Found %i passwords matching repo URL.', len(items))
             if len(items) > 0:
-                return items[0].get_secret().decode("utf-8")
+                item = items[0]
+                if item.is_locked():  # Some providers lock items until the user manually approves it
+                    item.unlock()
+                return item.get_secret().decode("utf-8")
```
real-yfprojects, line 57: This won't work if the user dismisses the unlock. An unlocking mechanism is already implemented for collections here:
https://github.com/borgbase/vorta/blob/8126f4e696ed7bb40b8fc870b1dfc746672d9d11/src/vorta/keyring/secretstorage.py#L57-L66

real-yfprojects (changes requested):
While we are at it it would be great if you'd implement using `is_unlocked` when setting the password in [set_password](https://github.com/borgbase/vorta/blob/8126f4e696ed7bb40b8fc870b1dfc746672d9d11/src/vorta/keyring/secretstorage.py#L24-L41) since this method currently doesn't handle the case of a locked collection.
I also left a comment for `get_password`. Maybe there is a clever way to recycle the code in `is_unlocked` for items or both items and collections.

src/vorta/keyring/secretstorage.py
Aeysol: Alright. I will address these issues.

real-yfprojects: I can also take on the remaining issues if you don't have the time.

Aeysol:
@real-yfprojects Kindly asking for re-review.

src/vorta/keyring/secretstorage.py
```diff
@@ -19,53 +18,48 @@ def __init__(self):
...
     @property
     def is_unlocked(self):
         try:
-            collection = secretstorage.get_default_collection(self.connection)
-            if collection.is_locked():  # Prompt for unlock
-                collection.unlock()
-            return not collection.is_locked()  # In case of denial
+            if self.collection.is_locked():  # Prompt for unlock
+                self.collection.unlock()
+            return not self.collection.is_locked()  # In case of denial
```
real-yfprojects, line 60:
```suggestion
            # unlock will return True if the unlock is dismissed
            return self.collection.is_locked() and self.collection.unlock()
```

The try-except statement in this method can also be removed.

real-yfprojects (changes requested):
Your changes are really great and elegant. I noticed some necessary code that you didn't write but could be refactored anyway.

src/vorta/keyring/secretstorage.py
Aeysol: What necessary code is still needed?

real-yfprojects: Only these two lines in the suggested change.

src/vorta/keyring/secretstorage.py
```diff
@@ -15,55 +14,55 @@ class VortaSecretStorageKeyring(VortaKeyring):
...
-        asyncio.set_event_loop(asyncio.new_event_loop())
-        secretstorage.get_default_collection(self.connection)
+        try:
+            self.connection = secretstorage.dbus_init()
+        except secretstorage.exceptions.SecretServiceNotAvailableException as e:
+            logger.debug("SecretStorage provider or DBus daemon is not available.")
+            raise e
+        else:
+            asyncio.set_event_loop(asyncio.new_event_loop())
+            self.collection = secretstorage.get_default_collection(self.connection)
```
real-yfprojects, line 26:
```suggestion
        asyncio.set_event_loop(asyncio.new_event_loop())
        self.collection = secretstorage.get_default_collection(self.connection)
```

The else is not needed since the raise will stop code execution.

src/vorta/keyring/secretstorage.py
```diff
@@ -15,55 +14,55 @@ class VortaSecretStorageKeyring(VortaKeyring):
...
-            collection = secretstorage.get_default_collection(self.connection)
-            if collection.is_locked():  # Prompt for unlock
-                collection.unlock()
-            return not collection.is_locked()  # In case of denial
-        except secretstorage.exceptions.SecretServiceNotAvailableException:
-            logger.debug('SecretStorage is closed.')
+        # unlock() will return True if the unlock prompt is dismissed
+        if self.collection.is_locked() and self.collection.unlock():
             return False
+        return True
```
real-yfprojects, line 65:
```suggestion
        return not (self.collection.is_locked() and self.collection.unlock())
```
This is just a matter of preference, you can also keep the current code.

real-yfprojects (changes requested):

src/vorta/keyring/secretstorage.py
Aeysol: This was just my personal preference. But I've changed it nonetheless.

src/vorta/keyring/secretstorage.py
Aeysol: You're right. It's obvious that it can be simplified to your solution.

src/vorta/keyring/secretstorage.py
real-yfprojects: This is cleaner code due to less branching.

real-yfprojects (approved):
Did you test the code already?
I didn't manage to get KeepassXC running.

Aeysol:
> Did you test the code already? I didn't manage to get KeepassXC running.

Yes. Tried it with an already unlocked instance, completely locked and with the user prompt dismissed. The message Vorta gives after dismissing the user prompt is confusing ("unsupported password manager") but I've found no cleaner way yet.

real-yfprojects:
> Yes. Tried it with an already unlocked instance, completely locked and with the user prompt dismissed. 

Did you test locking only a single item?

> The message Vorta gives after dismissing the user prompt is confusing ("unsupported password manager") but I've found no cleaner way yet.

Where and when do you see this prompt?

Aeysol:
> Did you test locking only a single item?

Yes.
| Collection unlocked? | Item unlocked? | Vorta behavior |
| --- | --- | --- |
| False | Not applicable | "Please unlock your system password manager or disable it under Misc" |
| True | False | "Your repo passphrase was stored in a password manager which is no longer available" |
| True | True | Backup starts |

> Where and when do you see this prompt?

It's the item unlock prompt of KeePassXC which comes up when Vorta starts a backup.

real-yfprojects:
> > Where and when do you see this prompt?
> 
> It's the item unlock prompt of KeePassXC which comes up when Vorta starts a backup.

This is particular to KeepassXC i think. Using gnome keyring I saw a different unlock prompt.

m3nu:
Don't have a lot of Linux desktops to try this, but this PR clarifies the code and gets rid of some try-excepts, which is already a win.

m3nu (approved):

PR #1265, opened 2022-04-19
Add support for fifos in `borg diff` output.

real-yfprojects:
Support for many new file types (fifo, chrdev and blkdev) was added to `borg diff` with borgbackup/borg#6624 and will be included in the next patch.

* src/vorta/views/diff_result.py : Add handling of `added {item_type}` and `removed {item_type}` for all the new item_types to parsing.


m3nu:
This is ready to merge? Then it would be time for a release.

real-yfprojects:
> This is ready to merge? Then it would be time for a release.

Yes, I think so.

m3nu (approved):

PR #1296, opened 2022-05-15
Exclude the last backup creation from log deletion.

real-yfprojects:
The scheduler uses the log entry of the last (scheduled) `create` command to determine the next backup time. Logs older than 180 days (actually 6 days but that seems to be typo in the code) are automatically deleted. 
This PR ensures  that t latest `create` entry and the latest scheduled one will be excluded from deletion so that the scheduler always knows the last backup time.

* src/vorta/store/connection.py (init_db): Remove the doubling of the deletion statement. One statement used 6 days the other one 180 days as expire time. 180 days it is.

* src/vorta/store/connection.py (init_db): Add a subquery to determine last `create` call per profile. Only delete entries that are not in the subquery.

### DIFF
--- a/src/vorta/store/connection.py
+++ b/src/vorta/store/connection.py
@@ -1,6 +1,7 @@
 import os
 from datetime import datetime, timedelta
 
+from peewee import Tuple, fn
 from playhouse import signals
 
 from vorta.autostart import open_app_at_startup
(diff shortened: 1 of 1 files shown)


m3nu (approved):
Verified locally and working as described. Thanks!

PR #1308, opened 2022-05-21
Remove leftover from changing repo menu

### LINKED ISSUE #1298: Switching between profiles doesn't update the archive list
**Describe the bug**
This happened after updating from 0.8.3 to 0.8.4. When I switch between two profiles (each is a different repo but every other setting is the same), the archive list does not update accordingly. I can't pick an archive from my second profile. The name on top of the list doesn't change as well. It seems like after updating my two profiles are set to one repository instead of each one having their own repo.

**To Reproduce**
1. Start vorta
2. Go to the archive tab
3. Switch profile

- OS: Arch Linux
- Vorta version: 0.8.4
- Installed from: AUR

Vorta and Borg versions can be found in Main Window > Misc Tab.

**Additional context**
Edit: here are the log entries when switching profiles.
```
2022-05-17 15:13:37,674 - vorta.views.source_tab - DEBUG - Added item number 0 from 1
2022-05-17 15:13:37,675 - vorta.views.source_tab - DEBUG - Added item number 1 from 2
2022-05-17 15:13:39,083 - vorta.views.source_tab - DEBUG - Added item number 0 from 1
2022-05-17 15:13:39,084 - vorta.views.source_tab - DEBUG - Added item number 1 from 2
2022-05-17 15:13:40,822 - vorta.views.source_tab - DEBUG - Added item number 0 from 1
2022-05-17 15:13:40,823 - vorta.views.source_tab - DEBUG - Added item number 1 from 2
```

m3nu:
Fixes #1298

### DIFF
--- a/src/vorta/views/repo_tab.py
+++ b/src/vorta/views/repo_tab.py
@@ -220,7 +220,7 @@ class RepoTab(RepoBase, RepoUI, BackupProfileMixin):
         # window.rejected.connect(lambda: self.repoSelector.setCurrentIndex(0))
         window.open()
 
-    def repo_select_action(self, index):
+    def repo_select_action(self):
         profile = self.profile()
         profile.repo = self.repoSelector.currentData()
         profile.save()
--- a/src/vorta/views/repo_tab.py
+++ b/src/vorta/views/repo_tab.py
@@ -221,9 +221,6 @@ class RepoTab(RepoBase, RepoUI, BackupProfileMixin):
         window.open()
 
     def repo_select_action(self, index):
-        if index == 0:
-            return
-
         profile = self.profile()
         profile.repo = self.repoSelector.currentData()
         profile.save()


src/vorta/views/repo_tab.py
```diff
@@ -220,10 +220,7 @@ def add_existing_repo(self):
         # window.rejected.connect(lambda: self.repoSelector.setCurrentIndex(0))
         window.open()
 
-    def repo_select_action(self, index):
-        if index == 0:
-            return
-
+    def repo_select_action(self):
```
real-yfprojects, line 223:
```suggestion
def repo_select_action(self, index: int):
```

The method is connected to the `currentIndexChanged(int index)` signal of the QComboBox `repoSelector`.

real-yfprojects (changes requested):
Additionally the following isn't up-to-date as well:

https://github.com/m3nu/vorta/blob/20d723d13c144feac027e59bc541f134b8a0a54f/src/vorta/views/repo_tab.py#L268-L271

src/vorta/views/repo_tab.py
real-yfprojects: For some reason it still works and doesn't produce an error. :shrug:

m3nu: Probably Qt function overloading. I tested it of course.

PR #1309, opened 2022-05-21
Determine source files for `translations-from-source` dynamically.

real-yfprojects:
Makefile (translations-from-source): Use a call to `find` to get a list
	of source files (*.ui and *.py) that is passed to `pylupdate5`.


m3nu (approved):
Awesome! 👍

PR #1311, opened 2022-05-22
Fix macOS input form size

m3nu:
Addendum to #1176. Fixes input field width on macOS. Missed this when fixing the same for the scheduler tab.

Before:
Screen Shot 2022-05-22 at 09 22 28

After:
Screen Shot 2022-05-22 at 09 53 30

### DIFF
--- a/src/vorta/assets/UI/archivetab.ui
+++ b/src/vorta/assets/UI/archivetab.ui
@@ -606,22 +606,6 @@
             <property name="verticalSpacing">
              <number>4</number>
             </property>
-            <item row="2" column="0">
-             <widget class="QLabel" name="label_3">
-              <property name="minimumSize">
-               <size>
-                <width>100</width>
-                <height>0</height>
-               </size>
-              </property>
-              <property name="text">
-               <string>Archive Name:</string>
-              </property>
-              <property name="alignment">
-               <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
-              </property>
-             </widget>
-            </item>
             <item row="0" column="1">
              <widget class="QLineEdit" name="prunePrefixTemplate">
               <property name="toolTip">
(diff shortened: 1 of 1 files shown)


real-yfprojects (changes requested):
Looks the same on KDE.
But you changed the design. Now `Archive Name` and `Prune Prefix` are grouped together. Previously the entry for the prune prefix was grouped in one frame with the other prune options and the archive name entry was placed below this frame.
Maybe this PR could also fix some other minor issues with the current layout. I noticed a big space between the preview labels and the entries. I think it would look better if the horizontal spacing was `4` in their QLayout. Additionally the font attribute of `archiveNamePrefix` has to be unset.

m3nu:
Will try to address the spacing.

You mean the pruning prefix template should be grouped with the other pruning options? Makes sense both ways to be honest. Pruning options together and template options together.

m3nu:
Changed the spacing.

Font is set to make it a bit smaller and grey.

real-yfprojects:
> Font is set to make it a bit smaller and grey.

The grey font colour is set in the `styleSheet` property. Currently the font size is too large. This might be because the default size differs from platform to platform. I therefore suggest unsetting the font size and setting the italic property to true instead.

real-yfprojects:
> You mean the pruning prefix template should be grouped with the other pruning options? 

Yes, I do.

>Makes sense both ways to be honest. Pruning options together and template options together.

At first glance it does. However, if the user wants to adjust pruning he will expect all relevant options (parameters for the prune action) to be (grouped together) at the same place, while design principles request options that do not belong together (parameters of different actions) have a larger space between them.

m3nu:
New grouping.
Screen Shot 2022-05-22 at 12 10 29

real-yfprojects (approved):
Great that this now works on MacOS too.

PR #1313, opened 2022-05-22
Fix copying the current repo to the clipboard.

real-yfprojects:
* src/vorta/views/repo_tab.py (RepoTab.copy_URL_action): Remove check of selected index. Construct a proper QMimeData with text and url field.


m3nu:
Right.. Same issue as with the repo selector. Thanks for catching!

m3nu (approved):

PR #1316, opened 2022-05-24
Remove call to `QTranslator.language()` in `get_locale`.

real-yfprojects:
The `language` method was added with Qt 5.15, which not all linux distributions use yet.

* src/vorta/i18n/__init__.py (get_locale): Return the global locale.

*That's a disadvantage of using a clean test system: You won't notice such issues.*

### DIFF
--- a/src/vorta/i18n/__init__.py
+++ b/src/vorta/i18n/__init__.py
@@ -71,8 +71,11 @@ def init_translations(app):
 
 def get_locale():
     """Get the locale used for translation."""
-    global translator
-    return QLocale(translator.language())
+    # global translator
+    # QTranslator.language was added with Qt 15.15
+    # return QLocale(translator.language())
+    global locale
+    return locale
 
 
 def translate(*args, **kwargs):
--- a/tests/test_schedule.py
+++ b/tests/test_schedule.py
@@ -68,5 +68,11 @@ def test_schedule_tab(qapp: VortaApp, qtbot, clockmock):
 
     qapp.scheduler.set_timer_for_profile(profile.id)
     tab.draw_next_scheduled_backup()
-    assert next_backup.strftime('%B %Y %H:%M:%S') in tab.nextBackupDateTimeLabel.text()
+
+    assert tab.nextBackupDateTimeLabel.text() not in [
+        "Run a manual backup first",
+        "None scheduled",
+    ]
+    assert qapp.scheduler.next_job_for_profile(profile.id).time == next_backup
+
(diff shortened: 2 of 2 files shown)


m3nu:
Tests fail due to a hardcoded date string:

```
        qapp.scheduler.set_timer_for_profile(profile.id)
        tab.draw_next_scheduled_backup()
>       assert next_backup.strftime('%B %Y %H:%M:%S') in tab.nextBackupDateTimeLabel.text()
E       AssertionError: assert 'May 2020 23:59:00' in 'Wednesday, May 6, 2020 11:59:00 PM UTC'
E        +  where 'May 2020 23:59:00' = <built-in method strftime of datetime.datetime object at 0x7f00da0df750>('%B %Y %H:%M:%S')
E        +    where <built-in method strftime of datetime.datetime object at 0x7f00da0df750> = datetime.datetime(2020, 5, 6, 23, 59).strftime
E        +  and   'Wednesday, May 6, 2020 11:59:00 PM UTC' = <built-in method text of QLabel object at 0x7f00c3191af0>()
E        +    where <built-in method text of QLabel object at 0x7f00c3191af0> = <PyQt5.QtWidgets.QLabel object at 0x7f00c3191af0>.text
E        +      where <PyQt5.QtWidgets.QLabel object at 0x7f00c3191af0> = <vorta.views.schedule_tab.ScheduleTab object at 0x7f00c31f1b80>.nextBackupDateTimeLabel
```

m3nu:
This became a really popular error. 😄

real-yfprojects:
> This became a really popular error. smile

Yeah, it is directly triggered by the scheduler. I don't understand why this many people are not able to write proper bug reports, formulate a descriptive title or find that the exact issue was already reported.

real-yfprojects:
Before a release i will now install the master branch using pip to ensure it is working on my machine too which I can confirm for the current one. Will we release the patch today/tomorrow?

m3nu:
Yes, soon. When you're happy with it.

m3nu:
Would we catch this by adding an older Ubuntu to our tests? Github Actions also has Ubuntu 18.04, which comes with PyQt 5.10.

https://packages.ubuntu.com/bionic/python3-pyqt5
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

real-yfprojects:
> Would we catch this by adding an older Ubuntu to our tests? Github Actions also has Ubuntu 18.04, which comes with PyQt 5.10.

The python dependencies like PyQt are installed by pip that uses the newest version.

PR #1368, opened 2022-06-27
Fix seconds for fixed schedule and fix scheduling for the same day

### LINKED ISSUE #1363: Shift of a few seconds with fixed schedule.
I am currently using the flatpak of Vorta 0.8.7 and have observed that the displayed time of my next scheduled backup varies from day to day at the seconds position. I would expect if I set a daily backup via the GUI to 12 o'clock, that the backup would then show 12:00:00 for the next day. Whereby the seconds could also be hidden.

real-yfprojects:
Previously the seconds of the current time or the last backup time were be copied. Fixes #1363.
This also fixes an issue when the scheduled time was missed and the backup is scheduled
for the current day although the time of day set has already passed.

* src/vorta/scheduler.py (VortaScheduler.set_timer_for_profile)

### DIFF
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -246,17 +246,19 @@ class VortaScheduler(QtCore.QObject):
 
             # calculate next scheduled time
             if profile.schedule_mode == 'interval':
-                last_time = last_run_log.end_time
+                last_time: dt = last_run_log.end_time
 
                 interval = {profile.schedule_interval_unit: profile.schedule_interval_count}
                 next_time = last_time + timedelta(**interval)
 
             elif profile.schedule_mode == 'fixed':
-                last_time = last_run_log.end_time + timedelta(days=1)
+                last_time = last_run_log.end_time
 
                 next_time = last_time.replace(
                     hour=profile.schedule_fixed_hour,
-                    minute=profile.schedule_fixed_minute)
+                    minute=profile.schedule_fixed_minute,
+                    second=0,
+                    microsecond=0) + timedelta(days=1)
 
             else:
                 # unknown schedule mode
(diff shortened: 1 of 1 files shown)


m3nu:
I didn't mind having random seconds. At BorgBase we get swamped by too many backups at exactly 0:00 UTC and again a bit less at 6:00 UTC. So some randomness isn't bad in such a system. So if you ever see a chance to put random minutes, etc as default somewhere, just do it. 😄

src/vorta/scheduler.py
```diff
@@ -246,17 +246,19 @@ def set_timer_for_profile(self, profile_id: int):
 
             # calculate next scheduled time
             if profile.schedule_mode == 'interval':
-                last_time = last_run_log.end_time
+                last_time: dt = last_run_log.end_time
```
m3nu, line 249: Is this a type hint? Didn't know one would put them inside functions.

real-yfprojects: It is. You can use them with every variable but often there is no need to in case of local variable since linters will be able to recognise the time on their own. In this case however the linter doesn't know the type of `last_run_log.end_time` since peewee is not typed.

real-yfprojects:
> At BorgBase we get swamped by too many backups at exactly 0:00 UTC and again a bit less at 6:00 UTC. So some randomness isn't bad in such a system. So if you ever see a chance to put random minutes, etc as default somewhere, just do it. 

Maybe we can hide the randomness from the user then. I will override this branch for that.

m3nu:
> Maybe we can hide the randomness from the user then. I will override this branch for that. Or should I create a new PR?

It's fine to merge this and run at a full minute. This was more meant as a comment/background.

m3nu (approved):

PR #1384, opened 2022-08-13
cleanup temp files at end of job

### LINKED ISSUE #1383: exclude pattern tmpfile is not cleaned up after the backup
I suppose it's the same for include pattern files. The file is created, written into, we do the backup, but it's never cleaned. So I end up with hundreds of tmp* files in my /tmp. It's not that big of an issue, but it's still a bit of a mess :)

Is a patch worth it ? I can write one for this if it's deemed useful (it would be for me). This happens with 0.8.7 on current archlinux.

marco44:
Closes #1383

This only has been tested locally, but seems to work. The file is only cleaned in case of a success

### DIFF
--- a/src/vorta/borg/create.py
+++ b/src/vorta/borg/create.py
@@ -145,11 +145,11 @@ class BorgCreateJob(BorgJob):
                     exclude_dirs.append(expanded_directory)
 
             if exclude_dirs:
-                pattern_file = tempfile.NamedTemporaryFile('w', delete=False)
+                pattern_file = tempfile.NamedTemporaryFile('w', delete=True)
                 pattern_file.write('\n'.join(exclude_dirs))
                 pattern_file.flush()
                 cmd.extend(['--exclude-from', pattern_file.name])
-                ret['cleanup_files'].append(pattern_file.name)
+                ret['cleanup_files'].append(pattern_file)
 
         if profile.exclude_if_present is not None:
             for f in profile.exclude_if_present.split('\n'):
--- a/src/vorta/borg/create.py
+++ b/src/vorta/borg/create.py
@@ -149,6 +149,7 @@ class BorgCreateJob(BorgJob):
                 pattern_file.write('\n'.join(exclude_dirs))
                 pattern_file.flush()
                 cmd.extend(['--exclude-from', pattern_file.name])
+                ret['cleanup_files'].append(pattern_file.name)
 
         if profile.exclude_if_present is not None:
             for f in profile.exclude_if_present.split('\n'):
--- a/src/vorta/borg/borg_job.py
+++ b/src/vorta/borg/borg_job.py
@@ -301,9 +301,8 @@ class BorgJob(JobInterface, BackupProfileMixin):
             self.process_result(result)
 
         self.finished_event(result)
-        for filename in self.cleanup_files:
-            if os.path.exists(filename):
-                os.remove(filename)
+        for tmpfile in self.cleanup_files:
+            tmpfile.close()
 
     def process_result(self, result):
         pass
--- a/src/vorta/borg/borg_job.py
+++ b/src/vorta/borg/borg_job.py
@@ -104,6 +104,7 @@ class BorgJob(JobInterface, BackupProfileMixin):
         self.cwd = params.get('cwd', None)
         self.params = params
         self.process = None
+        self.cleanup_files = params.get('cleanup_files', [])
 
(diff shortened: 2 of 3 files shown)


src/vorta/borg/create.py
```diff
@@ -149,6 +149,7 @@ def prepare(cls, profile):
                 pattern_file.write('\n'.join(exclude_dirs))
                 pattern_file.flush()
                 cmd.extend(['--exclude-from', pattern_file.name])
+                ret['cleanup_files'].append(pattern_file.name)
```
real-yfprojects, line 152:
```suggestion
                ret['pattern_temp_file'] = pattern_file
```

What about removing the `delete=False` when creating the temporary file and leaving it to the garbage collector? This will allow reusing a preparation object while removing the need for other changes.

real-yfprojects (commented):
I think this could be solved a little bit simpler - see below. Can you also implement the cleanup of the temporary file of the extract job?

src/vorta/borg/create.py
marco44: The idea was to be able to be more generic in case there are other temp files in the future. Or the extract job, which I didn't notice when looking at the code

Regarding delete=False, the file would be destroyed as soon as it is closed. Which I suppose we don't want, as this file is going to be passed later on to the borg command, right ?

real-yfprojects: > The idea was to be able to be more generic in case there are other temp files in the future. Or the extract job, which I didn't notice when looking at the code

That is a good call. In that case keep in mind that one might want to implement cleaning up directories in the future.

> Regarding delete=False, the file would be destroyed as soon as it is closed. Which I suppose we don't want, as this file is going to be passed later on to the borg command, right ?

I think it is only closed when the python object is garbage collected.

marco44: > That is a good call. In that case keep in mind that one might want to implement cleaning up directories in the future.
Yes, well, I suppose I could make it more complicated by having two arrays for temp_dirs, or just delete any name in the list, without distinction, which seems sufficient to me

> I think it is only closed when the python object is garbage collected.
Yes, so I suppose anytime after the end of the method, when pattern_file is out of scope, so it's going to be deleted, and we don't really know when, right ?

real-yfprojects: > Yes, well, I suppose I could make it more complicated by having two arrays for temp_dirs, or just delete any name in the list, without distinction, which seems sufficient to me

One array is sufficient since it can store any kind of path.


> Agreed, so I suppose anytime after the end of the method, when pattern_file is out of scope and the destructor is going to be called, so it's going to be deleted, and we don't really know when, right ?

Exactly. However, if we store the file object in the `ret` dictionary, it won't be destructed until the dictionary is.

marco44: > Exactly. However, if we store the file object in the ret dictionary, it won't be destructed until the dictionary is.

It feels really error prone, and hard to debug, just hoping it'll get destroyed when we want it, don't you think ? Really feels like using a side effect of us returning the name, and hoping the caller won't destroy ret too soon

real-yfprojects: I think python garbage collection works well enough. Though your approach is very valid since it allows all kinds of leftover files.

m3nu:
If it's all temp files, we want to clean up, this would work too and is less risky. The reference could go in your list. So only the deletion is slightly different.

```
# create a temporary file and write some data to it
>>> fp = tempfile.TemporaryFile()
...
# close the file, it will be removed
>>> fp.close()
```

https://docs.python.org/3/library/tempfile.html#examples

m3nu:
(Linting errors are fixed now. Thanks, @real-yfprojects)

marco44:
Ok, so I changed this to use the close method instead, you're right, it's slightly simpler.

If you're ok with this way of doing things, I'll implement it for extract too

m3nu:
Looks good. Anything that uses tmpfile would be best. Only create and extract from what I see. Thanks!

@real-yfprojects: Let's cherry-pick this into the borg-v2 branch soon. Will be harder to merge, if we really start using Black there.

marco44:
Should be ok now, extract is also done

real-yfprojects (approved):

m3nu (approved):

real-yfprojects:
Thanks for implementing this @marco44 and congrats to your first contribution! 🎉

PR #1391, opened 2022-08-15
Move linting to the top in test.yml

real-yfprojects:
This shows linting as first github action check for commits and PRs. Before you always had to scroll down past all the test for different python and os versions.


m3nu (approved):

m3nu:
Both fine with me.

Will move on to rebase `borg-v2` now and work on it when I can.

PR #1390, opened 2022-08-15
Format with Black and isort

m3nu:
```
black .
isort .
```

Fixes #1386

- [ ] Update contributor docs

Issue #1386, opened 2022-08-13
Unify code formatting and setup some kind of automation.


real-yfprojects:
- Choose a preferred code formatter (black or yapf)
- Run it on the whole code base.
- Think of and setup some system for PRs so that the code formatter is run on all commits/all code. Possible Solutions:
	- Linter only and committer has to run himself
   - formatter bot
   - github action
   - pre-commit script

m3nu:
Yes, please! I mostly use Black and [HN](https://news.ycombinator.com/item?id=17155048) also prefers it.

We can start using it in the v2 branch. Wanna add the config?

To keep it simple, I'd just raise a GH Actions linting error, if Black would detect a change (`black --check`).

real-yfprojects:
> We can start using it in the v2 branch. Wanna add the config?

I would merge #1384 first and then add a commit formatting the whole repository and another one adding the additional linting action. The `borg-v2-branch` as well as pending PRs can be rebased after that.

m3nu:
Fine with me. I don't mind redoing my v2 changes. Not much yet.

real-yfprojects:
> Fine with me. I don't mind redoing my v2 changes. Not much yet.

There shouldn't be much redoing. You rebase onto master interactively while editing the commit. You run black and isort on the modified files when editing. Or you rebase and then amend after running black 
and isort. Something along those lines should work (you can always reflog if it doesn't).

PR #1394, opened 2022-08-15
Pre commit setup

real-yfprojects:
This adds a [pre-commit](https://pre-commit.com/) configuration.

Developers can install and setup pre-commit:

```bash
> pip3 install pre-commit
> cd vorta
> pre-commit install
```

Now before committing the configured hooks will be run. If a check fails or a formatter makes changes to the code the commit is aborted. This ensures the developer is always in control of commit contents.

Pre-commit can be integrated into the CI throught the [pre-commit bot](https://github.com/marketplace/pre-commit-ci/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW42MTI2#pricing-and-setup). This makes sense since especially first-time contributors might not setup pre-commit.


real-yfprojects:
This is a follow up on #1390 and #1386.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+      # sort requirements.txt files
+      - id: requirements-txt-fixer
+
+  # format python files
+  - repo: https://github.com/psf/black
+    rev: 22.6.0
+    hooks:
+      - id: black
+        files: ^src/vorta/
```
m3nu, line 26: Shall we add formatting of tests here, as you pointed out? Also in Pyproject.toml. 🙏

m3nu:
Looks like a useful addition to auto-formatting. Thanks for adding!

real-yfprojects:
As described on [pre-commit.ci](https://pre-commit.ci/) the bot will automatically add commits to PRs that apply the fixes of the configured hooks by default. But there is also a configuration option to only apply the fixes when somebody comments `pre-commit.ci autofix`. There are also other configuration options available that have to be placed into the `.pre-commit-config.yaml` file. Therefore it makes sense to include the bot config in this PR.

m3nu:
No bot commits please. Whoever wishes, can apply those changes himself in his name.

real-yfprojects:
> No bot commits please. Whoever wishes, can apply those changes himself in his name.

Also no autoupdate PRs with custom title and description to update the version of the hooks?

Should I add a pre-commit check to the GA checks? This would replace the current `lint` check.

m3nu:
Problem is that if we set up too much magic and layers here, it will be hard to understand in the future. Now we have 3 tools that check the formatting. Those are run from GH actions. I can understand that in 6 or 12 months time.

If we add more layers now to run those tools, I will forget it in 6 months or new contributors will have to dig deep to find the actual process.

If the GA checks can run alongside the other tools that would be better and not add another level of tooling.

real-yfprojects:
Basically pre-commit allows defining the checks that have to pass on a new commit as well as automatic formatters in one place. Then it can be used to check commits locally before committing and in CI after committing. The CI bit can be done comfortable with pre-commit.ci but there is no need to use that bot instead we can also define a github action that suits our needs.

Then we can explain the three or four simple steps for setting up pre-commit in the docs and all contributors can use the same code quality tools like black, flake, ... almost out of the box. In the docs we can also lay out the workflow including the github actions and pre-commit, what they do and how the do it. In the result nobody will have to dig into the code for such common contributing affairs.

Basically pre-commit allows bundling code tools in one place that can then be accessed locally, from github actions, ... while docs and docstrings make the important information human and maintainer readable. Documentation consumes quite some time that doesn't enhance the program but it saves the maintainer and contributors a lot of time in the feature.

real-yfprojects:
> If the GA checks can run alongside the other tools that would be better and not add another level of tooling.

This is possible but it will slow down PR checks since pre-commit AND our existing lint GA run black and flake.

m3nu (approved):

m3nu:
Got it. This bundles all other code-checking tools. Like a Makefile, but before each commit.

Let's see how it goes and how many people use it. Then maybe run all formatting checks via pre-commit on GH actions. To avoid duplicate lists of checks.

PR #1430, opened 2022-10-11
Migrate to `--match-archives` in mount point detection.

real-yfprojects:
* src/vorta/utils.py (get_mount_points)


m3nu:
Considered the same change, but this would break it with older versions. But well, since it's a very minor feature, we'll survive.

PR #1433, opened 2022-10-15
Update `acions/checkout` to v3.

real-yfprojects:
The second version of the checkout action is deprecated and won't be supported by Github Actions in the future because it uses Node.js 12.


m3nu (approved):

m3nu:
Great! Thanks!

PR #1434, opened 2022-10-15
Borg v1.2.2 support

### LINKED ISSUE #1409: Borg v1.2.2 compatibility
There were some breaking changes although being a patch :shrug:

- [x] --prefix is deprecated, use -a / --glob-archives, see #6806
- [x] list: corrected mix-up of “isomtime” and “mtime” formats. Previously, “isomtime” was the default but produced a verbose human format, while “mtime” produced a ISO-8601-like format. The behaviours have been swapped (so “mtime” is human, “isomtime” is ISO-like), and the default is now “mtime”. “isomtime” is now a real ISO-8601 format (“T” between date and time, not a space).
- [x] exit with 128 + signal number, #5161. if you have scripts expecting rc == 2 for a signal exit, you need to update them to check for >= 128
- [x] matching of path patterns has been aligned with borg storing relative paths. Borg archives file paths without leading slashes. Previously, include/exclude patterns could contain leading slashes. You should check your patterns and remove leading slashes

Part of it might already be implemented for #1366.

Changelog: https://borgbackup.readthedocs.io/en/stable/changes.html#version-1-2-2-2022-08-20

real-yfprojects:
Closes #1409.

### DIFF
--- a/src/vorta/borg/_compatibility.py
+++ b/src/vorta/borg/_compatibility.py
@@ -5,7 +5,8 @@ MIN_BORG_FOR_FEATURE = {
     'ZSTD': parse_version('1.1.4'),
     'JSON_LOG': parse_version('1.1.0'),
     'DIFF_JSON_LINES': parse_version('1.1.16'),
-    'COMPACT_SUBCOMMAND': parse_version('1.2.0a1')
+    'COMPACT_SUBCOMMAND': parse_version('1.2.0a1'),
+    'V122': parse_version('1.2.2'),
     # add new version-checks here.
 }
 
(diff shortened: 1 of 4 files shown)


m3nu (dismissed):

m3nu:
Shouldn't this go into the current master branch, since it doesn't deal with Borg v2 specifically?

real-yfprojects:
> Shouldn't this go into the current master branch, since it doesn't deal with Borg v2 specifically?

This builds on the `borg-v2-support` branch and not the other way around because the `borg-v2-support` branch was developed earlier.

real-yfprojects:
But with `git rebase` we could change that of course.

m3nu (approved):

PR #1447, opened 2022-10-31
add AUR as install choice

RubenKelevra:


real-yfprojects (commented):
How does the proposed AUR package installation option defer from `Distribution Package`?

RubenKelevra:
Well, the AUR is not considered an official part of ArchLinux, so not part of the distribution ArchLinux.

It's just a platform where users of Arch and (and distributions based on it) can send build instructions for packages. The package itself is compiled/built locally on the system by Arch users. As this can introduce unique issues by the system on which it's build I would recommend having this as a unique source on the list, as it's closer to something like pip than distribution packages.

real-yfprojects:
After doing some research, I think AUR belongs in the `Other` category. It is like building a distribution package from source using unofficial instructions found somewhere on the internet.

PR #1466, opened 2022-11-03
Fix diff sorting

### LINKED ISSUE #1464: Diff view: Sorting by size doesn't work in tree mode.
When sorting the diff results by size in tree mode or in the simplified tree mode, the items aren't sorted correctly. It seems like they are sorted by name. In flat mode however, the sorting works fine.

real-yfprojects:
Fixes #1464.

* Remove legacy tree_view (tree model) implementation
* src/vorta/views/diff_result.py (DiffSortProxyModel.choose_data): Return correct data for column `2`.


m3nu (approved):

m3nu (approved):

m3nu:
Always happy to see something removed instead of added. 👏

PR #1473, opened 2022-11-05
Use new ssh://-style URL format as placeholder and in test

m3nu:


src/vorta/assets/UI/repoadd.ui
```diff
@@ -115,7 +115,7 @@
          <item>
           <widget class="QLineEdit" name="repoURL">
            <property name="placeholderText">
-            <string>csvis8xq@csvis8xq.repo.borgbase.com:repo</string>
+            <string>ssh://abc123@abc123.repo.borgbase.com/./repo</string>
```
real-yfprojects, line 118: The `./` is redundant, I think.

m3nu: Sadly not. Without it my URLs would look better.

```
$ borg init -e none ssh://h2w8590e@h2w8590e.repo.borgbase.com/repo
Remote: Warning: Permanently added 'h2w8590e.repo.borgbase.com' (ED25519) to the list of known hosts.
Repository path not allowed: /repo
```

Without it, the path is *absolute* instead of relative the the current user.

real-yfprojects (approved):

Issue #1214, opened 2022-03-06
Vorta is missing scheduled backup when using suspend


fixmeee:
I'm trying to use Vorta for backups on my laptop. In the evening, when I finished work the machine is put to suspend.
I setup an event to do an ACPI wake on the machine. This works fine. I can see the machine is woken up and I tested with some shell script it is really alive. 
But Vorta is missing the schedule after resume. If I just leave the machine powered, schedule is matched. But after resume it is missed. I guess the internal timer is somehow messed up due to suspend. Sadly I can't find anything in the logs. See below. As said the machine is defiantly powered up at 00:03:00.   

Thank you for your support.

1. Install Vorta via flatpak
2. Create a systemd auto resume service to wake up your machine before Vorta schedule shall take place.

Create /etc/systemd/system/auto-resume.timer:
[Unit]
Description=Automatically resume on a schedule

[Timer]
OnCalendar=*-*-* 03:00:00
WakeSystem=true

[Install]
WantedBy=timers.target

Create /etc/systemd/system/auto-resume.service:
[Unit]
Description=Does nothing

[Service]
Type=oneshot
ExecStart=/bin/true

Reload configuration: 
`systemctl daemon-reload`

Start the service:
`sudo systemctl start auto-resume.timer`

3. As the auto resume is triggered for 03:00:00 in the night, create a Vorta schedule at 03:05:00 (I tested also longer periods to give Vorta time to check actual time) 
4. You'll see backup is missed

**Environment (please complete the following information):**
- OS: cat /etc/lsb-release 
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=20.3
DISTRIB_CODENAME=una
DISTRIB_DESCRIPTION="Linux Mint 20.3 Una"

- Vorta version: 0.8.3
- Installed from: flatpak

**Log expert:
You can see the scheduled time is missed, but it had been catched up. The machine was powered and alive with network access since 00:03:00.

[...]
2022-03-05 22:17:22,464 - vorta.scheduler - INFO - Setting timer for profile 3
**2022-03-05 22:17:22,466 - vorta.scheduler - DEBUG - Scheduling next run for 2022-03-06 03:15:00**
2022-03-06 09:42:51,848 - vorta.scheduler - DEBUG - Refreshing all scheduler timers
2022-03-06 09:42:51,857 - vorta.scheduler - INFO - Setting timer for profile 1
[...]

PR #1511, opened 2022-12-12
Restart timers after hibernation/sleep

fixmeee:
Schedules get missed, if system has been in hibernation/sleep state as described in [#1214](https://github.com/borgbase/vorta/issues/1214) Although, timers should be refreshed every 15min, it seems that all timers get deranged by sleep state and schedules is still missed in most cases. This seems to be a general issue with the underlying QTimer implementation.

This fix doesn't distinguish host OS and has only by tested on Linux. If 'org.freedesktop.login1.Manager' is not available, Vorta simple doesn't get notified if system awakes again. So no impact on other OS is expected

### DIFF
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -76,8 +76,7 @@ class VortaScheduler(QtCore.QObject):
         if not data:
             now = dt.now()
 
-            current_time = now.strftime("%H:%M:%S")
-            logger.debug(f"Got login suspend/resume notification at %s", current_time)
+            logger.debug(f"Got login suspend/resume notification at {now.strftime("%H:%M:%S")}")
             self.reload_all_timers()
 
     def tr(self, *args, **kwargs):
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -59,20 +59,21 @@ class VortaScheduler(QtCore.QObject):
         # connect signals
         self.app.backup_finished_event.connect(lambda res: self.set_timer_for_profile(res['params']['profile_id']))
 
-        # connect to DBus login Manager to receive sleep/resume events
+        # connect to `systemd-logind` to receive sleep/resume events
+        # The signal `PrepareForSleep` will be emitted before and after hibernation.
         service = "org.freedesktop.login1"
+        path = "/org/freedesktop/login1"
+        interface = "org.freedesktop.login1.Manager"
+        name = "PrepareForSleep"
         if QtDBus.QDBusConnection.systemBus().interface().isServiceRegistered(service).value():
-            path = "/org/freedesktop/login1"
-            interface = "org.freedesktop.login1.Manager"
-            name = "PrepareForSleep"
             self.bus = QtDBus.QDBusConnection.systemBus()
             self.bus.connect(service, path, interface, name, "b", self._slotOverloaded)
         else:
             logger.warn('Failed to connect to DBUS interface to detect sleep/resume events')
 
     @QtCore.pyqtSlot(bool)
-    def _slotOverloaded(self, data):
-        if data is not True:
+    def _slotOverloaded(self, data: bool):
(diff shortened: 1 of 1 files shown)


m3nu:
Hi and thanks for the addition. Just some linting issues.

fixmeee:
Hi,

commit mail is fixed. The commits could be squashed anyway,

Thanks for taking care and BR
Dennis

src/vorta/scheduler.py
```diff
@@ -58,6 +59,26 @@ def __init__(self):
         # connect signals
         self.app.backup_finished_event.connect(lambda res: self.set_timer_for_profile(res['params']['profile_id']))
 
+        # connect to DBus login Manager to receive sleep/resume events
```
real-yfprojects, line 62:
```suggestion
        # connect to `systemd-logind` to receive sleep/resume events
        # The signal `PrepareForSleep` will be emitted before and after hibernation.
```
More detailed comment so that one doesn't have to look up the documentation of the interface.

real-yfprojects (changes requested):
Great work! Only some minor changes I suggest adding.

fixmeee:
I hope I caught everything. I did a few test and it seems to work.
For others who want to test, setup of suspend/resume with systemd is also described in [#1214](https://github.com/borgbase/vorta/issues/1214l)

src/vorta/scheduler.py
```diff
@@ -58,6 +59,27 @@ def __init__(self):
...
+        interface = "org.freedesktop.login1.Manager"
+        name = "PrepareForSleep"
+        if QtDBus.QDBusConnection.systemBus().interface().isServiceRegistered(service).value():
+            self.bus = QtDBus.QDBusConnection.systemBus()
+            self.bus.connect(service, path, interface, name, "b", self.loginSuspendNotify)
+        else:
+            logger.warn('Failed to connect to DBUS interface to detect sleep/resume events')
+
+    @QtCore.pyqtSlot(bool)
+    def loginSuspendNotify(self, data: bool):
```
real-yfprojects, line 75: The var name `data` isn't very descriptive either. Maybe somthing along the lines of `suspend` or `begin`.

src/vorta/scheduler.py
```diff
@@ -58,6 +59,27 @@ def __init__(self):
...
+        if not data:
+            now = dt.now()
+
+            current_time = now.strftime("%H:%M:%S")
+            logger.debug(f"Got login suspend/resume notification at %s", current_time)
```
real-yfprojects, line 80:
```suggestion
            logger.debug(f"Got login suspend/resume notification at {now.strftime("%H:%M:%S")}")
```
No idea whether this comes with any drawback but it seems cleaner.

real-yfprojects (changes requested):
Thanks for taking care of. Again there aren't any problems with your Implementation. However maintainability can be improved slightly.

src/vorta/scheduler.py
```diff
@@ -58,6 +59,26 @@ def __init__(self):
...
+            self.bus.connect(service, path, interface, name, "b", self.loginSuspendNotify)
+        else:
+            logger.warn('Failed to connect to DBUS interface to detect sleep/resume events')
+
+    @QtCore.pyqtSlot(bool)
+    def loginSuspendNotify(self, suspend: bool):
+        if not suspend:
+            now = dt.now()
+
+            logger.debug(f"Got login suspend/resume notification at {now.strftime('%H:%M:%S')}")
```
real-yfprojects, line 79: Just noticed that the logger will capture the time by itself. We don't have to do that.
```suggestion
            logger.debug(f"Got login suspend/resume notification")
```

real-yfprojects (approved):

m3nu:
LGTM. Thanks for resolving and reviewing!

m3nu (approved):

PR #1521, opened 2022-12-27
Add tooltips to settings.

### LINKED ISSUE #1517: The security option "Try to replace existing permissions when mounting an archive" should be explained somewhere
As a user, I have no idea whether I should check the box there or not.

real-yfprojects:
Fixes #1517

* src/vorta/store/models.py (SettingsModel): Add `tooltip` column.

* src/vorta/store/migrations.py (run_migrations): Create `tooltip` column.

* src/vorta/store/connection.py (init_db): Populate `tooltip` column. Increase `SCHEMA_VERSION`.

* src/vorta/views/misc_tab.py (MiscTab.populate): Set tooltip of checkbox widgets.

* src/vorta/store/settings.py : Add tooltips and update label of `override_mount_permissions`

### DIFF
--- a/src/vorta/assets/icons/help-about.svg
+++ b/src/vorta/assets/icons/help-about.svg
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg viewBox="0 0 64 64" version="1.1" id="svg6" sodipodi:docname="help-about.svg" width="64" height="64" inkscape:version="1.1.1 (3bf5ae0d25, 2021-09-20)" xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape" xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
+    <sodipodi:namedview id="namedview8" pagecolor="#ffffff" bordercolor="#666666" borderopacity="1.0" inkscape:pageshadow="2" inkscape:pageopacity="0.0" inkscape:pagecheckerboard="0" showgrid="false" inkscape:zoom="13.068182" inkscape:cx="52.37913" inkscape:cy="41.972174" inkscape:window-width="3840" inkscape:window-height="2107" inkscape:window-x="0" inkscape:window-y="0" inkscape:window-maximized="1" inkscape:current-layer="svg6" />
+    <defs id="defs3051">
+        <style type="text/css" id="current-color-scheme">
+      .ColorScheme-Text {
+      color:#000000;
+      }
+    </style>
+    </defs>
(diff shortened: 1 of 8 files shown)


real-yfprojects:
Any tooltips you would like to add or rephrase @m3nu?

m3nu:
Tooltip texts make sense. I just wonder if it makes more sense to add it below the checkbox label? Since almost nobody looks at tooltips really? Or have (?) icon that has the tooltip associated.

Screenshot 2022-12-27 at 11 18 51

real-yfprojects:
That is a good idea. However I do look for tooltips in many application. The [KDE HIG](https://develop.kde.org/hig/components/assistance/tooltip/) states:

> Tips eliminate the need to always have descriptive text on the screen. Novice users will use hovering to show tooltips to become familiar with the interface. This time-delay mechanism makes tips very convenient, but it also reduces their discoverability. When tips are used consistently they support user's expectation and foster predictability

However I do like the MacOS way of explaining a setting. I will look for a way to integrate this nicely into our app style.

real-yfprojects:
What about something like this?
grafik

m3nu:
Yesss! That's exactly what I was thinking of. That's what I see on macOS and on the web sometimes.

real-yfprojects:
What about something like this?
grafik

m3nu:
Still better than a normal tooltip. I use this on the web extensively. The (?) icon just shouldn't look like a button that can be clicked.

A secondary label would be good too, but clutters the space more.

real-yfprojects:
grafik

m3nu:
Usually I put the (?) in the same line instead of aligning them after. But no strong feelings about either. Aligning them per-group looks OK too, given the checkboxes on the other side.

help-icons

real-yfprojects:
> Usually I put the (?) in the same line instead of aligning them after. But no strong feelings about either.

You are right. I didn't notice that they were group aligned.

real-yfprojects:
Implemented the about buttons:
shot

m3nu:
Looks perfect! You'll have to show us how to use this tooltip elsewhere. I think it's real useful for explaining stuff behind complex options.

real-yfprojects:
> You'll have to show us how to use this tooltip elsewhere.

It can be used like a normal widget. To add it to an existing ui element, y
- wrap the ui element in a QHBoxLayout 
- add the ToolTipButton to it
- Set the tooltip of the ToolTipButton to the tooltip of the ui element

Example:
https://github.com/borgbase/vorta/blob/391ad2ad05419d15abc73db373ba5b1f3505ad01/src/vorta/views/misc_tab.py#L88-L94

m3nu:
One small problem: The SVG needs `<svg fill="#000000"` somewhere so changing the color works. And we have a function that updates icons when the color scheme changes. That's not used right now. I got this after minifying the file:

```
<svg xmlns="http://www.w3.org/2000/svg" width="64" height="64"><path d="M32 8.727A23.222 23.222 0 0 0 8.727 32 23.222 23.222 0 0 0 32 55.273 23.222 23.222 0 0 0 55.273 32 23.222 23.222 0 0 0 32 8.727Zm0 2.91A20.32 20.32 0 0 1 52.364 32 20.32 20.32 0 0 1 32 52.364 20.32 20.32 0 0 1 11.636 32 20.32 20.32 0 0 1 32 11.636zm-2.91 5.818v5.818h5.82v-5.818zm0 8.727v20.363h5.82V26.182Z" style="fill:currentColor;fill-opacity:1;stroke:none;stroke-width:2.90909;color:#000000"/></svg>
```

real-yfprojects:
> One small problem: The SVG needs `<svg fill="#000000"` somewhere so changing the color works.

The css attribute should be sufficient.

> And we have a function that updates icons when the color scheme changes. That's not used right now.

I did forget to implement the code for updating the icon when the colour scheme changes but I don't think we have such a function -- we only have `get_colored_icon` which I used.

m3nu:
> The css attribute should be sufficient.

Yes, the minified version works well. I think Vorta is replaceing `#000000`. So that needs to be somewhere. Before there was only `#000`, the short version.

There is some code that refreshes the icons when changing color. It's probably not critical for just this tab. Maybe there is a simple way.

real-yfprojects:
Icons now change on palette change.

m3nu (approved):
Looks good! This icon can be used in more places in the future. 👍

PR #609, opened 2020-09-03
Improve support for displaying Vorta's logo in Linux notifications

SergioRAgostinho:
1. Moved logo icon from packaging content to a formal asset of the application, since the notification image benefits all Linux users, regardless of the installation method;
2. Renamed the logo to clarify its content, better fitting its new location.

A follow up on #563. A description of the problem can be found in [here](https://github.com/borgbase/vorta/pull/563#issuecomment-686440404) and it affected non-flatpack users.

Screenshot from 2020-09-03 17-21-33

PR #1522, opened 2022-12-28
Notifications improvement.

real-yfprojects:
Closes #609.

I went a bit overboard and reimplemented the notifications module. It now supports the full `org.freedesktop.Notifications` feature set and uses the new MacOS `UserNotifications` API. The symbolic icon added in #563 is removed because notifications use the normal app icon afaik.

It would be a miracle if the MacOS code worked. Testing and fixing it is up to you.
The linux code works as expected:

Bildschirmfoto vom 2022-12-28 08-36-07
notification-screenshot


real-yfprojects:
> The linux code works as expected

I also checked XFCE. Works and looks very similar to GNOME.

real-yfprojects:
@m3nu Have you got time for this in the near future? Or should I exclude the MacOS changes from this PR?

m3nu:
Nice rewrite. Will work on it for macOS of course.

m3nu:
Fixed up macOS notifications to the point where they don't cause an error and open the main window when clicked.

sammcj:
This is definitely still needed - https://github.com/borgbase/vorta/issues/1773

real-yfprojects:
Yeah, the Linux part is all done, but the MacOS code needs to be tested and completed. I can't do that because I dont have MacOS.

PR #1549, opened 2023-01-18
Avoid autostart warning when removing item

### LINKED ISSUE #1508: Vorta automatically adding itself to Login Items in macOS
### Description

Everytime I start Vorta, it adds itself to the Login Items, ignoring that the option "Automatically start Vorta at login" is not checked. 

I run Vorta using a script that detects whenever my external media hard drive is mounted and then it runs Vorta. Since my computer takes some time to mount both the encrypted external disk and the remote disk where the backups go, running Vorta on login will cause an error since both the files to back up and the destination folders are not yet mounted and present. 

Both running the app with the script and doing so manually without script will add Vorta to the Login Items. I found nothing useful in the logs.

### Reproduction

### OS

macOS Ventura 13.0

### Version of Vorta

0.8.9

Binary

### Version of Borg

1.2.0

### Logs

_No response_

m3nu:
We get a warning about adding an autostart item, even if the feature isn't enabled. This is because we **always** add an autostart item and then remove it, if it's not needed. That's because it was easier to add that way.

This PR only adds an item if autostart is enabled and tries to remove all items called `Vorta.app`. So it would even work, if the app is moved or updated at some point.

Fixes #1508


m3nu:
Bonus change: Move linting and testing commands to Makefile, so they can be run manually without being duplicated.

real-yfprojects:
Neither can I test this nor am I  familiar with pyobjc.

m3nu:
Neither am I. But found a similar project. Works well from what I see. And the notification only comes when it's enabled.

PR #1578, opened 2023-02-02
Fully remove setup_requires, pin black 

abdeoliveira:
I'm far from being an expert, but it seems `pip` is not necessary in `setup_requires` as per this [suggestion](https://github.com/void-linux/void-packages/pull/41672#discussion_r1071176890).

Also, testing fails complaining about missing module `six`. Could you please check it out?

### DIFF
--- a/requirements.d/dev.txt
+++ b/requirements.d/dev.txt
@@ -1,4 +1,4 @@
-black
+black<23.0
 coverage
 flake8
 isort
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,8 +36,8 @@ package_dir =
     =src
 include_package_data = true
 python_requires = >=3.7
-setup_requires =
-  pip >= 10
+#setup_requires =
+#  pip >= 10
 install_requires =
   appdirs
   paramiko
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,8 +36,6 @@ package_dir =
     =src
 include_package_data = true
 python_requires = >=3.7
-#setup_requires =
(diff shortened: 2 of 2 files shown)


m3nu:
We can probably remove `pip`, but why add `six`? Isn't that for Python 2? We never supported that.

abdeoliveira:
Indeed, you are right. My bad about `six`. Fixed the commit.

requirements.d/dev.txt
```diff
@@ -1,4 +1,4 @@
-black
+black<23.0
```
real-yfprojects, line 1:
```suggestion
black==22.*
```

real-yfprojects:
We should confirm to [PEP 518](https://peps.python.org/pep-0518) instead as stated in the [setuptools docs](https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-setup-requires).
This require a section `buildsystem` in `pyproject.toml`.

m3nu:
So like this?

> For the vast majority of Python projects that rely upon setuptools, the pyproject.toml file will be:

```
[build-system]
requires = ["setuptools", "wheel"]  # PEP 508 specifications.
```

real-yfprojects:
The [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/quickstart.html) suggest:

```toml
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

```

m3nu:
Added in #1580 😁

My bad for merging too quickly.

PR #1586, opened 2023-02-07
[Enhancement] Add Setting For Running Vorta In The Background

### LINKED ISSUE #1582: Add setting for running vorta in the backround.
There is already a setting called `enable_background_question` that asks whether to continue in the background on exit when the system tray isn't available. Maybe this setting could be expanded to the case when a system tray is available.

### Discussed in https://github.com/borgbase/vorta/discussions/1581

<div type='discussions-op-text'>

<sup>Originally posted by **DaffyTheDuck** February  5, 2023</sup>
When you run the `vorta` command from the terminal to launch new session and close the application from GUI instead of closing it from the terminal, the command `vorta` keeps running though the application is closed

image

## How to reproduce ?
1. Launch vorta from terminal
2. Close it through GUI
3. See the terminal

## Desktop
Ubuntu 22.04

## Update!
I actually just found that it keeps running in background! But while closing the application it would be nice to prompt user whether he want's to continue the process or not 😄 </div>

DaffyTheDuck:
- Added setting to run vorta in background
- Added option in the mics tab to enable/disable the background question popup.

Closes #1582

### DIFF
--- a/src/vorta/store/settings.py
+++ b/src/vorta/store/settings.py
@@ -32,7 +32,7 @@ def get_misc_settings() -> List[Dict[str, str]]:
             'value': False,
             'type': 'checkbox',
             'group': notifications,
-            'label': trans_late('settings', 'Also notify about successful background tasks'),
+            'label': trans_late('settings', 'Notify about successful background tasks'),
         },
         {
             'key': 'autostart',
@@ -40,6 +40,7 @@ def get_misc_settings() -> List[Dict[str, str]]:
             'type': 'checkbox',
             'group': startup,
             'label': trans_late('settings', 'Automatically start Vorta at login'),
+            'tooltip': trans_late('settings', 'Add Vorta to the systems autostart list'),
         },
         {
             'key': 'foreground',
@@ -47,6 +48,7 @@ def get_misc_settings() -> List[Dict[str, str]]:
             'type': 'checkbox',
             'group': startup,
             'label': trans_late('settings', 'Open main window on startup'),
+            'tooltip': trans_late('settings', 'Open main window when the application is launched'),
         },
         {
             'key': 'get_srcpath_datasize',
@@ -54,6 +56,7 @@ def get_misc_settings() -> List[Dict[str, str]]:
             'type': 'checkbox',
             'group': information,
             'label': trans_late('settings', 'Get statistics of file/folder when added'),
+            'tooltip': trans_late('settings', 'When adding a new source, calculate its size and the number of files.'),
         },
         {
             'key': 'use_system_keyring',
@@ -64,6 +67,9 @@ def get_misc_settings() -> List[Dict[str, str]]:
                 'settings',
                 'Store repository passwords in system keychain, if available',
             ),
+            'tooltip': trans_late(
+                'settings', "Otherwise Vorta's configuration database stores the password in plaintext."
+            ),
         },
         {
             'key': 'override_mount_permissions',
@@ -72,8 +78,9 @@ def get_misc_settings() -> List[Dict[str, str]]:
             'group': security,
             'label': trans_late(
                 'settings',
-                'Try to replace existing permissions when mounting an archive',
+                'Try to replace file permissions when mounting an archive',
             ),
+            'tooltip': trans_late('settings', 'Set owner to current user and umask to 0277'),
         },
         {
             'key': 'previous_profile_id',
--- a/src/vorta/assets/icons/help-about.svg
+++ b/src/vorta/assets/icons/help-about.svg
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg viewBox="0 0 64 64" version="1.1" id="svg6" sodipodi:docname="help-about.svg" width="64" height="64" inkscape:version="1.1.1 (3bf5ae0d25, 2021-09-20)" xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape" xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
+    <sodipodi:namedview id="namedview8" pagecolor="#ffffff" bordercolor="#666666" borderopacity="1.0" inkscape:pageshadow="2" inkscape:pageopacity="0.0" inkscape:pagecheckerboard="0" showgrid="false" inkscape:zoom="13.068182" inkscape:cx="52.37913" inkscape:cy="41.972174" inkscape:window-width="3840" inkscape:window-height="2107" inkscape:window-x="0" inkscape:window-y="0" inkscape:window-maximized="1" inkscape:current-layer="svg6" />
+    <defs id="defs3051">
+        <style type="text/css" id="current-color-scheme">
+      .ColorScheme-Text {
+      color:#000000;
+      }
+    </style>
+    </defs>
+    <path style="fill:currentColor;fill-opacity:1;stroke:none;stroke-width:2.90909" d="m 31.999999,8.7272725 c -12.893091,0 -23.2727265,10.3796355 -23.2727265,23.2727265 0,12.893091 10.3796355,23.272729 23.2727265,23.272729 C 44.89309,55.272728 55.272728,44.89309 55.272728,32 55.272727,19.106908 44.89309,8.7272725 31.999999,8.7272725 Z m 0,2.9090905 c 11.281455,0 20.363637,9.082182 20.363637,20.363636 0,11.281455 -9.082182,20.363637 -20.363637,20.363637 -11.281454,0 -20.363636,-9.082182 -20.363636,-20.363637 0,-11.281454 9.082182,-20.363636 20.363636,-20.363636 z m -2.909091,5.818182 v 5.818182 h 5.818182 v -5.818182 z m 0,8.727272 V 46.545454 H 34.90909 V 26.181817 Z" class="ColorScheme-Text" id="path4" />
+</svg>
--- a/src/vorta/store/connection.py
+++ b/src/vorta/store/connection.py
@@ -18,7 +18,7 @@ from .models import (
 )
 from .settings import get_misc_settings
 
-SCHEMA_VERSION = 19
+SCHEMA_VERSION = 20
 
 
 @signals.post_save(sender=SettingsModel)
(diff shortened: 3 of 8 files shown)


DaffyTheDuck:
Hi! I know that the code is just repeated when the system tray is available!, i'm trying to figure out what changes I can do so the repetition is prevented 

would love to hear about changes/improvements 😄

m3nu:
Why not just remove the `if` check for the system tray [here](https://github.com/borgbase/vorta/pull/1586/files#diff-5fc41e635bb90172ba4736314836f73b9488a0f3e590644a292c86e7f02c1164R280) and show it all the time, unless the setting to hide it is there?

real-yfprojects:
:point_up: In your implementation the top-level if-statement in line 280 contains the same code in its if-clause and the corresponding else-clause. The if-statement in line 280 can therefore be dropped.

https://github.com/borgbase/vorta/blob/fee18b37b8f68051860859ce51c96b06e4df4b39/src/vorta/views/main_window.py#L280

DaffyTheDuck:
Hi! I know this might be silly, but it works as expected, please review 😄 

also it's weird that I've amended the commit but it's still showing 2 commits to merge @real-yfprojects @m3nu

src/vorta/views/main_window.py
```diff
@@ -277,7 +277,7 @@ def closeEvent(self, event):
             SettingsModel.key == 'previous_window_height'
         ).execute()
 
-        if not is_system_tray_available():
+        if not is_system_tray_available() or is_system_tray_available():
```
real-yfprojects, line 280: You are progressing into the right direction. You now managed to avoid the redundancy. However, given a boolean variable `a` the statement `a or not a` is always true. The if-statement in this line can therefore be dropped.

DaffyTheDuck: Hmm! Insightful

DaffyTheDuck: Removed The If Check 😄

real-yfprojects:
Great! This is a good solution and works as expected. One small problem: The setting for showing/not showing the exit dialog is not displayed in the GUI.

DaffyTheDuck:
> Great! This is a good solution and works as expected. One thing though: The setting for showing/not showing the exit dialog is not displayed in the GUI.

Hi! I didn't get that actually 😓 

can you please elaborate ?

do you mean the dialogue which is displayed when the close button is clicked ?

real-yfprojects:
Vorta has some settings which you can change in the *Misc* tab. However there is no setting for showing or not showing the exit dialog. This should be changed.

DaffyTheDuck:
> Vorta has some settings which you can change in the _Misc_ tab. However there is no setting for showing or not showing the exit dialog. This should be changed.

Hi! what if we show the run in background checkbox instead of giving an option to enable the popup message again ?

src/vorta/store/settings.py
```diff
@@ -48,6 +48,17 @@ def get_misc_settings() -> List[Dict[str, str]]:
...
         },
+        {
+            'key': 'run_in_background',
+            'value': False,
+            'type': 'checkbox',
+            'group': startup,
+            'label': trans_late(
+                'settings',
+                'Run vorta in background',
+            ),
```
real-yfprojects, line 59: How does this new setting work differently from `enable_background_question`? Can't you adjust that setting to work as needed?

DaffyTheDuck: Hi! I'm a bit confused here! I actually was unable to see the checkbox for the `enable_background_question` in mics tab, that's why I added a different one 😓, but now it's visible for me!

real-yfprojects: >  I actually was unable to see the checkbox

I noticed that too. Settings without a group aren't displayed at the moment (which is a bug). But you can easily work around that by adding the group attribute.

> but now it's visible for me

That's weird.

DaffyTheDuck: > > I actually was unable to see the checkbox
> 
> I noticed that too. Settings without a group aren't displayed at the moment (which is a bug). But you can easily work around that by adding the group attribute.
> 
> > but now it's visible for me
> 
> That's weird.

Hmm did and it displayed the setting, but if I restore the setting to previous state i.e without the group attribute it still displays the setting checkbox and it works as expected without any changes

DaffyTheDuck: weird

Here's, the example


Code from `settings.py`

```
settings += [
            {
                'key': 'enable_background_question',
                'value': True,
                'type': 'checkbox',
                'label': trans_late(
                    'settings',
                    "Toggle run in background dialogue",
                ),
            },
            {
                'key': 'disable_background_state',
                'value': False,
                'type': 'internal',
                'label': 'Previous background exit button state',
            },
        ]
```
back to original version

real-yfprojects: Ah, This is because you local settings database still has the group stored inside it. Vorta doesn't remove old settings keys.

DaffyTheDuck: > Ah, This is because you local settings database still has the group stored inside it. Vorta doesn't remove old settings keys.

um, is there any way to remove it ? 😅

real-yfprojects: You can edit the database located in `~/.local/share/Vorta/settings.db`. I use [sqlitebrowser](https://sqlitebrowser.org/) for that.

DaffyTheDuck: > You can edit the database located in `~/.local/share/Vorta/settings.db`. I use [sqlitebrowser](https://sqlitebrowser.org/) for that.

Thank You 😄

DaffyTheDuck:
Hi! I was curious about the pytest, is there any documentation I can refer to? and I also added the group to the setting to make it visible, but is that correct, it works as expected 😄

real-yfprojects:
Is _startup_ a fitting group? I think something like _background_ would be better.

DaffyTheDuck:
> Is _startup_ a fitting group? I think something like _background_ would be better.

Done 😄

Hofer-Julian:
> > Is _startup_ a fitting group? I think something like _background_ would be better.
> 
> Done smile
> 
> But I Still have a questions! how github is showing 2 commits even though I amended the commit every time thinking

It seems like there are indeed two commits.
If you want to combine them into one, you can run:
1. `git reset --soft HEAD~2`
2. `git commit`
3. `git push --force-with-lease`

DaffyTheDuck:
> > > Is _startup_ a fitting group? I think something like _background_ would be better.
> > 
> > 
> > Done smile
> > But I Still have a questions! how github is showing 2 commits even though I amended the commit every time thinking
> 
> It seems like there are indeed two commits. If you want to combine them into one, you can run:
> 
>     1. `git reset --soft HEAD~2`
> 
>     2. `git commit`
> 
>     3. `git push --force-with-lease`

Hi! Pushed it but, is that okay like merging the changes which are merged already ?

src/vorta/store/settings.py
```diff
@@ -115,9 +116,10 @@ def get_misc_settings() -> List[Dict[str, str]]:
                 'key': 'enable_background_question',
                 'value': True,
                 'type': 'checkbox',
+                'group': background,
                 'label': trans_late(
                     'settings',
-                    "If the system tray isn't available, " "ask whether to continue in the background " "on exit",
+                    "Toggle run in background dialogue",
```
real-yfprojects, line 122:
```suggestion
                    "When closing the window, ask whether to continue in the background",
```

DaffyTheDuck: 👍🏼

real-yfprojects (approved):

real-yfprojects (changes requested):
Sorry, that this PR takes so long to merge. There are two issues I found: 

- [x] The checkbox in the Misc tab isn't updated when `Do not show again` is checked in the dialog.
- [x] The dialog is also shown when the tray icon is clicked for hiding the window. This is a minor issue however.

DaffyTheDuck:
> Sorry, that this PR takes so long to merge. There are two issues I found: 
> 
> - [ ] The checkbox in the Misc tab isn't updated when `Do not show again` is checked in the dialog.
> - [ ] The dialog is also shown when the tray icon is clicked for hiding the window. This is a minor issue however.

Hi! For the first one, what is the exact expected behaviour is ? Like if I check the don't show again box and then exit, after re-opening the mics tab the Background check box should be unchecked by default?

real-yfprojects:
> Like if I check the don't show again box and then exit, after re-opening the mics tab the Background check box should be unchecked by default?

Yes, the dialog saves the setting correctly, however the checkbox is only updated when quitting and restarting vorta.

DaffyTheDuck:
> > Like if I check the don't show again box and then exit, after re-opening the mics tab the Background check box should be unchecked by default?
> 
> Yes, the dialog saves the setting correctly, however the checkbox is only updated when quitting and restarting vorta.

Hi! Do I need to update the `Run In Background` checkbox as soon as the `Do not show again` checkbox is clicked ? because the since in both cases i.e choosing yes or no, the application closes and when you open the application again the checkbox is updated 🤔 

And actually I've confusion about the second, how exactly I can reproduce it ?

real-yfprojects:
> because the since in both cases i.e choosing yes or no, the application closes and when you open the application again the checkbox is updated

No, when I activate `Do not show again` and hit `Yes` the checkbox isn't updated, when opening vorta again (e.g. by clicking on the icon in the status bar).

> Do I need to update the Run In Background checkbox as soon as the Do not show again checkbox is clicked ?

Yes. You could adjust the `clicked` handler or add a listener for database changes that then updates the checkboxes similar to this one:

https://github.com/borgbase/vorta/blob/b36fa481855d4431547afea65e91882d41afe247/src/vorta/store/connection.py#L24-L27

DaffyTheDuck:
> 

Hi! I'm extremely sorry that this is taking too long from my side my mid sem exam is on 😄 


by the `clicked` handler do you mean this one ?

`dont_show_box.clicked.connect(lambda x: self.miscTab.save_setting("enable_background_question", not x))`

but as per my understanding, the lambda function should update it when the since it is receiving either True or False from `clicked.connect`. Am I assuming something wrong here ? I would love to get more knowledge about it 😄

real-yfprojects:
> I'm extremely sorry that this is taking too long from my side my mid sem exam is on smile

No worries. This doesn't have an impact on us.



> by the `clicked` handler do you mean this one ?
> 
> ```py
> dont_show_box.clicked.connect(lambda x: self.miscTab.save_setting("enable_background_question", not x))
> ```

Yes, I  do.

> but as per my understanding, the lambda function should update it, since it is receiving either True or False from `clicked.connect`. Am I assuming something wrong here ? I would love to get more knowledge about it smile


Yes, the setting (stored in the database) is correctly updated. However you will find that the checkbox widget doesn't update when the value in the database changes. The checkbox only updates when the vorta process/program is restarted.

PR #1587, opened 2023-02-07
Disable archive action buttons when running backup.

### LINKED ISSUE #1360: Compact button in Archive tab isn't disabled during backup
**Describe the bug**
*Compact* button should be disabled, when another Borg process is running.
Screen Shot 2022-06-18 at 09 41 59

**To Reproduce**
1. Go to *Archive* tab
2. Start any Borg action
3. See that all buttons, except *Compact* are disabled.

- OS: macOS
- Vorta version: 0.8.7
- Installed from: Homebrew

### LINKED ISSUE #1563: Buttons (Archives Tab) are not deactivated while creating a backup
### Description

The Buttons Refresh, Mount, Extract, Rename, Diff and Delete are not greyed out as unavailable when running a backup job. In addition, Compact button is also still active, which is already reported (see #1360).

Screenshot from 2023-01-21 22-48-11

### Reproduction

### OS

ArchLinux latest

### Version of Vorta

0.8.9

Other

### Version of Borg

1.2.3

### Logs

_No response_

sunny775:
Resolves #1563, resolves #1360,

### DIFF
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -222,6 +222,12 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
             self.bDiff,
             self.fArchiveActions,
             self.bMountRepo,
+            self.bRefreshArchive,
+            self.bMountArchive,
+            self.bExtract,
+            self.bRename,
+            self.bDelete,
+            self.compactButton,
         ]:
             button.setEnabled(enabled)
             button.repaint()


m3nu:
Passes tests, does what it promises and closes 2 issues. 5/5

real-yfprojects (changes requested):
At first glance, this is a very elegant solution. At second glance, it is elegant still. However it doesn't work correctly in all cases. 

The `bDelete` and `bDiff` buttons aren't disabled when I run a backup.

Secondly the archive actions which are grouped in the `fArchiveActions` frame are disabled twice. When `fArchiveActions` is disabled, they are automatically too. Then you disable each of them again. This double-disabling is the reason your solution works for these buttons, but not for the other two. The other two buttons aren't disabled twice. And they shouldn't -- in my opinion that would be a hackish solution.

sunny775:
> 

Okay, noted. Thanks @real-yfprojects!

sunny775:
@real-yfprojects Disabling them individually works for other buttons, but the `bDelete` and `bDiff` buttons never get disabled.
I've tried everything I can think of, any idea on what the possible fix could be?

real-yfprojects:
> any idea on what the possible fix could be?

Solved it in sunny775/vorta#1

m3nu:
Is this ready to merge?

real-yfprojects:
I think so

m3nu (approved):

PR #1588, opened 2023-02-08
Allow creating an SSH key when bootstrapping Vorta

### LINKED ISSUE #1579: Can't create SSH key without existing remote repo
Hello

I had confirmed the same problem as other members. But I don't see a resolution?

I confirm that I can't follow the tutorial ~10mn https://www.youtube.com/watch?v=asZX2YbTaNE
If I watch a video ~ 10' I can't select anything
Or https://norangeb.it/en/posts/2022/backup-with-borg-borgbase-and-vorta/
 
On the screenshot, it can select the creation of a new key
I see this also on other tutorials.

**Me, everything is greyed out, not selectable.**

I just started from scratch by installing vorta Flatpak 0.8.10 on ubuntu 22.10.
How do I do this?

vorta

tobast:
Fixes #1579

As discussed in the issue, I am not sure that this fix is the most relevant — it only makes the "+" button for SSH keys available at all times, in order to be able to bootstrap a remote backup without having to run `ssh-keygen` manually (eg. for non-tech users). A quickstart wizard, or even a similar "+" option in the repository creation dialog itself, would be better suited, I believe.

### DIFF
--- a/src/vorta/views/repo_tab.py
+++ b/src/vorta/views/repo_tab.py
@@ -2,7 +2,7 @@ import os
 from pathlib import PurePath
 from PyQt5 import QtCore, uic
 from PyQt5.QtCore import QMimeData, QUrl
-from PyQt5.QtWidgets import QApplication, QMenu, QMessageBox
+from PyQt5.QtWidgets import QApplication, QLayout, QMenu, QMessageBox
 from vorta.store.models import ArchiveModel, BackupProfileMixin, RepoModel
 from vorta.utils import borg_compat, get_asset, get_private_keys, pretty_bytes
 from .repo_add_dialog import AddRepoWindow, ExistingRepoWindow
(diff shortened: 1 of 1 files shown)


src/vorta/views/repo_tab.py
```diff
@@ -147,7 +147,12 @@ def init_repo_stats(self):
             self.repoEncryption.setText(str(repo.encryption))
         else:
             # Compression and SSH key are only valid entries for a repo
-            self.frameRepoSettings.setEnabled(False)
+            # Yet Add SSH key button must be enabled for bootstrapping
+            self.frameRepoSettings.setEnabled(True)
```
real-yfprojects, line 151: unnecessary line (l. 114 likewise)

src/vorta/views/repo_tab.py
```diff
@@ -147,7 +147,12 @@ def init_repo_stats(self):
...
+            for child in self.frameRepoSettings.children():
+                if not isinstance(child, QLayout):
+                    child.setEnabled(False)
```
real-yfprojects, line 154: The children are disable, but not enabled again if the repo is set.

src/vorta/views/repo_tab.py
tobast: Indeed, both lines removed

src/vorta/views/repo_tab.py
tobast: Indeed. Fixed in a new revision.

real-yfprojects (approved):
Thank you for your work!

m3nu:
Works as advertised.

m3nu (approved):

HDInfautre:
Hi
`As discussed in the issue, I am not sure that this fix is the most relevant — it only makes the "+" button for SSH keys available at all times, in order to be able to bootstrap a remote backup without having to run ssh-keygen manually (eg. for non-tech users). A quickstart wizard, or even a similar "+" option in the repository creation dialog itself, would be better suited, I believe.`
Just a user, I share this opinion, but as I am not a coder, maybe you need more shared opinions?
In any case, thank you
Subsidiary question: what timeframe do you think you will release a new release: week? month?

PR #1592, opened 2023-02-11
Add Windows support

RayyanAnsari:


m3nu:
Good stuff! I'll try to test this on a VM, but the changes look sensible.

Could we add Windows as testing OS on Github Actions? And maybe build a binary there too?

m3nu:
Minor linting error about ordered imports. And one import will need a condition to check the OS.

RayyanAnsari:
I'm not sure how to approach the fix for the temporary files, see my last commit 15f214d.

> Could we add Windows as testing OS on Github Actions? And maybe build a binary there too?

There are no Borg releases yet with Windows builds, so we would have to fetch a build off the CI. This PR is still a draft, because I haven't thoroughly tested it, and I know for a fact that remote repositories do not yet work in Borg.

> Minor linting error about ordered imports. And one import will need a condition to check the OS.

Trivial fix, will do.

Issue #1595, opened 2023-02-17
FR: the Size column in the archives list is hard to read due to the mixed units. My proposed change (pull request incoming, need help if possible)


hariseldon78:
I find hard to read the sizes of the archives at a glance when they are all mixed units: usually the longer numbers represent smaller sizes, making it difficult to find the biggest archives. So I made a basic algorithm to select a good unit that can adequately represent all sizes, that makes it much easier to read. Also I made the column align to the right, which further improves readability.

This is a screenshot of my proposed change
![image](https://user-images.githubusercontent.com/133625/219548261-26dbe769-e511-4861-8290-52f3f4eac366.png)

hariseldon78:
sorry, the pull request is not ready yet, the tests are failing and I can't pinpoint the cause.. if somebody can help me here's my fork:

real-yfprojects:
Just open a pull request, possibly as a draft. I can have a look then.

hariseldon78:
> Just open a pull request, possibly as a draft. I can have a look then.

ok, here it is: https://github.com/borgbase/vorta/pull/1598

PR #1598, opened 2023-02-17
improve size column readability in archives tab

hariseldon78:
I find hard to read the sizes of the archives at a glance when they are all mixed units: usually the longer numbers represent smaller sizes, making it difficult to find the biggest archives. So I made a basic algorithm to select a good unit that can adequately represent all sizes, that makes it much easier to read. Also I made the column align to the right, which further improves readability.

### Related Issue
https://github.com/borgbase/vorta/issues/1595

### Motivation and Context
Readability

### How Has This Been Tested?
Tested on my archive list, which sizes goes from about 100 MB to 150 GB

image

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -227,7 +227,7 @@ def get_private_keys():
 def sort_sizes(size_list):
     """Sorts sizes with extensions. Assumes that size is already in largest unit possible"""
     final_list = []
-    for suffix in [" B", " KB", " MB", " GB", " TB"]:
+    for suffix in [" B", " KB", " MB", " GB", " TB", " PB", " EB", " ZB", " YB"]:
         sub_list = [
             float(size[: -len(suffix)])
             for size in size_list
@@ -288,7 +288,7 @@ def pretty_bytes(size: int, metric: bool = True, sign: bool = False, precision: 
         n = find_best_unit_for_size(size, metric=metric, precision=precision)
     else:
         n = fixed_unit
-    n = clamp(n, 0, len(units))
+    n = clamp(n, 0, len(units) - 1)
     size /= power**n
     try:
         unit = units[n]
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -2,6 +2,7 @@ import argparse
 import errno
 import fnmatch
 import getpass
+import math
 import os
 import platform
 import re
@@ -240,22 +241,25 @@ def sort_sizes(size_list):
     return final_list
 
 
-def find_best_size_formatting(size_list, metric=True):
-    power, units = (
-        (10**3, ['', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'])
-        if metric
-        else (2**10, ['', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'])
-    )
-    min_size = min(s for s in size_list if isinstance(s, int))
-    # we tolerate a maximum precision of 1
-    n = 0
-    while abs(min_size / power**n > 0.1) and n + 1 < len(units):
-        n += 1
-    n = max(n - 1, 0)
-    return 1, units[n]
+def find_best_size_unit(sizes: Iterable[int], metric: bool = True, precision: int = 1) -> int:
+    """
+    Selects the index of the biggest unit (see the lists in the pretty_bytes function) capable of
+    representing the smallest size in the sizes iterable.
+    """
+    power = 10**3 if metric else 2**10
+    min_size = min((s for s in sizes if isinstance(s, int)), default=None)
+    if min_size is None:
+        return 0
+    n = math.floor(math.log(min_size * 10**precision, power))
+    return n
 
 
-def pretty_bytes(size, metric=True, sign=False, precision=1, fixed_unit=None):
+def pretty_bytes(size: int, metric: bool = True, sign: bool = False, precision: int = 1, fixed_unit: int = None) -> str:
+    """
+    Formats the size with the requested unit and precision. The find_best_size_unit function
+    can be used to find the correct unit for a list of sizes. If no fixed_unit is passed it will
+    find the biggest unit to represent the size
+    """
     if not isinstance(size, int):
         return ''
     prefix = '+' if sign and size > 0 else ''
@@ -270,7 +274,7 @@ def pretty_bytes(size, metric=True, sign=False, precision=1, fixed_unit=None):
             size /= power
             n += 1
     else:
-        n = units.index(fixed_unit)
+        n = fixed_unit
         size /= power**n
     try:
         unit = units[n]
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -241,10 +241,10 @@ def sort_sizes(size_list):
     return final_list
 
 
-T = TypeVar("T")
+Number = TypeVar("Number", int, float)
 
 
-def clamp(n: T, min_: T, max_: T) -> T:
+def clamp(n: Number, min_: Number, max_: Number) -> Number:
     """Restrict the number n inside a range"""
     return min(max_, max(n, min_))
 
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -258,7 +258,7 @@ def find_best_unit_for_sizes(sizes: Iterable[int], metric: bool = True, precisio
     return find_best_unit_for_size(min_size, metric=metric, precision=precision)
 
 
-def find_best_unit_for_size(size: int, metric: bool = True, precision: int = 1) -> int:
+def find_best_unit_for_size(size: Optional[int], metric: bool = True, precision: int = 1) -> int:
     """
     Selects the index of the biggest unit (see the lists in the pretty_bytes function) capable of
     representing the passed size.
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -246,7 +246,7 @@ def find_best_size_formatting(size_list, metric=True):
         if metric
         else (2**10, ['', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'])
     )
-    min_size = min(size_list)
+    min_size = min(s for s in size_list if isinstance(s, int))
     # we tolerate a maximum precision of 1
     n = 0
     while abs(min_size / power**n > 0.1) and n + 1 < len(units):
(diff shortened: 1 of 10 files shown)


src/vorta/utils.py
```diff
@@ -240,7 +240,22 @@ def sort_sizes(size_list):
...
+        if metric
+        else (2**10, ['', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'])
+    )
+    min_size = min(s for s in size_list if isinstance(s, int))
+    # we tolerate a maximum precision of 1
+    n = 0
+    while abs(min_size / power**n > 0.1) and n + 1 < len(units):
+        n += 1
+    n = max(n - 1, 0)
+    return 1, units[n]
```
real-yfprojects, line 255:
```suggestion
def find_best_size_formatting(sizes, metric=True, precision=1):
    power =  10**3 if metric else 2**10
    min_size = min((s for s in sizes if isinstance(s, int)), default=None)
    if min_size is None:
        return 1, 0
    n = math.ceil(math.log(min_size, power))
    return 1, n if min_size / power**n >= 0.1**precision else n-1
```

real-yfprojects: This simplifies the code and fixes the error.

src/vorta/views/source_tab.py
```diff
@@ -21,6 +21,10 @@ class SourceColumn:
 
 
 class SizeItem(QTableWidgetItem):
+    def __init__(self, s):
+        super().__init__(s)
+        self.setTextAlignment(int('0x82', 16))
```
real-yfprojects, line 26: Please use the `Qt.Alignment` enum.

real-yfprojects (commented):
Please add signature type hints and docstrings for `pretty_bytes` and `find_best_size_formatting`.

real-yfprojects:
[Pre-commit](pre-commit.com) will format your code (including sorting imports) automatically.

src/vorta/utils.py
hariseldon78: thanks, I'll apply it asap. so I deduce it was crashing on an empty list.. what a rookie mistake!

real-yfprojects: >  I deduce it was crashing on an empty list.. what a rookie mistake!

Yes, error message was `src/vorta/utils.py:249: ValueError: min() arg is an empty sequence`.

hariseldon78:
unfortunately pre-commit is not working on my system, this is the error I receive: 
[pre-commit.log](https://github.com/borgbase/vorta/files/10774263/pre-commit.log)
the fixes I found talks about upgrading the isort package, but the version is not pinned in the requirements file, so it should already be at the highest version compatible with the other packages...

hariseldon78:
now that I have pre-commit running, if I run it with --all-files it finds a bunch of other little linting problems; should I commit those? how can I see the github builder error logs?

real-yfprojects:
> [Edit] ok this problem is related to https://github.com/borgbase/vorta/pull/1601, so i've been able to install and run it by doing the `pre-commit autoupdate`

Rebasing your branch onto master will also do the job.

src/vorta/views/archive_tab.py
```diff
@@ -243,12 +250,16 @@ def populate_from_profile(self):
 
             sorting = self.archiveTable.isSortingEnabled()
             self.archiveTable.setSortingEnabled(False)
+            precision = 1  # could be a setting
```
real-yfprojects, line 253: I think, a (global) constant would fit well for this.

src/vorta/utils.py
```diff
@@ -249,10 +268,14 @@ def pretty_bytes(size, metric=True, sign=False, precision=1):
...
     )
-    n = 0
-    while abs(round(size, precision)) >= power and n + 1 < len(units):
-        size /= power
-        n += 1
+    if fixed_unit is None:
+        n = 0
+        while abs(round(size, precision)) >= power and n + 1 < len(units):
+            size /= power
+            n += 1
```
real-yfprojects, line 275: Can this be convertes to a one/twoliner as well?

src/vorta/utils.py
```diff
@@ -249,10 +268,14 @@ def pretty_bytes(size, metric=True, sign=False, precision=1):
...
+    else:
+        n = fixed_unit
+        size /= power**n
     try:
         unit = units[n]
```
real-yfprojects, line 280: What if index out of range? See comment above.

src/vorta/utils.py
```diff
@@ -240,7 +241,25 @@ def sort_sizes(size_list):
...
+    """
+    Selects the index of the biggest unit (see the lists in the pretty_bytes function) capable of
+    representing the smallest size in the sizes iterable.
+    """
+    power = 10**3 if metric else 2**10
+    min_size = min((s for s in sizes if isinstance(s, int)), default=None)
+    if min_size is None:
+        return 0
+    n = math.floor(math.log(min_size * 10**precision, power))
+    return n
```
real-yfprojects, line 254: n might be of metric range, i think

src/vorta/utils.py
```diff
@@ -240,7 +241,25 @@ def sort_sizes(size_list):
...
+def find_best_size_unit(sizes: Iterable[int], metric: bool = True, precision: int = 1) -> int:
+    """
+    Selects the index of the biggest unit (see the lists in the pretty_bytes function) capable of
+    representing the smallest size in the sizes iterable.
+    """
+    power = 10**3 if metric else 2**10
+    min_size = min((s for s in sizes if isinstance(s, int)), default=None)
+    if min_size is None:
+        return 0
+    n = math.floor(math.log(min_size * 10**precision, power))
```
real-yfprojects, line 253: Neat!

real-yfprojects (commented):
Great, you added test cases and cleverly improved my code!

src/vorta/utils.py
hariseldon78: right, better clamp it

src/vorta/utils.py
hariseldon78: not sure what you mean... those lines should work both for metric and power of 2, do you think I'm missing some test cases?

real-yfprojects: Sorry left out some words, updated the comment.

hariseldon78: ok, yes i resolved by clamping in the pretty_bytes function

real-yfprojects (commented):
Can you add a test for a very large size like $10^{30}$?

hariseldon78:
> Can you add a test for a very large size like 1030?

it was failing actually :)

src/vorta/utils.py
```diff
@@ -240,7 +241,41 @@ def sort_sizes(size_list):
...
+    representing the passed size.
+    """
+    if not isinstance(size, int) or size == 0:  # this will also take care of the None case
+        return 0
+    power = 10**3 if metric else 2**10
+    n = math.floor(math.log(size * 10**precision, power))
+    return n
+
+
+def pretty_bytes(size: int, metric: bool = True, sign: bool = False, precision: int = 1, fixed_unit: int = None) -> str:
```
real-yfprojects, line 273:
```suggestion
def pretty_bytes(size: int, metric: bool = True, sign: bool = False, precision: int = 1, fixed_unit: Optional[int] = None) -> str:
```

src/vorta/views/archive_tab.py
```diff
@@ -44,6 +51,8 @@
 
 logger = logging.getLogger(__name__)
 
+PRECISION = 1
```
real-yfprojects, line 54: Please add a docstring for the constant.

hariseldon78: Not sure how to do that, only way I found is the Annotated typing, but that needs python 3.9 at least (https://stackoverflow.com/a/8820636/527935). So instead I renamed the constant to `SIZE_DECIMAL_DIGITS` which seems more clear

real-yfprojects: Sphinx allows documenting python variables. I always use [their `#:`-format](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute).

hariseldon78: ok. I still find the new name self explanatory enough, don't you think?

real-yfprojects: Method, function and class names should be self-explanatory too. However docstrings are mandatory still.

hariseldon78: ok done. If they are mandatory then please update the Contributing guide.

real-yfprojects: > If they are mandatory then please update the Contributing guide.

Quoting the docs:

> Write docstrings at least for methods and functions. Docstrings for classes, attributes and global variables are very welcome too.

So docstrings are mandatory for methods and functions, although you are right that I didn't write the same about other docstrings in the guide. In these cases I will request them as applicable.

src/vorta/utils.py
```diff
@@ -240,7 +241,43 @@ def sort_sizes(size_list):
...
+def find_best_unit_for_sizes(sizes: Iterable[int], metric: bool = True, precision: int = 1) -> int:
+    """
+    Selects the index of the biggest unit (see the lists in the pretty_bytes function) capable of
+    representing the smallest size in the sizes iterable.
+    """
+    min_size = min((s for s in sizes if isinstance(s, int)), default=None)
+    return find_best_unit_for_size(min_size, metric=metric, precision=precision)
+
+
+def find_best_unit_for_size(size: int, metric: bool = True, precision: int = 1) -> int:
```
real-yfprojects, line 261:
```suggestion
def find_best_unit_for_size(size: Optional[int], metric: bool = True, precision: int = 1) -> int:
```

src/vorta/utils.py
```diff
@@ -240,7 +241,43 @@ def sort_sizes(size_list):
     return final_list
 
 
-def pretty_bytes(size, metric=True, sign=False, precision=1):
+T = TypeVar("T")
+
+
+def clamp(n: T, min_: T, max_: T) -> T:
+    """Restrict the number n inside a range"""
+    return min(max_, max(n, min_))
```
real-yfprojects, line 249:
```suggestion
    return min(max_, max(n, min_))  # type: ignore
```

real-yfprojects (commented):
Only some minor typing problems which make them even more minor.

src/vorta/utils.py
hariseldon78: ok I activated mypy in my emacs config, didn't know about that checker! I added the `int, float` specification to the typevar and now it doesn't complain

real-yfprojects (approved):

m3nu:
Thanks all! Give me until this evening to run this locally.

m3nu:
Works as advertised. Interesting addition. Thanks for your work and patience on this!

m3nu (approved):

PR #1603, opened 2023-02-19
Allow copying the public part of the first SSH key

redstrate:
Previously this button would only work if you had more than one SSH key which prevents the first one from being copied.

### Description
This simple change allows copying the public part of the first ssh key, which wasn't accessible previously due to the if check failing on index 1 (which could be a valid key).

### Related Issue
Steps to reproduce:
* Have only one SSH key available.
   * The dropdown should list _"Auto"_ and then _"somekey (ssh-rsa)"_
* Select the _"somekey (ssh-rsa)"_
* The "Copy to Clipboard" function doesn't work, and asks you to select a key - despite you having selected one.

### Motivation and Context
I only have one ssh key registered in Vorta usually, so it's annoying that I can't use the "Copy to Clipboard" functionality to test the key.

### How Has This Been Tested?
I tested this via a local dev environment. The change is so extremely simple I can't imagine any side-effects.

Example of the bug:

image

image

### Types of changes

### Checklist:

- [X] My code follows the code style of this project.
- [X] I have read the [CONTRIBUTING](https://vorta.borgbase.com/contributing/) guide.
- [X] All new and existing tests passed.

### DIFF
--- a/src/vorta/views/repo_tab.py
+++ b/src/vorta/views/repo_tab.py
@@ -203,7 +203,7 @@ class RepoTab(RepoBase, RepoUI, BackupProfileMixin):
         msg.setParent(self, QtCore.Qt.Sheet)
 
         index = self.sshComboBox.currentIndex()
-        if index > 1:
+        if index > 0:
             ssh_key_filename = self.sshComboBox.itemData(index)
             ssh_key_path = os.path.expanduser(f'~/.ssh/{ssh_key_filename}.pub')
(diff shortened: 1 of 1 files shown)


m3nu:
Good find. I never noticed.

m3nu (approved):
I can reproduce the bug and confirm the fix.

m3nu:
Thanks for the contribution and taking the time to fix this! 🙏

Issue #1559, opened 2023-01-21
Remove Paramiko Dependency


m3nu:
This is a heavy depdency and we only use it to verify a file is a SSH key. Would favor removing it and just checking files for starting with `-----BEGIN OPENSSH PRIVATE KEY-----`.

And check for common SSH issues, like wrong ownership or permissions.

Related : #167

real-yfprojects:
Found this: 
https://superuser.com/questions/1515261/how-to-quickly-identify-ssh-private-key-file-formats

m3nu:
Nice. So this confirms the first line. Then I'd just use that to verify the file and not show the key format (doesn't add much value). Most users will just keep it on *Default* anyways. Algo to populate the dropdown would be like this:

1. Grab files in `~/.ssh`
2. Ignore `config` and `known_hosts*`. And `*.pub`
3. For other files read first line.
4. If it matches, show the filename in dropdown.

Right?

diivi:
Hi, I would like to work on this issue, please assign it to me, thanks!

real-yfprojects:
Fixed by #1606

PR #1606, opened 2023-02-22
feat: remove paramiko

diivi:
### Description
Removed paramiko as a dependency and replaced it with a simple function that checks the first line of a file to verify if it's a ssh key.

### Related Issue
#1559  

### Motivation and Context
Helped remove a heavy dependecy used just for checking if a file is a SSH key.

### How Has This Been Tested?
Checked the combobox options

image

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -262,7 +262,7 @@ def find_best_unit_for_size(size: Optional[int], metric: bool = True, precision:
     if not isinstance(size, int) or size == 0:  # this will also take care of the None case
         return 0
     power = 10**3 if metric else 2**10
-    n = math.floor(math.log(size * 10**precision, power))
+    n = math.floor(math.log(abs(size) * 10**precision, power))
     return n
 
 
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -12,10 +12,6 @@ from datetime import datetime as dt
 from functools import reduce
 from typing import Any, Callable, Iterable, Optional, Tuple, TypeVar
 import psutil
-from paramiko import SSHException
-from paramiko.ecdsakey import ECDSAKey
-from paramiko.ed25519key import Ed25519Key
-from paramiko.rsakey import RSAKey
 from PyQt5 import QtCore
 from PyQt5.QtCore import QFileInfo, QThread, pyqtSignal
 from PyQt5.QtWidgets import QApplication, QFileDialog, QSystemTrayIcon
@@ -179,9 +175,17 @@ def choose_file_dialog(parent, title, want_folder=True):
     return dialog
 
 
+def get_ssh_file_fingerprint(filename):
+    """Check if the file is a SSH key."""
+    filepath = os.path.expanduser('~/.ssh/' + filename)
+    with open(filepath, 'rb') as f:
+        first_line = f.readline()
+    if not first_line.startswith(b'-----BEGIN OPENSSH PRIVATE KEY-----'):
+        return None
+
+
 def get_private_keys():
     """Find SSH keys in standard folder."""
-    key_formats = [RSAKey, ECDSAKey, Ed25519Key]
 
     ssh_folder = os.path.expanduser('~/.ssh')
 
@@ -191,35 +195,33 @@ def get_private_keys():
             key_file = os.path.join(ssh_folder, key)
             if not os.path.isfile(key_file):
                 continue
-            for key_format in key_formats:
-                try:
-                    parsed_key = key_format.from_private_key_file(key_file)
+            # ignore config, known_hosts*, *.pub, etc.
+            if key.endswith('.pub') or key.startswith('known_hosts') or key == 'config':
+                continue
+            try:
+                ssh_fingerprint = get_ssh_file_fingerprint(key)
+                if ssh_fingerprint:
                     key_details = {
                         'filename': key,
-                        'format': parsed_key.get_name(),
-                        'bits': parsed_key.get_bits(),
-                        'fingerprint': parsed_key.get_fingerprint().hex(),
+                        'fingerprint': "abc",
                     }
                     available_private_keys.append(key_details)
-                except (
-                    SSHException,
-                    UnicodeDecodeError,
-                    IsADirectoryError,
-                    IndexError,
-                    ValueError,
-                    PermissionError,
-                    NotImplementedError,
-                ):
-                    logger.debug(
-                        f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True
-                    )
+            except (
+                UnicodeDecodeError,
+                IsADirectoryError,
+                IndexError,
+                ValueError,
+                PermissionError,
+                NotImplementedError,
+            ):
+                logger.debug(f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True)
+                continue
+            except OSError as e:
+                if e.errno == errno.ENXIO:
+                    # when key_file is a (ControlPath) socket
                     continue
-                except OSError as e:
-                    if e.errno == errno.ENXIO:
-                        # when key_file is a (ControlPath) socket
-                        continue
-                    else:
-                        raise
+                else:
+                    raise
 
     return available_private_keys
 
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -209,7 +209,7 @@ def get_private_keys() -> List[str]:
                 else:
                     logger.debug(f'Not a private SSH key file: {key}')
             except PermissionError:
-                logger.warning(f'Permission error while opening file in .ssh: {key}', exc_info=True)
+                logger.warning(f'Permission error while opening file: {key_file}', exc_info=True)
                 continue
             except OSError as e:
                 if e.errno == errno.ENXIO:
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -207,11 +207,9 @@ def get_private_keys():
                     available_private_keys.append(key_details)
             except (
                 UnicodeDecodeError,
-                IsADirectoryError,
                 IndexError,
                 ValueError,
                 PermissionError,
-                NotImplementedError,
             ):
                 logger.debug(f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True)
                 continue
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -201,16 +201,8 @@ def get_private_keys():
                 continue
             try:
                 if is_ssh_file(key):
-                    key_details = {
-                        'filename': key,
-                    }
-                    available_private_keys.append(key_details)
-            except (
-                UnicodeDecodeError,
-                IndexError,
-                ValueError,
-                PermissionError,
-            ):
+                    available_private_keys.append(key)
+            except (PermissionError,):
                 logger.debug(f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True)
                 continue
             except OSError as e:
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -10,7 +10,6 @@ import sys
 import unicodedata
 from datetime import datetime as dt
 from functools import reduce
-from pathlib import Path
 from typing import Any, Callable, Iterable, List, Optional, Tuple, TypeVar
 import psutil
 from PyQt5 import QtCore
@@ -200,8 +199,7 @@ def get_private_keys() -> List[str]:
             if key.endswith('.pub') or key.startswith('known_hosts') or key == 'config':
                 continue
             try:
-                filepath = Path.home() / '.ssh' / key
-                if is_ssh_file(filepath):
+                if is_ssh_file(key_file):
                     available_private_keys.append(key)
             except (PermissionError,):
                 logger.debug(f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True)
--- a/src/vorta/utils.py
+++ b/src/vorta/utils.py
@@ -177,11 +177,10 @@ def choose_file_dialog(parent, title, want_folder=True):
 
 def is_ssh_file(filepath: str) -> bool:
     """Check if the file is a SSH key."""
-    with open(filepath, 'rb') as f:
+    with open(filepath, 'r') as f:
         first_line = f.readline()
-    if not first_line.startswith(b'-----BEGIN OPENSSH PRIVATE KEY-----'):
-        return False
-    return True
+    pattern = r'^-----BEGIN(\s\w+)? PRIVATE KEY-----'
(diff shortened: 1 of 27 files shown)


src/vorta/utils.py
```diff
@@ -191,35 +196,31 @@ def get_private_keys():
...
-                        f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True
-                    )
+            except (
+                UnicodeDecodeError,
+                IsADirectoryError,
+                IndexError,
+                ValueError,
+                PermissionError,
+                NotImplementedError,
+            ):
```
real-yfprojects, line 215: Do we need to catch all of those errors?

diivi: yeah, I needed some context about why they were here at the first place, thanks! I will remove `NotImplementedError` and `IsADirectoryError`

m3nu: Probably some more to remove.

src/vorta/utils.py
```diff
@@ -191,35 +196,21 @@ def get_private_keys():
...
-                    logger.debug(
-                        f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True
-                    )
+            # ignore config, known_hosts*, *.pub, etc.
+            if key.endswith('.pub') or key.startswith('known_hosts') or key == 'config':
+                continue
+            try:
+                if is_ssh_file(key):
+                    available_private_keys.append(key)
+            except (PermissionError,):
```
real-yfprojects, line 205: @m3nu in which cases is a permission error expected?

src/vorta/utils.py
```diff
@@ -179,9 +175,18 @@ def choose_file_dialog(parent, title, want_folder=True):
     return dialog
 
 
+def is_ssh_file(filename):
+    """Check if the file is a SSH key."""
+    filepath = os.path.expanduser('~/.ssh/' + filename)
```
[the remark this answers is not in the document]
diivi: should I replace this line with `filepath = Path.home() / '.ssh' / filename`?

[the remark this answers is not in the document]
real-yfprojects: No, you don't have to use `pathlib`. I thought you were passing the complete filepath to this function but I just noticed that you have to variables `key_file` and `key`. So this line is correct. 

However I think, the function should expect a complete file path removing redundancy and making the code more reusable.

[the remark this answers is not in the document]
diivi: Agreed, I will use filepath here instead of filename for reusability 👍🏼

src/vorta/utils.py
```diff
@@ -179,9 +175,17 @@ def choose_file_dialog(parent, title, want_folder=True):
     return dialog
 
 
-def get_private_keys():
+def is_ssh_file(filepath: str) -> bool:
+    """Check if the file is a SSH key."""
+    with open(filepath, 'rb') as f:
+        first_line = f.readline()
+    if not first_line.startswith(b'-----BEGIN OPENSSH PRIVATE KEY-----'):
```
real-yfprojects, line 182: I just wondered whether we should only support openssh key files instead of all files starting with `-----BEGIN( \w+)? PRIVATE KEY-----`. But I guess if anyone needs that, we will see an FR opened.

real-yfprojects (dismissed):

src/vorta/utils.py
m3nu: Yes, would be better. I have one SSH key here that starts with `-----BEGIN RSA PRIVATE KEY-----`. It's an older key I made for AWS EC2.

So there could be 3 different headers: https://superuser.com/a/1444343/890011

m3nu (changes requested):

src/vorta/utils.py
diivi: Okay, I'll use the regex @real-yfprojects mentioned

src/vorta/utils.py
m3nu: It's possible. Like when people run commands as root, while `$HOME` is still set for their own user.

m3nu: We should also add the `UnicodeDecodeError`. Else it crashes for me while parsing a macOS-specific file in that folder

```
2023-02-24 12:58:35,597 - root - DEBUG - Expected error parsing file in .ssh: .DS_Store (You can safely ignore this)
  File "/Users/manu/Workspace/vorta/src/vorta/utils.py", line 201, in get_private_keys
    if is_ssh_file(key_file):
       ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/manu/Workspace/vorta/src/vorta/utils.py", line 181, in is_ssh_file
    first_line = f.readline()
                 ^^^^^^^^^^^^
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 3131: invalid start byte
```

m3nu (changes requested):

src/vorta/utils.py
```diff
@@ -191,35 +194,19 @@ def get_private_keys():
...
-                    logger.debug(
-                        f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True
-                    )
+            # ignore config, known_hosts*, *.pub, etc.
+            if key.endswith('.pub') or key.startswith('known_hosts') or key == 'config':
+                continue
+            try:
+                if is_ssh_file(key_file):
+                    available_private_keys.append(key)
+            except (PermissionError, UnicodeDecodeError):
```
real-yfprojects, line 203: Please add a comment in the code on why we need special handling for these errors. (They are subclasses of `OSError`.)

real-yfprojects (approved):

real-yfprojects (changes requested):
Almost forgot: This PR introduced some clutter in the form on line break removals. I can fix that during merging though.

diivi:
Yeah, the pre-commit hooks made some deletions after my changes 😅

src/vorta/utils.py
```diff
@@ -179,9 +175,16 @@ def choose_file_dialog(parent, title, want_folder=True):
     return dialog
 
 
-def get_private_keys():
+def is_ssh_file(filepath: str) -> bool:
```
ThomasWaldmann, line 178: how about `is_ssh_private_key_file`?

src/vorta/utils.py
```diff
@@ -191,35 +194,22 @@ def get_private_keys():
...
+            # ignore config, known_hosts*, *.pub, etc.
+            if key.endswith('.pub') or key.startswith('known_hosts') or key == 'config':
+                continue
+            try:
+                if is_ssh_file(key_file):
+                    available_private_keys.append(key)
+            except (PermissionError, UnicodeDecodeError):
+                # Handling PermissionError separately from OSError because it can be safely ignored
+                # (If the user doesn't have permission to read the file, it's not an unexpected error)
+                # If UnicodeDecodeError is raised, it's because the file is not a valid SSH key
```
ThomasWaldmann, line 206: Maybe you could avoid UnicodeDecodeError locally in the called function.
If you get that there, you will know for sure it is not a private key and it can return False.

PermissionError is another problem, we can't know what it is in that case (we can just know we can not use it due to that).

Also check the comment for orthography and punctuation.

src/vorta/utils.py
```diff
@@ -191,35 +194,22 @@ def get_private_keys():
...
+            except OSError as e:
+                if e.errno == errno.ENXIO:
+                    # when key_file is a (ControlPath) socket
                     continue
-                except OSError as e:
-                    if e.errno == errno.ENXIO:
-                        # when key_file is a (ControlPath) socket
-                        continue
-                    else:
-                        raise
```
ThomasWaldmann, line 222: You have an "if", but you dropped the "else: raise".

The "continue" in the if-block basically means "do nothing" and as you have removed the else-block, it is also doing nothing there.

Looks suspicious.

diivi: So yeah, I guess I should raise the error in the end.

src/vorta/utils.py
```diff
@@ -191,35 +196,21 @@ def get_private_keys():
...
+                else:
```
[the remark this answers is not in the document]
real-yfprojects: As *ThomasWaldmann* pointed out, the `raise` is needed. However the else isn't.

src/vorta/utils.py
```diff
@@ -198,18 +201,19 @@ def get_private_keys() -> List[str]:
...
-                if is_ssh_file(key_file):
+                if is_ssh_private_key_file(key_file):
                     available_private_keys.append(key)
-            except (PermissionError, UnicodeDecodeError):
+            except PermissionError:
                 # Handling PermissionError separately from OSError because it can be safely ignored
-                # (If the user doesn't have permission to read the file, it's not an unexpected error)
-                # If UnicodeDecodeError is raised, it's because the file is not a valid SSH key
+                # (if the user doesn't have permission to read the file, it's not an unexpected error).
                 logger.debug(f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True)
```
ThomasWaldmann, line 209: Why is this "expected"?

real-yfprojects: Maybe issueing a warning for the file would be better.

ThomasWaldmann: Hmm, just noticed that this hasn't changed, it was "debug" and "expected" before this PR already.

ThomasWaldmann (approved):
LGTM.

Not sure where the empty line removals come from or whether they are desired or not.

ThomasWaldmann:
Hmm, the github actions lint check runs black and is fine with these changes.

Considering the formatting was changed (empty lines), does that mean the formatting was non-black before this PR (why?)?

src/vorta/utils.py
real-yfprojects: Your point still applies I guess

real-yfprojects:
I will have a look whether the changes are needed for ci. If they are, I will move them into a seperate commit.

Hofer-Julian:
> I will have a look whether the changes are needed for ci. If they are, I will move them into a seperate commit.

We probably want to run pre-commit in the CI instead of relying on `make lint` :)
https://github.com/borgbase/vorta/blob/a00ed62e491919bd9dc342d61579c8f0f17bc417/.github/workflows/test.yml#L28

src/vorta/utils.py
m3nu: This log line used to be for any private key error. It's much more specific now and would only trigger, when the current user can't read the file. If we want to keep the old behavior of logging anything that's not a private key, the line needs to go up:

```
if is_ssh_private_key_file(key_file):
    available_private_keys.append(key)
else:
    logger.debug("Not a private SSH key file...")
```

m3nu: Since we already have the permission check here, we could also check for correct key permissions. This sometimes comes up in support requests. Specifically SSH won't use the key if perms are **too open**. Like 644 or higher. Could be worth including and logging too.

m3nu (changes requested):

src/vorta/utils.py
ThomasWaldmann: Guess it should check for `== 0o600` and else log a user-visible warning (not debug).

ThomasWaldmann: Or even better: `mode & 0o077 == 0`.

Permissions for the owner could be r or rw (or even rwx).

diivi:
```python
logger.debug(f'Expected error parsing file in .ssh: {key} (You can safely ignore this)', exc_info=True)
```
should I replace this line then? With this:
```python
logger.debug(f'Permission error while opening file in .ssh: {key}', exc_info=True)
```

real-yfprojects: Yes.

src/vorta/utils.py
```diff
@@ -202,11 +202,16 @@ def get_private_keys() -> List[str]:
...
-                    available_private_keys.append(key)
+                    if os.stat(key_file).st_mode & 0o077 == 0:
+                        available_private_keys.append(key)
+                    else:
+                        logger.warning(f'Permissions for {key_file} are too open.')
+                else:
+                    logger.debug(f'Not a private SSH key file: {key}')
             except PermissionError:
                 # Handling PermissionError separately from OSError because it can be safely ignored
                 # (if the user doesn't have permission to read the file, it's not an unexpected error).
```
real-yfprojects, line 213: This comment line is no longer up-to-date.

src/vorta/utils.py
```diff
@@ -191,35 +197,27 @@ def get_private_keys():
...
+                        available_private_keys.append(key)
                     else:
-                        raise
+                        logger.warning(f'Permissions for {key_file} are too open.')
+                else:
+                    logger.debug(f'Not a private SSH key file: {key}')
+            except PermissionError:
+                # Handling PermissionError separately from OSError because it can be safely ignored
+                # (if the user doesn't have permission to read the file, it's not an unexpected error).
+                logger.debug(f'Permission error while opening file in .ssh: {key}', exc_info=True)
```
real-yfprojects, line 214: Is debug the right logging level?

diivi: How about warning?

src/vorta/utils.py
diivi: Do I need the comment now? (The warning seems self-explanatory, we handled PermissionError separately to let the user know about an expected error)

real-yfprojects: You are right it is not needed any more.

src/vorta/utils.py
real-yfprojects: :+1:

src/vorta/utils.py
```diff
@@ -191,35 +197,25 @@ def get_private_keys():
...
+                if is_ssh_private_key_file(key_file):
+                    if os.stat(key_file).st_mode & 0o077 == 0:
+                        available_private_keys.append(key)
                     else:
-                        raise
+                        logger.warning(f'Permissions for {key_file} are too open.')
+                else:
+                    logger.debug(f'Not a private SSH key file: {key}')
+            except PermissionError:
+                logger.warning(f'Permission error while opening file in .ssh: {key}', exc_info=True)
```
ThomasWaldmann, line 212: Use `key_file` (== full path) in all log messages?

ThomasWaldmann: (then you do not need to mention "in .ssh" either)

ThomasWaldmann (changes requested):
minor consistency change

m3nu (dismissed):
Looks good. It's not a big shiny feature, but an important incremental improvement and refactoring.

ThomasWaldmann (approved):
LGTM.

m3nu (dismissed):

m3nu:
Thanks for your patience.

You wanna merge this, @real-yfprojects ? There was some note on formatting you added. Also needs squashing.

real-yfprojects (approved):

Issue #1486, opened 2022-11-09
Say where the logs are when telling about warnings in borg execution


christophehenry:
#### The problem

Borg execution may complete with warning in which case a message advising to consult the logs a displayed in Vorta. But nowhere it sayd how to find theese logs. 

#### Requested Solution

Provide a link to the log files or the CLI command to execute to consult the logs about the warnings.

christophehenry:
I found this link after searching on internet where to locate Vorta's logs. The fact that I had to search and it was not immediatly obvious is the sign of a UX issue here. If you advise to look at the logs, you should immediatly provide a simple way to consult those logs, that's a good practice.

Saiesh-Agre:
You could provide a button or a link for accessing log in the message itself.
Ex : You can find logs 'here'
or you could point where the logs are.

diivi:
@real-yfprojects This can be closed now :)

PR #1609, opened 2023-02-24
feat: add link to the logs folder in borg warnings

diivi:
### Description
Adds a link to the vorta logs folder on the user's device in the `progressText` qLabel.
Please let me know in what state I should leave the strings for translation, thanks.

### Related Issue
#1486 

### Motivation and Context
Easier to open the logs folder without having to look for it (better UX).

image

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -327,9 +327,8 @@ class VortaApp(QtSingleApplication):
                 # warning
                 msg.setIcon(QMessageBox.Icon.Warning)
                 text = translate(
-                    'VortaApp',
-                    f'Borg exited with a warning message. See <a href="{Path(LOG_DIR).as_uri()}">logs</a> for details.',
-                )
+                    'VortaApp', 'Borg exited with a warning message. See <a href="{0}">logs</a> for details.'
+                ).format(Path(LOG_DIR).as_uri())
                 infotext = error_message
             elif returncode > 128:
                 # 128+N - killed by signal N (e.g. 137 == kill -9)
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -23,7 +23,7 @@ from vorta.views.main_window import MainWindow
 
 logger = logging.getLogger(__name__)
 
-APP_ID = os.path.join(TEMP_DIR, "socket")
+APP_ID = TEMP_DIR / "socket"
 
 
 class VortaApp(QtSingleApplication):
@@ -42,7 +42,7 @@ class VortaApp(QtSingleApplication):
     check_failed_event = QtCore.pyqtSignal(dict)
 
     def __init__(self, args_raw, single_app=False):
-        super().__init__(APP_ID, args_raw)
+        super().__init__(str(APP_ID), args_raw)
         args = parse_args()
         if self.isRunning():
             if single_app:
@@ -194,8 +194,8 @@ class VortaApp(QtSingleApplication):
         This function tries reading a file that is known to be restricted and warn the user about
         incomplete backups.
         """
-        test_path = os.path.expanduser('~/Library/Cookies')
-        if os.path.exists(test_path) and not os.access(test_path, os.R_OK):
+        test_path = Path('~/Library/Cookies').expanduser()
+        if test_path.exists() and not os.access(test_path, os.R_OK):
             msg = QMessageBox()
             msg.setIcon(QMessageBox.Warning)
             msg.setTextInteractionFlags(QtCore.Qt.LinksAccessibleByMouse)
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -328,7 +328,7 @@ class VortaApp(QtSingleApplication):
                 msg.setIcon(QMessageBox.Icon.Warning)
                 text = translate(
                     'VortaApp', 'Borg exited with warning status (rc 1). See the <a href="{0}">logs</a> for details.'
-                ).format(Path(LOG_DIR).as_uri())
+                ).format(LOG_DIR.as_uri())
                 infotext = error_message
             elif returncode > 128:
                 # 128+N - killed by signal N (e.g. 137 == kill -9)
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -327,7 +327,7 @@ class VortaApp(QtSingleApplication):
                 # warning
                 msg.setIcon(QMessageBox.Icon.Warning)
                 text = translate(
-                    'VortaApp', 'Borg exited with warning status (rc 1). See <a href="{0}">logs</a> for details.'
+                    'VortaApp', 'Borg exited with warning status (rc 1). See the <a href="{0}">logs</a> for details.'
                 ).format(Path(LOG_DIR).as_uri())
                 infotext = error_message
             elif returncode > 128:
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -47,11 +47,11 @@ class VortaApp(QtSingleApplication):
         if self.isRunning():
             if single_app:
                 self.sendMessage("open main window")
-                print('An instance of Vorta is already running. Opening main window.')
+                logger.info('An instance of Vorta is already running. Opening main window.')
                 sys.exit()
             elif args.profile:
                 self.sendMessage(f"create {args.profile}")
-                print('Creating backup using existing Vorta instance.')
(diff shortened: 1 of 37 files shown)


src/vorta/borg/check.py
```diff
@@ -20,7 +22,9 @@ def finished_event(self, result: Dict[str, Any]):
         self.app.backup_finished_event.emit(result)
         self.result.emit(result)
         if result['returncode'] != 0:
-            self.app.backup_progress_event.emit(self.tr('Repo check failed. See logs for details.'))
+            self.app.backup_progress_event.emit(
+                trans_late('messages', f'Repo check failed. See <a href="file://{LOG_DIR}">logs</a> for details.')
```
real-yfprojects, line 26:
```suggestion
                trans_late('BorgCheckJob', f'Repo check failed. See <a href="file://{LOG_DIR}">logs</a> for details.')
```

Usually the context equals the class the string is used in. In this particular case there might be good reasons to use a different context, though I wouldn't consider `messages` a descriptive context.

src/vorta/application.py
```diff
@@ -325,7 +325,10 @@ def check_failed_response(self, result: Dict[str, Any]):
             if returncode == 1:
                 # warning
                 msg.setIcon(QMessageBox.Icon.Warning)
-                text = self.tr('Borg exited with a warning message. See logs for details.')
+                text = trans_late(
+                    'messages',
+                    f'Borg exited with a warning message. See <a href="file://{LOG_DIR}">logs</a> for details.',
```
real-yfprojects, line 330: For platform independency `pathlib.Path.as_uri()` would find good use here.

real-yfprojects (commented):
What purpose serves the last commit? You don't have to modify any translation files. That's all done automatically in preparation of a release. There is a rule for that in our make file. 

You will need to use `translate` instead of `trans_late` else the translations won't be loaded at runtime. Please tell me if any part of the contributing docs is unclear.

src/vorta/borg/check.py
diivi: from the docs - `However sometimes you want to mark a string, use the english version and translate it later. In these cases you have to use trans_late and call translate on the string later.` I thought `translate it later` here meant that this function is used for strings which are not yet translated (untranslated), and translations for these strings will be done by people later, so I used that. But yes, now I understand that the "later" relates to the runtime 😄. I will replace these with translate, thanks.

src/vorta/application.py
diivi: Should I replace this for the `Log` text in the misc tab in this PR too?

real-yfprojects: That would be great. It doesn't matter for Linux and MacOS support but in anticipation of the Windows alpha being worked on in #1592 it does.

src/vorta/borg/check.py
real-yfprojects: > I thought `translate it later` here meant that this function is used for strings which are not yet translated (untranslated), and translations for these strings will be done by people later, so I used that.

I see the source of misunderstanding. I'll think about a way to clarify that.

src/vorta/application.py
diivi: Cool, thanks for the context.

src/vorta/borg/check.py
real-yfprojects: What do you think about borgbase/vorta.borgbase.com#37?

diivi: It's better. thanks :)

src/vorta/application.py
diivi: I just checked `config.py`. 
isn't this enough for platform independence?
```python
dirs = appdirs.AppDirs(APP_NAME, APP_AUTHOR)
LOG_DIR = dirs.user_log_dir
```

real-yfprojects: Windows path use backslashes instead of forward slashes and they have drive names.

src/vorta/application.py
```diff
@@ -325,7 +326,10 @@ def check_failed_response(self, result: Dict[str, Any]):
             if returncode == 1:
                 # warning
                 msg.setIcon(QMessageBox.Icon.Warning)
-                text = self.tr('Borg exited with a warning message. See logs for details.')
+                text = translate(
+                    'VortaApp',
+                    f'Borg exited with a warning message. See <a href="{Path(LOG_DIR).as_uri()}">logs</a> for details.',
+                )
```
real-yfprojects, line 332: I noticed a few things, i wasn't aware of before. In my local testing Qt won't recognise the translate statements spanning over multiple lines, for some reason this works when using `self.tr` :shrug:. Additionally, Qt won't be able to translate a string if it changes on runtime. You therefore have to format the string after you translate it.

diivi: should I try - 
```python
translate('VortaApp','Borg exited with a warning message. See <a href="{0}">logs</a> for details.'.format(Path(LOG_DIR).as_uri()))
```
 - from https://stackoverflow.com/a/18367425

real-yfprojects: Almost. I think the StackOverflow answer is wrong.
```python
translate('VortaApp','Borg exited with a warning message. See <a href="{0}">logs</a> for details.').format(Path(LOG_DIR).as_uri())
 ```

**Review by real-yfprojects — APPROVED** [review]
>  In my local testing Qt won't recognise the translate statements spanning over multiple lines

Luckily it seems to work again/with this PR.

File: src/vorta/borg/create.py [hunk 4]
```diff
@@ -33,7 +35,13 @@ def process_result(self, result):
                 repo.save()
 
             if result['returncode'] == 1:
-                self.app.backup_progress_event.emit(self.tr('Backup finished with warnings. See logs for details.'))
+                self.app.backup_progress_event.emit(
+                    self.app.backup_progress_event.emit(
+                        translate(
+                            'BorgCreateJob', 'Backup finished with warnings.See <a href="{0}">logs</a> for details.'
```
ThomasWaldmann, line 41: better wording here, but note that there is a space missing after the dot.

src/vorta/application.py
```diff
@@ -325,7 +326,9 @@ def check_failed_response(self, result: Dict[str, Any]):
             if returncode == 1:
                 # warning
                 msg.setIcon(QMessageBox.Icon.Warning)
-                text = self.tr('Borg exited with a warning message. See logs for details.')
+                text = translate(
+                    'VortaApp', 'Borg exited with warning status (rc 1). See <a href="{0}">logs</a> for details.'
```
ThomasWaldmann, line 330: considering that the translations need an update now anyway, guess we could fix the original string:

"See **the** logs for details."

Also:

Do we have multiple logs or why the plural?

And even if we have multiple - for that specific backup, only one of them is meant, right?

real-yfprojects: > Do we have multiple logs or why the plural?

We have a single log spread over multiple files (RotatingFileHandler).

ThomasWaldmann (approved):
LGTM.

src/vorta/borg/create.py
```diff
@@ -33,7 +35,14 @@ def process_result(self, result):
                 repo.save()
 
             if result['returncode'] == 1:
-                self.app.backup_progress_event.emit(self.tr('Backup finished with warnings. See logs for details.'))
+                self.app.backup_progress_event.emit(
```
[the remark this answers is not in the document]
m3nu: No. You're emitting this event 2x. The first sets the correct text. The second just sets `None`, hence the status stays empty.

Did you test this case at all before committing it?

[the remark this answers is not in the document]
m3nu: You have nested 2 event calls within one another. Check the line below and compare to the other files you changed.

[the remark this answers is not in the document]
diivi: Oh, I am sorry, a really silly mistake from my end. 😓

real-yfprojects:
@diivi Can you rebase onto master and adjust this PR. `LOG_DIR` is of type `pathlib.Path` now.

m3nu (approved):
Works as advertised. 👏

real-yfprojects (approved):

ThomasWaldmann (approved):
Guess the checking for rc 0, 1, 2 could be improved, but that is not in scope of this PR.

0 == OK
1 == Warning
2 == Error

Any messages derived from this should use a corresponding wording.

Issue #1611, opened 2023-02-25
Add `flake8-print` linter.


real-yfprojects:
I would like to add the [flake8-print] plugin to our ci and pre-commit hooks. It checks production code for `print` and `pprint` statements and rejects them.

real-yfprojects:
Closing because we decided to try out [ruff](https://beta.ruff.rs/docs/) instead. See [this thread](https://github.com/borgbase/vorta/pull/1612#issuecomment-1447066184).

PR #1612, opened 2023-02-26
ci: add flake8-print linter

diivi:
#1611 
Added [flake8-print](https://github.com/JBKahn/flake8-print), 
referred the flake-8 docs for additional-dependencies - https://flake8.pycqa.org/en/latest/user/using-hooks.html

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,6 +61,7 @@ repos:
     rev: 6.0.0
     hooks:
       - id: flake8
+        additional_dependencies: [flake8-print]
 
 # configuration for the pre-commit.ci bot
 # only relevant when actually using the bot
--- a/requirements.d/dev.txt
+++ b/requirements.d/dev.txt
@@ -1,9 +1,10 @@
-pre-commit
 black==22.*
 coverage
 flake8
+flake8-print
 isort
 macholib
+pre-commit
 pyinstaller
 pylint
 pytest


real-yfprojects:
We do use `print` at some places. So try restricting the plugin to production code and add ignore comments to the valid lines.

diivi:
Okay, so should I ignore the `./package` directory and add ignore comments to the print statements used in `./src`?

src/vorta/application.py
```diff
@@ -46,11 +46,11 @@ def __init__(self, args_raw, single_app=False):
         if self.isRunning():
             if single_app:
                 self.sendMessage("open main window")
-                print('An instance of Vorta is already running. Opening main window.')
+                print('An instance of Vorta is already running. Opening main window.')  # noqa: T20
```
m3nu, line 49: Could this be a log statement instead?

diivi: I don't really know why this was a print statement at the first place, can you give me some context? cc:@real-yfprojects

Hofer-Julian: I am to blame for this line.
log would indeed be better

m3nu:
It's good to clean up those print statements and ensure none are left after debugging. 👍

Just noting that the `flake8-print` package isn't very popular and looks like the thing that may become unmaintained eventually. There's also supply chain attacks from such less popular packages. So always good to think twice before adding a new dependency. Maybe one of the larger linters adds this feature by then. 🤞

m3nu (approved):

Hofer-Julian:
> It's good to clean up those print statements and ensure none are left after debugging. 👍
> 
> Just noting that the `flake8-print` package isn't very popular and looks like the thing that may become unmaintained eventually. There's also supply chain attacks from such less popular packages. So always good to think twice before adding a new dependency. Maybe one of the larger linters adds this feature by then. 🤞

Agreed, I like ruff as a flake8 and isort alternative.
it also implements this rule: https://beta.ruff.rs/docs/rules/#flake8-print-t20

real-yfprojects:
Ruff seems to gain popularity already. I don't know how stable it is at the moment but I am open for switching once it is.

Hofer-Julian:
> Ruff seems to gain popularity already. I don't know how stable it is at the moment but I am open for switching once it is.

Fair enough.
Regarding this PR, how about we merge the changes, but don't add the flake8-dependency?

m3nu:
> Regarding this PR, how about we merge the changes, but don't add the flake8-dependency?

If @diivi wants to get this merged, this would be the fastest way.

For extra points, he could add [Ruff](https://github.com/charliermarsh/ruff) and make sure it detects the print statement. And we run it in parallel with flake8 for a while. Then remove it, once we are comfortable with it. Better to add a dependency with 10k stars than 100.

diivi:
Yup, ruff looks like a better alternative, I'll remove flake8-print from this PR and make a new one to add it in ruff.

Hofer-Julian:
> For extra points, he could add [Ruff](https://github.com/charliermarsh/ruff) and make sure it detects the print statement. And we run it in parallel with flake8 for a while. Then remove it, once we are comfortable with it. Better to add a dependency with 10k stars than 100.

Sounds good to me.

>  Should I rename this to chore: replaced print with logging in application.py?

If I look at past commits, `Replace print with logging in application.py` without the `chore` would be fine

Hofer-Julian (dismissed):

real-yfprojects (commented):
There still is one `print` in production code. But I don't think replacing it would make sense.

https://github.com/borgbase/vorta/blob/c7c935c5fcb679ad16829fcd9306ce5543f5869f/src/vorta/__main__.py#L50

m3nu (approved):
Looks like a big improvement. 😄

real-yfprojects (approved):

PR #1613, opened 2023-02-26
feat: more descriptive messages when no sources are selected to backup

### LINKED ISSUE #1463: FR: More descriptive message when there are no sources.
When the user didn't configure any sources, the message `borg: error: Need at least one PATH argument`.
Instead vorta should show something a long the lines of `Select sources to backup first`

diivi:
fixes #1463 
image


real-yfprojects (changes requested):
I think a change to create.py would be better suited.

diivi:
I looked through `create.py`, but didn't understand how it can detect an error from borg.
The `run` method is responsible for executing a command, and only that method knows whether borg returned this error message.
This method is called from the JobsManager which should remain untouched.
So how can I parse the output that borg gives from a subclass of `BorgJob`, when only BorgJob actually runs the command? Thanks

real-yfprojects:
While you can't catch the error in `create.py` you can foresee it. This is already done but not working correctly as we noticed:

https://github.com/borgbase/vorta/blob/e20e923ba92ddd2de9f8a3a5a404c7a954c0317a/src/vorta/borg/create.py#L99-L101

real-yfprojects (approved):
Great! This is exactly what I had in mind.

PR #1617, opened 2023-02-28
Migrate from appdirs to platformdirs

ganeshrevadi:
Migrated form Appdirs to platformdirs. Also used *_path api returning pathlib.Path instead of str.


src/vorta/autostart.py
```diff
@@ -58,7 +58,7 @@ def open_app_at_startup(enabled=True):
         if is_flatpak:
             autostart_path = Path.home() / '.config' / 'autostart'
         else:
-            autostart_path = Path(user_config_dir("autostart"))
+            autostart_path = Path(user_config_path("autostart"))
```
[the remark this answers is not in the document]
ganeshrevadi: Thanks for the inputs . Creating another PR with the suggested changes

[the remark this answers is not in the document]
diivi: No need to create another one, you can add commits to this PR. 😄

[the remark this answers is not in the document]
ganeshrevadi: Thanks for the inputs. Suggestion commit done!

src/vorta/config.py
```diff
@@ -1,11 +1,11 @@
...
+import platformdirs
 
 APP_NAME = 'Vorta'
 APP_AUTHOR = 'BorgBase'
 APP_ID_DARWIN = 'com.borgbase.client.macos'
-dirs = appdirs.AppDirs(APP_NAME, APP_AUTHOR)
+dirs = platformdirs.PlatformDirs(APP_NAME, APP_AUTHOR)
 SETTINGS_DIR = dirs.user_data_dir
 LOG_DIR = dirs.user_log_dir
 CACHE_DIR = dirs.user_cache_dir
```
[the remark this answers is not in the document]
ganeshrevadi: Done!

src/vorta/config.py
```diff
@@ -1,14 +1,14 @@
...
 APP_AUTHOR = 'BorgBase'
 APP_ID_DARWIN = 'com.borgbase.client.macos'
-dirs = appdirs.AppDirs(APP_NAME, APP_AUTHOR)
-SETTINGS_DIR = dirs.user_data_dir
-LOG_DIR = dirs.user_log_dir
-CACHE_DIR = dirs.user_cache_dir
+dirs = platformdirs.PlatformDirs(APP_NAME, APP_AUTHOR)
+SETTINGS_DIR = dirs.user_data_path
+LOG_DIR = dirs.user_log_path
+CACHE_DIR = dirs.user_cache_path
```
real-yfprojects, line 11: Have you checked all uses of these constants for compatibility with this change?

src/vorta/config.py
```diff
@@ -1,14 +1,14 @@
...
 TEMP_DIR = os.path.join(CACHE_DIR, "tmp")
 PROFILE_BOOTSTRAP_FILE = Path.home() / '.vorta-init.json'
```
real-yfprojects, line 14: What about simplifying the existing code like that?

```py
# ensure directories exist
for dir in (SETTINGS_DIR, LOG_DIR, CACHE_DIR, TEMP_DIR):
    dir.mkdir(parents=True, exist_ok=True)
```

src/vorta/config.py
ganeshrevadi: I have checked and everything works fine according to me

src/vorta/config.py
ganeshrevadi: I tried simplifying as per your suggestion but **TEMP_DIR** return a *str* so I have converted it into *path* is that fine?

real-yfprojects: That is what I intended.

src/vorta/config.py
```diff
@@ -1,25 +1,18 @@
...
-dirs = appdirs.AppDirs(APP_NAME, APP_AUTHOR)
-SETTINGS_DIR = dirs.user_data_dir
-LOG_DIR = dirs.user_log_dir
-CACHE_DIR = dirs.user_cache_dir
-TEMP_DIR = os.path.join(CACHE_DIR, "tmp")
+dirs = platformdirs.PlatformDirs(APP_NAME, APP_AUTHOR)
+SETTINGS_DIR = dirs.user_data_path
+LOG_DIR = dirs.user_log_path
+CACHE_DIR = dirs.user_cache_path
+TEMP_DIR = Path(os.path.join(CACHE_DIR, "tmp"))
```
real-yfprojects, line 12:
```suggestion
TEMP_DIR = CACHE_DIR / "tmp"
```

src/vorta/config.py
```diff
@@ -1,25 +1,18 @@
...
 
-if not os.path.exists(LOG_DIR):
-    os.makedirs(LOG_DIR)
-
-if not os.path.exists(CACHE_DIR):
-    os.makedirs(CACHE_DIR)
-
-if not os.path.exists(TEMP_DIR):
-    os.makedirs(TEMP_DIR)
+# # ensure directories exist
```
real-yfprojects, line 16:
```suggestion
# ensure directories exist
```

.tx/config
```diff
@@ -7,4 +7,3 @@ source_file  = src/vorta/i18n/ts/vorta.en.ts
 source_lang  = en
 type         = QT
 minimum_perc = 80
-
```
ThomasWaldmann, line 10: why do we have i18n changes in here?

ThomasWaldmann (changes requested):

.tx/config
real-yfprojects: I'll remove them when merging.

src/vorta/config.py
```diff
@@ -1,25 +1,17 @@
...
 APP_ID_DARWIN = 'com.borgbase.client.macos'
-dirs = appdirs.AppDirs(APP_NAME, APP_AUTHOR)
-SETTINGS_DIR = dirs.user_data_dir
-LOG_DIR = dirs.user_log_dir
-CACHE_DIR = dirs.user_cache_dir
-TEMP_DIR = os.path.join(CACHE_DIR, "tmp")
+dirs = platformdirs.PlatformDirs(APP_NAME, APP_AUTHOR)
+SETTINGS_DIR = dirs.user_data_path
+LOG_DIR = dirs.user_log_path
+CACHE_DIR = dirs.user_cache_path
```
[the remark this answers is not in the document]
ThomasWaldmann: (was already asked/answered)

setup.cfg
```diff
@@ -37,7 +37,7 @@ package_dir =
 include_package_data = true
 python_requires = >=3.7
 install_requires =
-  appdirs
+  platformdirs
```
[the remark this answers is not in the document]
ganeshrevadi: `  platformdirs >=3.0.0, <4.0.0; sys_platform == 'darwin'  # for macOS: breaking changes in 3.0.0,
  platformdirs >=2.6.0, <4.0.0; sys_platform != 'darwin'  # for others: 2.6+ works consistently.`
Did you mean these lines ?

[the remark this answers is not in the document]
ganeshrevadi:
```
    platformdirs >=3.0.0, <4.0.0; sys_platform == 'darwin'  # for macOS: breaking changes in 3.0.0,
  platformdirs >=2.6.0, <4.0.0; sys_platform != 'darwin'  # for others: 2.6+ works consistently.
```
Did you mean these lines?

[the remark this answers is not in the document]
ThomasWaldmann: yes.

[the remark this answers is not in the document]
ganeshrevadi: Done!

ThomasWaldmann (approved):

real-yfprojects (changes requested):
Can you migrate all usages of the config vars to using the pathlib API instead of `os.path`?

ganeshrevadi:
There are around 23 occurrences of  `os.path` , do you want me to change them all?

real-yfprojects:
> There are around 23 occurrences of  `os.path` , do you want me to change them all? 

If you want... I only had those related to the config vars in mind.

ganeshrevadi:
I'm a little confused here , can you clarify and what are these files that consist config vars?

real-yfprojects:
For `LOG_DIR` it is:
https://github.com/borgbase/vorta/blob/a00ed62e491919bd9dc342d61579c8f0f17bc417/src/vorta/log.py#L26 
https://github.com/borgbase/vorta/blob/a00ed62e491919bd9dc342d61579c8f0f17bc417/src/vorta/views/misc_tab.py#L27
Though the latter doesn't need adjustment.

For `SETTINGS_DIR` it would be
https://github.com/borgbase/vorta/blob/a00ed62e491919bd9dc342d61579c8f0f17bc417/src/vorta/__main__.py#L61 and so on.

ganeshrevadi:
There are 4 occurrences of `os.path` with config vars out of which I migrated 3 of them to pathlib API , one not modified was *misc_tab.py*. Do let me know if anything was missed!

src/vorta/application.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import os
 import sys
+from pathlib import Path, PurePath
```
Hofer-Julian, line 4: Why do you use PurePath instead of Path here?

ganeshrevadi: According to the [documentation](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module) this is best possible replacement of `os.path.join` and I think it works fine with `PurePath` as well. If there is any problem we can always use `Path.join`

Hofer-Julian: Yes, I think there is no real problem now with `PurePath`.

But you will also find this in the same doc page:
> If you’ve never used this module before or just aren’t sure which class is right for your task, [Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) is most likely what you need. It instantiates a [concrete path](https://docs.python.org/3/library/pathlib.html#concrete-paths) for the platform the code is running on.

I think it would be better if you would change `PurePath` to `Path`

ganeshrevadi: Done!

Hofer-Julian (commented):
Just minor things, looks good otherwise

src/vorta/__main__.py
```diff
@@ -58,7 +59,7 @@ def exception_handler(type, value, tb):
 
     # Init database
     sqlite_db = SqliteDatabase(
-        os.path.join(SETTINGS_DIR, 'settings.db'),
+        Path.joinpath(SETTINGS_DIR, 'settings.db'),
```
real-yfprojects, line 62:
```suggestion
        SETTINGS_DIR / 'settings.db',
```

src/vorta/__main__.py
```diff
@@ -58,7 +58,7 @@ def exception_handler(type, value, tb):
 
     # Init database
     sqlite_db = SqliteDatabase(
-        os.path.join(SETTINGS_DIR, 'settings.db'),
+        SETTINGS_DIR / 'settings.db',
```
real-yfprojects, line 61:
```suggestion
        str(SETTINGS_DIR / 'settings.db'),
```

real-yfprojects (commented):
And tests setup is failing.

ganeshrevadi:
> And test setup is failing.

how can that be solved?

diivi:
> how can that be solved?

the error:
https://github.com/borgbase/vorta/actions/runs/4367086852/jobs/7638084074#step:11:5467

real-yfprojects:
Have a look at the test output. `APP_ID` has to have type _str_.
```py
____________________ ERROR at setup of test_prune_intervals ____________________

tmpdir_factory = TempdirFactory(_tmppath_factory=TempPathFactory(_given_basetemp=None, _trace=<pluggy._tracing.TagTracerSub object at 0x7f13222b4ad0>, _basetemp=PosixPath('/tmp/pytest-of-runner/pytest-0')))

    @pytest.fixture(scope='session')
    def qapp(tmpdir_factory):
        # DB is required to init QApplication. New DB used for every test.
        tmp_db = tmpdir_factory.mktemp('Vorta').join('settings.sqlite')
        mock_db = SqliteDatabase(str(tmp_db))
        vorta.store.connection.init_db(mock_db)
    
        from vorta.application import VortaApp
    
        VortaApp.set_borg_details_action = MagicMock()  # Can't use pytest-mock in session scope
        VortaApp.scheduler = MagicMock()
    
>       qapp = VortaApp([])  # Only init QApplication once to avoid segfaults while testing.

tests/conftest.py:54: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/vorta/application.py:45: in __init__
    super().__init__(APP_ID, args_raw)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <vorta.application.VortaApp object at 0x7f132241aaf0>
id = PosixPath('/home/runner/.cache/Vorta/tmp/socket'), argv = ([],)

    def __init__(self, id, *argv):
    
        super().__init__(*argv)
        self._id = id
    
        # Is there another instance running?
        self._outSocket = QLocalSocket()
>       self._outSocket.connectToServer(self._id)
E       TypeError: arguments did not match any overloaded call:
E         connectToServer(self, name: str, mode: Union[QIODevice.OpenMode, QIODevice.OpenModeFlag] = QIODevice.ReadWrite): argument 1 has unexpected type 'PosixPath'
E         connectToServer(self, mode: Union[QIODevice.OpenMode, QIODevice.OpenModeFlag] = QIODevice.ReadWrite): argument 1 has unexpected type 'PosixPath'

src/vorta/qt_single_application.py:49: TypeError
```

ganeshrevadi:
will this work?
@real-yfprojects

real-yfprojects:
> will this work? @real-yfprojects

Yes. However I think `APP_ID` itself should be `str` since it is an id instead of file system location.

src/vorta/application.py
```diff
@@ -194,7 +194,7 @@ def check_darwin_permissions(self):
         This function tries reading a file that is known to be restricted and warn the user about
         incomplete backups.
         """
-        test_path = Path.expanduser('~/Library/Cookies')
+        test_path = Path.expanduser(PosixPath('~/Library/Cookies'))
```
ganeshrevadi, line 197: I think this might work!

real-yfprojects: Why? It won't help fixing the tests.

real-yfprojects:
```suggestion
        test_path = Path('~/Library/Cookies').expanduser()
```

will do the trick. `expanduser` is an instance method not a static method.

src/vorta/application.py
```diff
@@ -194,7 +194,7 @@ def check_darwin_permissions(self):
         This function tries reading a file that is known to be restricted and warn the user about
         incomplete backups.
         """
-        test_path = Path.expanduser(PosixPath('~/Library/Cookies'))
+        test_path = Path('~/Library/Cookies').expanduser()
         if Path.exists(test_path) and not os.access(test_path, os.R_OK):
```
real-yfprojects, line 198: Same here: `Path.exists` is a instance method.

src/vorta/application.py
```diff
@@ -195,7 +195,7 @@ def check_darwin_permissions(self):
         incomplete backups.
         """
         test_path = Path('~/Library/Cookies').expanduser()
-        if Path.exists(test_path) and not os.access(test_path, os.R_OK):
+        if Path(test_path).exists() and not os.access(test_path, os.R_OK):
```
Hofer-Julian, line 198:
```suggestion
        if test_path.exists() and not os.access(test_path, os.R_OK):
```

ThomasWaldmann (dismissed):
looks good except for the out-of-scope whitespace and i18n changes.

Hofer-Julian (dismissed):

real-yfprojects (approved):

Issue #1618, opened 2023-02-28
FR: A clear button for the log messages in vorta's application window.


diivi:
#### The problem
When I get an error in the log message (for an unhandled exception maybe), I would like the option to clear this message and continue using vorta normally.
Example:
![image](https://user-images.githubusercontent.com/41837037/221920744-952e4baa-9665-4c02-b972-575ec4c50a35.png)
![image](https://user-images.githubusercontent.com/41837037/221922188-21b5938a-c970-47f3-81e2-c5eae1d04836.png)
More interesting errors here^ (the squashing of other components)

Here, a clear logs button would make the window look better.

#### Requested Solution
The disabled cancel button getting converted to a "Clear" button?

#### Alternatives
The alternative right now is probably just waiting for another backup to finish which might not be considered as good UX.

m3nu:
There should be an open issue to catch and report such errors in a more elegant way.

thesakshidiggikar:
Hi, I would like to work on this issue if it’s still available.

I’ve set up the project locally and started exploring the log display component. I’ll look into how the current log_text widget is implemented and propose a clean way to add a “Clear” button.

Please let me know if I can proceed.

thesakshidiggikar:
I noticed PR #2345 improved log display. 
Just to confirm — is adding a dedicated "Clear" button still desired for this issue?

Meetali7380:
Hi, I would like to work on this issue. Could I take it?

m3nu:
There was a recent PR to truncate the output. So no button needed. Thanks for the interest though.

PR #1626, opened 2023-03-05
Clear contents of `log_text` after successfull backup

### LINKED ISSUE #1389: Clear contents of `logText` after successfull backup.
Currently the label `logText` is not cleared after a borg command finished. When creating a backup the label will show the path of the last backuped file even after backup completion.

Instead the `progressText` and `logText` labels should be cleared if they show information only relevant during the command execution.

aadityasinha-dotcom:
fixes #1389


diivi:
Just wanted to add here that personally, I'd like to see the backup finished message (as a success feedback) and then maybe click on a "Done" or "Clear" button to clear this text. This is also why I opened #1618

DaffyTheDuck:
> Just wanted to add here that personally, I'd like to see the backup finished message (as a success feedback) and then maybe click on a "Done" or "Clear" button to clear this text. This is also why I opened #1618. With the current approach, the user can easily miss the "backup finished" text.

Or I guess the other approach could be like adding some time constraint, after that the message will automatically disappear if the backup is successful, if there's failure then we could add a clear button to clear the log as you mentioned in #1618 because having a time constraint there will be a problem

aadityasinha-dotcom:
Yep we can add time constraint

aadityasinha-dotcom:
For me clear button can be a little hassle, we can add the clear button when we get some failure other wise it will show successful message for few seconds 😊

real-yfprojects:
> With the current approach, the user can easily miss the "backup finished" text.

Are you talking about this PR or the current master branch?
I am not in favour of the timed approach because I could easily miss a message when I am away from the computer. My intention of #1389 was only clearing the log for progress/status updates that aren't relevant after the borg job has finished. This would include showing the added files when creating a backup which looks like this:

```
/home/user/Documents/new.txt (A)
```

src/vorta/borg/create.py
```diff
@@ -36,6 +36,8 @@ def process_result(self, result):
                 self.app.backup_progress_event.emit(self.tr('Backup finished with warnings. See logs for details.'))
             else:
                 self.app.backup_progress_event.emit(self.tr('Backup finished.'))
+                self.app.backup_progress_event.emit(self.tr(''))
+                self.app.backup_log_event.emit('', {})
```
real-yfprojects, line 40:
```suggestion
                self.app.backup_log_event.emit('', {})
```

I imagine that to do the trick. Is there any log event emitted with vital information?

real-yfprojects: > Is there any log event emitted with vital information?

?

aadityasinha-dotcom: no

real-yfprojects (dismissed):

src/vorta/borg/create.py
```diff
@@ -42,8 +42,9 @@ def process_result(self, result):
                     ).format(LOG_DIR.as_uri())
                 )
             else:
+                self.app.backup_progress_event.emit(self.tr('Backup finished.'))
+                self.app.backup_log_event.emit('', {})
                 self.app.backup_progress_event.emit(f"[{self.params['profile_name']}] {self.tr('Backup finished.')}")
-
```
real-yfprojects, line 46: There must be an empty line between two methods.

real-yfprojects (changes requested):

m3nu (dismissed):

real-yfprojects (approved):

Issue #1379, opened 2022-08-03
Append profile information in the status messages


herrwusel:
This is a feature request not a bug :)

I use multiple profiles with vorta and often the backups run at the same time. So it would be a nice addition to have the status in the bottom of the gui with more information. In the picture below it is not clear from which profile the message originates:
![image](https://user-images.githubusercontent.com/8242787/182555982-c9e021cd-4fb1-49a8-807c-6b2078e7bbd0.png)

diivi:
Any decision on how this should be done?
How about `ERROR (Profile Name): Connection closed by remote host. Is borg working on the server?`

real-yfprojects:
What about other messages? They can't be associated to a profile either.

diivi:
Yeah, I meant adding the profile in parenthesis to the beginning of all messages in general, my bad.

herrwusel:
First of all, thanks for having a look a this. I would prefer the following pattern: ` (Profile Name) Log level: log message`

diivi:
Thinking of editing these lines in `main_window.py` then:
```python
def set_log(self, text=''):
  self.logText.setText(text)
  self.logText.repaint()
```
to concatenate the `text` after the profile name.
What do you think @real-yfprojects ?

real-yfprojects:
> I would prefer the following pattern: ` (Profile Name) Log level: log message`

Square or round brackets?



> Thinking of editing these lines in `main_window.py` then:
> 
> ```python
> def set_log(self, text=''):
>   self.logText.setText(text)
>   self.logText.repaint()
> ```
> 
> to concatenate the `text` after the profile name.

How do you know from which profile a log was sent?

diivi:
> How do you know from which profile a log was sent?

Oh, yeah, then this should probably be done in the places where we use `self.app.backup_log_event.emit`. 
Just checked `borg_job.py` and I think the profile info is available there right?

real-yfprojects:
> Just checked `borg_job.py` and I think the profile info is available there right?

I think so.

herrwusel:
> Square or round brackets?

Definitely square brackets

diivi:
@Hofer-Julian this can be closed now :)

herrwusel:
Thanks!

PR #1637, opened 2023-03-08
feat: add profile name to log messages

diivi:
### Description
Append profile name to every log message

### Related Issue
#1379 

### Motivation and Context
Helps the user know which profile the log is coming from.

### How Has This Been Tested?
image

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -46,11 +46,11 @@ class VortaApp(QtSingleApplication):
         if self.isRunning():
             if single_app:
                 self.sendMessage("open main window")
-                print('An instance of Vorta is already running. Opening main window.')
+                logger.info('An instance of Vorta is already running. Opening main window.')
                 sys.exit()
             elif args.profile:
                 self.sendMessage(f"create {args.profile}")
-                print('Creating backup using existing Vorta instance.')
+                logger.info('Creating backup using existing Vorta instance.')
                 sys.exit()
         elif args.profile:
             sys.exit('Vorta must already be running for --create to work')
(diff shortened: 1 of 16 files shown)


real-yfprojects:
This would also be useful for the other messages like `Backup finished`.

real-yfprojects:
> I did this for `logText` (the second one) only, should I do it for the first one too, i.e. `progressText` (all occurences of `backup_progress_event.emit`)?

Yes, since currently one has no way to tell which backup has finished.

diivi:
> Yes, since currently one has no way to tell which backup has finished.

So why not do this just for the Backup Finished message instead of every progress event?
Sorry, don't have much context about long running background backups.

real-yfprojects:
> So why not do this just for the Backup Finished message instead of every progress event?

Doesn't it also apply to other progress messages? None of them state the profile they are originating from.

diivi:
Hey, I need some help understanding how I can get the profile name in the various jobs in vorta. For example the create job:

```python
    def process_result(self, result):
```
Should I access `result[params][profile_name]` to get the profile name?
Also, what about the functions which have a signature like this: (no result)
```
    def started_event(self):
```
How do I get the profile name to display in the progress event emitted by these functions?
Or should I skip these?

real-yfprojects:
This could be cleaner and better but you can always access the profile from a BorgJob with 
```py
self.params['profile_id']
```

real-yfprojects:
I think you have to resolve the merge conflicts before the tests are run.

real-yfprojects (approved):
I don't know a better way to use the same message template over and over again with a qt signal -- besides maybe writing a function.

PR #1639, opened 2023-03-09
 Tests leave key in keyring

ganeshrevadi:
Adding `remove password` abstract method in *abc.py* and implementing in *secretstorage.py*.
@real-yfprojects

### DIFF
--- a/tests/test_repo.py
+++ b/tests/test_repo.py
@@ -107,7 +107,7 @@ def test_repo_add_success(qapp, qtbot, mocker, borg_json_output, keyring_fixture
 
     assert RepoModel.get(id=2).url == test_repo_url
 
-    assert long_password in keyring.get_password("vorta-repo", RepoModel.get(id=2).url)
+    assert long_password.lower() in keyring.get_password("vorta-repo", RepoModel.get(id=2).url).lower()
     assert main.repoTab.repoSelector.currentText() == test_repo_url
 
 


src/vorta/keyring/secretstorage.py
```diff
@@ -65,6 +65,24 @@ def get_password(self, service, repo_url):
...
+            asyncio.set_event_loop(asyncio.new_event_loop())
+            attributes = {
+                'application': 'Vorta',
+                'service': service,
+                'repo_url': repo_url,
+            }
+            items = list(self.collection.search_items(attributes))
+            logger.debug('Found %i passwords matching repo URL.', len(items))
+            for item in items:
+                self.collection.delete_item(item)
```
real-yfprojects, line 82: Does this method even exist? Secondly, i think we also have to check whether the item is unlocked before we can delete it.

real-yfprojects (commented):
Looking good! Next up would be KWallet. We use the dbus daemon exposed by KWallet. Refer to the [Qt DBus docs](https://doc.qt.io/qt-5/qtdbus-index.html) how to use dbus via qt. You will have to do some research to find a documentation of set kwallet daemon.

src/vorta/keyring/secretstorage.py
ganeshrevadi: My bad its not *delete_item* its just [*delete()*](https://secretstorage.readthedocs.io/en/latest/_modules/secretstorage/collection.html#Collection.delete) . We have already checked if the item is unlocked in `if self.is_unlocked:`

real-yfprojects: Nope, `is_unlocked` only checks whether the collection is unlocked while it is possible with some key managers that the items are locked individually as well. This is why we added a second test in `get_password`.

ganeshrevadi: So basically i have to check `if self.is_unlocked` again in the for loop ?

real-yfprojects: See
https://github.com/borgbase/vorta/blob/abb3ec82ee1de301355965e4d46e21c4e69064d2/src/vorta/keyring/secretstorage.py#L62-L63

src/vorta/keyring/kwallet.py
```diff
@@ -45,6 +45,11 @@ def get_password(self, service, repo_url):
         logger.debug(f"Retrieved password for repo {repo_url}")
         return password
 
+    def remove_password(self, service, repo_url):
+        if self.is_unlocked and self.get_result("hasEntry", args=[self.handle, self.folder_name, repo_url, service]):
+            self.get_result("removeEntry", args=[self.handle, self.folder_name, repo_url, service])
```
real-yfprojects, line 50:
```suggestion
        entry = [self.handle, self.folder_name, repo_url, service]
        if self.is_unlocked and self.get_result("hasEntry", args=entry):
            self.get_result("removeEntry", args=entry)
```

real-yfprojects (commented):
Great, it works. Where did you find the `removeEntry` dbus method? Next up would be MacOS support which is quite hard without access to a Mac. We use the [PyObj-C bridge](https://pyobjc.readthedocs.io/en/latest/) allowing access to MacOS native libraries. You can refer to the [apple developer docs](https://developer.apple.com/documentation/security/keychain_services/keychain_items/updating_and_deleting_keychain_items?language=objc).

m3nu:
I'll try macOS today.

m3nu:
Added for macOS:

```
In [1]: from vorta.keyring import darwin

In [2]: k = darwin.VortaDarwinKeyring()

In [3]: k.get_password('vorta', 'google.com')

In [4]: k.set_password('vorta', 'google.com', 'asdf')

In [5]: k.get_password('vorta', 'google.com')
Out[5]: 'asdf'

In [6]: k.remove_password('vorta', 'google.com')

In [7]: k.get_password('vorta', 'google.com')

In [8]: 
```

real-yfprojects:
That's bad you just removed the commits containing MacOS support.

ganeshrevadi:
> That's bad you just removed the commits containing MacOS support.

Hey I'm really sorry how can this be fixed now?

m3nu:
Rebased and pushed it again.

Please be careful with force-push and use it only when really needed.

real-yfprojects:
And when you do use `--force-with-lease`.

ganeshrevadi:
> And when you do use `--force-with-lease`.

when you update a branch changed by other people

ganeshrevadi:
should `remove_password` next implemented in *db.py*

src/vorta/keyring/db.py
```diff
@@ -33,6 +33,16 @@ def get_password(self, service, repo_url):
         except peewee.DoesNotExist:
             return None
 
+    def remove_password(self, service, repo_url):
+        from vorta.store.models import RepoPassword
```
real-yfprojects, line 37: Is there anything I overlooked that makes importing in the method this necessary? I would prefer having imports at the beginning of the file.

ganeshrevadi: I did the importing inside the method as it was the same in `set_password `and `get_password` , I don't know what's the reason !

m3nu:
For macOS-related stuff it's needed because some libraries aren't available on linux (pyobjc).

In this case it could be due to circular dependencies. Needs to be reviewed. If no issues, put it at the start of the file, where it belongs.

ganeshrevadi:
> For macOS-related stuff it's needed because some libraries aren't available on linux (pyobjc).
> 
> In this case it could be due to circular dependencies. Needs to be reviewed. If no issues, put it at the start of the file, where it belongs.

I tried adding it above but its failing the pre-commit run!

real-yfprojects:
> I tried adding it above but its failing the pre-commit run!

As always, nobody can help you if you just say that it failed. You will have to provide more details. The error/complaint message is a minimum.

ganeshrevadi:
I have imported the `RepoPassword` at beginning of the file and it works fine and sorry for the lack of info in above comment.

src/vorta/keyring/db.py
```diff
@@ -33,6 +30,15 @@ def get_password(self, service, repo_url):
...
             return None
 
+    def remove_password(self, service, repo_url):
+
+        try:
+            keyring_entry = RepoPassword.get(url=repo_url)
+            keyring_entry.delete_instance()
+            logger.debug(f"Removed password for repo {repo_url}")
+        except peewee.DoesNotExist:
+            pass
```
real-yfprojects, line 40: Let's log that the password wasn't found.

src/vorta/keyring/secretstorage.py
```diff
@@ -65,6 +65,26 @@ def get_password(self, service, repo_url):
...
+                'application': 'Vorta',
+                'service': service,
+                'repo_url': repo_url,
+            }
+            items = list(self.collection.search_items(attributes))
+            logger.debug('Found %i passwords matching repo URL.', len(items))
+            for item in items:
+                if item.is_locked() and item.unlock():
+                    return None
+                self.collection.delete(item)
```
real-yfprojects, line 84: This deletes the whole collection. Use `item.delete`.

src/vorta/keyring/secretstorage.py
```diff
@@ -65,6 +65,26 @@ def get_password(self, service, repo_url):
...
+            attributes = {
+                'application': 'Vorta',
+                'service': service,
+                'repo_url': repo_url,
+            }
+            items = list(self.collection.search_items(attributes))
+            logger.debug('Found %i passwords matching repo URL.', len(items))
+            for item in items:
+                if item.is_locked() and item.unlock():
+                    return None
```
real-yfprojects, line 83:
```suggestion
                    continue
```

src/vorta/keyring/secretstorage.py
```diff
@@ -65,6 +65,26 @@ def get_password(self, service, repo_url):
...
+                self.collection.delete(item)
+                logger.debug(f"Removed password for repo {repo_url}")
+        return None
```
real-yfprojects, line 86:
```suggestion
```

ganeshrevadi:
> You can use the newly implemented method now to remove passwords saved in `test_utils.py` and `test_repo.py`.

Can you please explain about `test_repo.py`?

m3nu:
> Can you please explain about `test_repo.py`?

Well, the main purpose of this PR was to avoid having hundreds of old passwords in our keychain from tests. Since our tests added a random repo and password, but never removed it. So you can now adjust the tests to remove the password when it's done. Best to see the pytest docs on *where* to best do this cleanup. The cleanup should also happen if the test fails.

m3nu:
https://github.com/borgbase/vorta/blob/master/tests/test_repo.py#L68
https://github.com/borgbase/vorta/blob/master/tests/test_repo.py#L84

m3nu:
Looks good. I just have these concerns:

- The assert to test if the password is gone isn't really needed all the time. Just enough to test it *one time* where the keyring is tested. You wouldn't test everything all the time. One unit, one test (unit test). (or integration test if you test clicking buttons, etc)
- What happens if the test fails in the middle? Then the password is never removed. For this reason pytest uses cleanup functions and fixtures. Would be better to use those. I'm not sure what's the best way in this case. Best to quickly consult the pytest docs.

m3nu:
Not quite.

- The keyring instance you create in the fixture isn't used in the test and Vorta will always make its own.
- You wouldn't need to remove the password inside tests, if the fixture was working. `keyring.remove_password('vorta-repo', test_repo_url)`
- Your fixture runs at the module level. So all tests in one file would use the same password and remove it??

So what you need is this:
- Fixture that runs again for each test
- Fixture that returns some random repo URL and random password for testing, as we do in some places: `test_repo_url = f'vorta-test-repo.{uuid.uuid4()}.com:repo'  # Random repo URL to avoid macOS keychain` and `password = str(uuid.uuid4())`
- The repo URL and password are then used during testing
- After each test *function* the password for the test repo is removed again.

This should be more clear and deduplicates some more code (creating a random repo URL and password)

One more thing: `vorta-test-repo.{uuid.uuid4()}.com:repo` would give a SFTP-style URL that's not supported in the future. Best if you change it to `ssh://` format.

m3nu:
Also be sure to rebase your branch sometimes, since the master branch will keep changing when a PR is open for longer.

ganeshrevadi:
> Not quite.
> 
> * The keyring instance you create in the fixture isn't used in the test and Vorta will always make its own.
> * You wouldn't need to remove the password inside tests, if the fixture was working. `keyring.remove_password('vorta-repo', test_repo_url)`
> * Your fixture runs at the module level. So all tests in one file would use the same password and remove it??
> 
> So what you need is this:
> 
> * Fixture that runs again for each test
> * Fixture that returns some random repo URL and random password for testing, as we do in some places: `test_repo_url = f'vorta-test-repo.{uuid.uuid4()}.com:repo'  # Random repo URL to avoid macOS keychain` and `password = str(uuid.uuid4())`
> * The repo URL and password are then used during testing
> * After each test _function_ the password for the test repo is removed again.
> 
> This should be more clear and deduplicates some more code (creating a random repo URL and password)
> 
> One more thing: `vorta-test-repo.{uuid.uuid4()}.com:repo` would give a SFTP-style URL that's not supported in the future. Best if you change it to `ssh://` format.

This means that the repo url and password that are generated in the test has to be removed and only to be generated in the fixture and then used in multiple tests . Same question for `remove_password` ?

m3nu:
Repo URL and password are generated by the fixture. The password manager entry is also removed by the fixture during cleanup.

ganeshrevadi:
@m3nu I've tried implementing pytest fixture as per your suggestion , but the tests are failing. Need some help here!

tests/test_repo.py
```diff
@@ -116,8 +110,6 @@ def test_repo_add_success(qapp, qtbot, mocker, borg_json_output, keyring_fixture
     keyring = VortaKeyring.get_keyring()
     assert keyring.get_password("vorta-repo", RepoModel.get(id=2).url) == LONG_PASSWORD
```
ganeshrevadi, line 110: gives me a assertion error here.

m3nu: In that case you should debug the test and see *why* it fails. Looking at the compared values would be a first good step.

real-yfprojects:
Are you still working on this @ganeshrevadi? This would be a big quality of life improvement for developers.

ganeshrevadi:
@real-yfprojects Hey I'm currently not working on this!

PR #1656, opened 2023-03-16
feat: block after vorta create and log status

### LINKED ISSUE #1380: vorta --create should block and return borg exit status
**Describe the bug**

Currently `vorta --create Profile` signals Vorta to start Borg and exits immediately. By doing so it cannot signal back when the backup is finished and whether it finished successfully. I'm thinking that `vorta --create` should be blocking during Borg run and return its exit status to make it more informative overall and to provide blocking ability for scripts that need it. Users that don't want it can still fork the process rather easily. Or an additional switch could be added to only signal Vorta and exit immediately (as is the current behavior), `vorta --create --start Profile` for example.

- OS: Arch Linux
- Vorta version: 0.8.7
- Installed from: OS repository

diivi:
### Description
Fixes #1380 
blocks `vorta --create` to log the result of the backing up operation.

### Motivation and Context
Better UX for `vorta --create`

### How Has This Been Tested?
Test GIF:
create-block

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -52,7 +52,6 @@ class VortaApp(QtSingleApplication):
             elif args.profile:
                 self.sendMessage(f"create {args.profile}")
                 logger.info('Creating backup using existing Vorta instance.')
-                sys.exit()
         elif args.profile:
             sys.exit('Vorta must already be running for --create to work')
 
@@ -147,6 +146,9 @@ class VortaApp(QtSingleApplication):
     def message_received_event_response(self, message):
         if message == "open main window":
             self.open_main_window_action()
+        elif message.startswith("created"):
+            logger.debug(f"Backup created: {message[8:]}")
+            sys.exit()
         elif message.startswith("create"):
             message = message[7:]  # Remove create
             if self.jobs_manager.is_worker_running():
--- a/src/vorta/borg/create.py
+++ b/src/vorta/borg/create.py
@@ -45,6 +45,7 @@ class BorgCreateJob(BorgJob):
         self.app.backup_progress_event.emit(self.tr('Backup started.'))
 
     def finished_event(self, result):
+        self.app.reply(f"created {result['data']['archive']['name']}")
         self.app.backup_finished_event.emit(result)
         self.result.emit(result)
         self.pre_post_backup_cmd(self.params, cmd='post_backup_cmd', returncode=result['returncode'])
--- a/src/vorta/qt_single_application.py
+++ b/src/vorta/qt_single_application.py
@@ -53,6 +53,7 @@ class QtSingleApplication(QApplication):
             # Yes, there is.
             self._outStream = QTextStream(self._outSocket)
             self._outStream.setCodec('UTF-8')
+            self._outSocket.readyRead.connect(self._onReadyReply)
         else:
             # No, there isn't.
             self._outSocket = None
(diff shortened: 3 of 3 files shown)


src/vorta/borg/create.py
```diff
@@ -45,6 +45,7 @@ def started_event(self):
         self.app.backup_progress_event.emit(self.tr('Backup started.'))
 
     def finished_event(self, result):
+        self.app.reply(f"created {result['data']['archive']['name']}")
```
diivi, line 48: The tests are failing because even during a normal backup I try to call `self.app.reply()`, which I should not. @real-yfprojects can you help me understand how I can check if the backup creation was triggered from the command line or normally from the GUI, so I can call `reply` optionally here?

real-yfprojects: You can connect a handle to `BorgJob.result` when triggered from the cmd command.

real-yfprojects (commented):
There is a problem: A second main window opens while the backup is running.

diivi:
> There is a problem: A second main window opens while the backup is running.

What do you suggest I do for the second window? If I close it, will it run in the background until it exits?

real-yfprojects:
> What do you suggest I do for the second window?

Don't create a main window instance (nor a scheduler instance and so on).

diivi:
> Don't create a main window instance (nor a scheduler instance and so on).
I changed the code so it only adds the message_received event to the other instance and nothing else.

src/vorta/application.py
```diff
@@ -147,6 +152,9 @@ def backup_cancelled_event_response(self):
     def message_received_event_response(self, message):
         if message == "open main window":
             self.open_main_window_action()
+        elif message.startswith("created"):
+            logger.debug(f"Backup created: {message[8:]}")
```
real-yfprojects, line 156: I think level `info` would fit better.

real-yfprojects (commented):
I notice that this can't handle when borg returns an error or when `vorta --create` is run twice simultaneously

diivi:
> I notice that this can't handle when borg returns an error or when `vorta --create` is run twice simultaneously

My last commit should handle borg errors right?

Also, how should I handle the second case? After handling Borg errors, I get this error when I run vorta --create  twice simultaneously:
`QAbstractSocket::waitForBytesWritten() is not allowed in UnconnectedState`

real-yfprojects:
> After handling Borg errors, I get this error when I run vorta --create twice simultaneously:
> `QAbstractSocket::waitForBytesWritten() is not allowed in UnconnectedState`

I see this being outputted too, but it doesn't stop code execution. I think the current networking code can only handle one client at a time resulting in this message when the first (or maybe the second) client has disconnected.

src/vorta/application.py
```diff
@@ -52,49 +52,50 @@ def __init__(self, args_raw, single_app=False):
             elif args.profile:
                 self.sendMessage(f"create {args.profile}")
                 logger.info('Creating backup using existing Vorta instance.')
-                sys.exit()
         elif args.profile:
             sys.exit('Vorta must already be running for --create to work')
 
-        init_translations(self)
+        self.message_received_event.connect(self.message_received_event_response)
+        if not args.profile:
```
m3nu, line 59: `if True` is generally quicker to read than `if not XX`. And since we have a large code block under this condition now, it should get at least a comment on *when* this path is chosen. Still, keeping most of the code on the default path and dealing with exceptions some other way is preferable.

Issue #1655, opened 2023-03-16
Settings for rejecting WIFIs by default and for showing `WIFI not allowed` messages.


real-yfprojects:
Add two settings called something along the lines of `Allow WiFis by default` and `Show notification when WiFi is disallowed`.

### Discussed in https://github.com/borgbase/vorta/discussions/1654

<div type='discussions-op-text'>

<sup>Originally posted by **bcelary** March 16, 2023</sup>
My Vorta runs daily backups when I am at home to my local NAS server. But I often move around and any other WiFi should not be used to even attempt a backup.

Would you consider adding an option that would require user to explicitly add a WiFi to allowed networks? Especially when using local address (like 192.168... or host.local name)? Also the message that WiFi is not allowed might become a non-error message, as Vorta keeps nagging me all the time when I am away from home.

These options could be added in the `Networks` panel, here:
![image](https://user-images.githubusercontent.com/376943/225550787-63a7e8ce-477a-4cc4-a605-94b83d9e4cbf.png)

E.g. `x Don't allow new WiFi connections` and `x silence "cuurent WiFi is not allowed" messages`.

Asking if such idea makes sense as this seems like something easy to add and I might as well create a PR for this.</div>

diivi:
Ah, I didn't know vorta checked every network you connected to by default. @bcleary I'd be happy to work on this for you, I'm trying to learn more about the project. Thanks!

diivi:
@real-yfprojects You can assign this to me, I have an approach ready.

bcelary:
wow, things move really fast in here! thank you @diivi :pray:

PR #1658, opened 2023-03-17
feat: settings to allow new networks and show notif when a network is disallowed

diivi:
Adds 2 columns to the profile model:
1. shows the "This wifi is not allowed" notification only when it is checked. Shows the notif by default.
2. adds new networks to allowed networks only when checked. Does not add by default.

### Related Issue
#1655 

### Motivation and Context
#1654 

### How Has This Been Tested?
Manually tested.

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -113,12 +113,16 @@ class VortaApp(QtSingleApplication):
             job = BorgCreateJob(msg['cmd'], msg, profile.repo.id)
             self.jobs_manager.add_job(job)
         else:
-            notifier = VortaNotifications.pick()
-            notifier.deliver(
-                self.tr('Vorta Backup'),
-                translate('messages', msg['message']),
-                level='error',
-            )
+            if not (
+                msg['message'] == 'Current Wifi is not allowed.'
+                and not profile.show_notification_when_network_disallowed
+            ):
+                notifier = VortaNotifications.pick()
+                notifier.deliver(
+                    self.tr('Vorta Backup'),
+                    translate('messages', msg['message']),
+                    level='error',
+                )
             self.backup_progress_event.emit(translate('messages', msg['message']))
             return None
 
--- a/src/vorta/store/migrations.py
+++ b/src/vorta/store/migrations.py
@@ -226,6 +226,22 @@ def run_migrations(current_schema, db_connection):
             migrator.add_column(SettingsModel._meta.table_name, 'tooltip', pw.CharField(default='')),
         )
 
+    if current_schema.version < 21:
+        _apply_schema_update(
+            current_schema,
+            21,
+            migrator.add_column(
+                BackupProfileModel._meta.table_name,
+                'allow_new_networks',
+                pw.BooleanField(default=False),
+            ),
+            migrator.add_column(
+                BackupProfileModel._meta.table_name,
+                "show_notification_when_network_disallowed",
+                pw.BooleanField(default=True),
+            ),
+        )
+
 
 def _apply_schema_update(current_schema, version_after, *operations):
     with DB.atomic():
--- a/src/vorta/store/models.py
+++ b/src/vorta/store/models.py
@@ -95,6 +95,8 @@ class BackupProfileModel(BaseModel):
     pre_backup_cmd = pw.CharField(default='')
     post_backup_cmd = pw.CharField(default='')
     dont_run_on_metered_networks = pw.BooleanField(default=True)
+    allow_new_networks = pw.BooleanField(default=False)
+    show_notification_when_network_disallowed = pw.BooleanField(default=True)
 
     def refresh(self):
         return type(self).get(self._pk_expr())
--- a/src/vorta/views/schedule_tab.py
+++ b/src/vorta/views/schedule_tab.py
@@ -75,6 +75,12 @@ class ScheduleTab(ScheduleBase, ScheduleUI, BackupProfileMixin):
         self.meteredNetworksCheckBox.stateChanged.connect(
             lambda new_val, attr='dont_run_on_metered_networks': self.save_profile_attr(attr, not new_val)
         )
+        self.allowNewNetworksCheckBox.stateChanged.connect(
+            lambda new_val, attr='allow_new_networks': self.save_profile_attr(attr, new_val)
+        )
+        self.networkDisallowedNotificationCheckBox.stateChanged.connect(
+            lambda new_val, attr='show_notification_when_network_disallowed': self.save_profile_attr(attr, new_val)
+        )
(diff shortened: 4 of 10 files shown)


src/vorta/views/schedule_tab.py
```diff
@@ -149,6 +155,12 @@ def populate_from_profile(self):
         self.meteredNetworksCheckBox.setChecked(
             QtCore.Qt.Unchecked if profile.dont_run_on_metered_networks else QtCore.Qt.Checked
         )
+        self.allowNewNetworksCheckBox.setChecked(
+            QtCore.Qt.Checked if profile.allow_new_networks else QtCore.Qt.Unchecked
+        )
```
real-yfprojects, line 160:
```suggestion
        self.allowNewNetworksCheckBox.setChecked(
            profile.allow_new_networks
        )
```

`setChecked` expects a boolean. Therefore the if isn't needed.

src/vorta/application.py
```diff
@@ -113,12 +113,16 @@ def create_backup_action(self, profile_id=None):
...
-            notifier = VortaNotifications.pick()
-            notifier.deliver(
-                self.tr('Vorta Backup'),
-                translate('messages', msg['message']),
-                level='error',
-            )
+            if not (
+                msg['message'] == 'Current Wifi is not allowed.'
+                and not profile.show_notification_when_network_disallowed
+            ):
```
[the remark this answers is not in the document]
diivi: Your suggestion is more readable to me, thanks.

src/vorta/store/migrations.py
```diff
@@ -226,6 +226,22 @@ def run_migrations(current_schema, db_connection):
...
         )
 
+    if current_schema.version < 21:
+        _apply_schema_update(
+            current_schema,
+            21,
+            migrator.add_column(
+                BackupProfileModel._meta.table_name,
+                'allow_new_networks',
+                pw.BooleanField(default=False),
```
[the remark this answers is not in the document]
diivi: You mean true? 
Also, I think not using new networks is a safer bet and better for users, but let me know.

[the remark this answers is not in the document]
real-yfprojects: > You mean true?

Yes

> I think not using new networks is a safer bet and better for users

We can discuss the default for new users in https://github.com/borgbase/vorta/blob/28c30f18b1a804ec06800753e214f016712109ae/src/vorta/store/models.py#L98
but the migration shouldn't change the behaviour the user is expecting.

Not allowing new networks is safer but some users might wonder why their backups don't run on WiFi. With *simple by default, powerful when needed* in mind I advocate a default of `True`.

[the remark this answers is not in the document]
diivi: Makes sense, I'll make `True` the default value.

real-yfprojects (approved):
LGTM. I didn't test the new notification setting.

src/vorta/application.py
```diff
@@ -113,12 +113,13 @@ def create_backup_action(self, profile_id=None):
             job = BorgCreateJob(msg['cmd'], msg, profile.repo.id)
             self.jobs_manager.add_job(job)
         else:
-            notifier = VortaNotifications.pick()
-            notifier.deliver(
-                self.tr('Vorta Backup'),
-                translate('messages', msg['message']),
-                level='error',
-            )
+            if msg['message'] != 'Current Wifi is not allowed.' or profile.show_notification_when_network_disallowed:
```
m3nu, line 116: Comparing the precise error message duplicates it and is prone to breaking, if it slightly changes later?

m3nu:
I feel like "Show notification when a network is disallowed" duplicates the option that's already there under *Misc* - "Display notification when background tasks fails". If you look at notifications and backups in general:

1. It may succeed: we have a notification option for that
2. It may fail for some random reason: we have a notification for that too
3. It may fail due to disallowed network: this PR

The question is if 2 and 3 are different enough to have their own option. Or if 3 is just some other failure. In any case, the option is better suited under *Misc* as it's so similar to the others.

diivi:
@bcelary what do you think (about the difference between 2 and 3)? Since you requested this in the first place.

Otherwise, yeah, maybe this should be under Misc.

src/vorta/application.py
diivi: Do you suggest adding a numeric value to the msg dict? Like a code or something? Vorta does not use consistent return codes, so I did not find anything.

bcelary:
Seems like misc is the right place to keep it consistent. It looks like this is really a different category of a message that is not really an error but like an info. Similarly there could be an error that there is no available network at all which seems pointless because user is likely aware of this (not sure if there's such a message but just as an example).

Caveat here that I'm not an UX expert.

real-yfprojects:
I think 2 and 3 can be merged to one notification.

diivi:
> I think 2 and 3 can be merged to one notification.

So I should drop the second part of this PR entirely, and @bcelary can use the `Misc - Display notification when background tasks fails` setting to disable notifications?

bcelary:
I realize that too many options is not ideal so I'm definitely good with whatever you decide.

Maybe some criticality should be added to notifications so that user would be able to decide the level of severity but that would require a new feature. I think the most sensible is to simple drop that part indeed.

diivi:
What should I do about this - https://github.com/borgbase/vorta/pull/1658#discussion_r1158378665?

real-yfprojects:
> What should I do about this - [#1658 (comment)](https://github.com/borgbase/vorta/pull/1658#discussion_r1158378665)?

Didn't we discuss that we treat *blocked wifi* like any other error?

real-yfprojects:
Can you upgrade to `PyQt6` (#1685)?

src/vorta/store/models.py
```diff
@@ -97,6 +97,7 @@ class BackupProfileModel(BaseModel):
     pre_backup_cmd = pw.CharField(default='')
     post_backup_cmd = pw.CharField(default='')
     dont_run_on_metered_networks = pw.BooleanField(default=True)
+    allow_new_networks = pw.BooleanField(default=False)
```
m3nu, line 100: Here it defaults to `False`, but in migrations it's `True`? Suggest to use `True` everywhere, which is the current behavior.

m3nu:
This is almost ready.

src/vorta/store/models.py
diivi: I think yf and I discussed this before (been long so idr exactly where, I tried searching). The new behaviour should not surprise existing users, so the migration changes nothing. But for new users, having this value as false is more beneficial.

diivi: found it - https://github.com/borgbase/vorta/pull/1658#discussion_r1140417145

PR #1662, opened 2023-03-18
Introduced password input widget

jetchirag:
### Description
Added a new **PasswordInput** widget for PyQt to improve functionality and reusability.

Not completed and looking for feedback.

### Related Issue

> I was thinking that we could use a common widget (a self-implemented one) for all password entries in vorta. Can you implement such thing?

_Originally posted by @real-yfprojects in https://github.com/borgbase/vorta/issues/1659#issuecomment-1474039819_
            

image
visiblity toggle:
image
validation error (min length):
image

### Types of changes

### Checklist:


jetchirag:
image

UI is mostly the same.
image

real-yfprojects:
Nice. Maybe we can reduce duplicated code even more by having a widget that unifies the two password lineEdits and the related labels. This would involve these steps:

- [ ] Rename `PasswordInput` to something like `PassswordLineEdit`
- [ ] Make it subclass `QLineEdit`
- [ ] Add and implement a password input class consisting of two `PasswordLineEdits`
- [ ] Move all validation logic/rules into the new class
- [ ] Use it in existing ui layouts

jetchirag:
> Add and implement a password input class consisting of two PasswordLineEdits

So a QWidget? What do you think about a `link_wth` parameter in same class to link confirm input with first one? We can still move validation to this class and use features of QLineEdit too.

jetchirag:
Can you please take a look at the use of translation?

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,127 @@
...
+            msg = trans_late('utils', "Passwords must be identical and greater than 8 characters long.")
+        elif not pass_equal:
+            self.confirmLineEdit.set_error_state(True)
+            msg = trans_late('utils', "Passwords must be identical.")
+        elif not pass_long:
+            self.confirmLineEdit.set_error_state(True)
+            msg = trans_late('utils', "Passwords must be greater than 8 characters long.")
+
+        # fire signal
+        self.passwordLabelChanged.emit(msg)
```
real-yfprojects, line 125: The contents of the label don't have to have changed.

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,127 @@
...
+            form_layout.addRow(self._label_2, self.confirmLineEdit)
+            form_layout.addRow(self.password_label)
+        else:
+            password_form = QFormLayout(self)
+            password_form.setContentsMargins(0, 0, 0, 0)
+            password_form.addRow(self._label_1, self.passwordLineEdit)
+            password_form.addRow(self._label_2, self.confirmLineEdit)
+            password_form.addRow(self.password_label)
+
+            password_form.setFieldGrowthPolicy(QFormLayout.AllNonFixedFieldsGrow)
```
real-yfprojects, line 80: Let's move this into a method, so that we can instantiate the widget and add it to a form later.

real-yfprojects (commented):
This is looking like what I had in mind. How Vorta uses the password inputs makes it hard to move it into its own class. Although inheriting `QWdiget` would be nice, it doesn't make much sense for this use. I think inheriting `QObject` would be better. This makes `PasswordInput` a manager of widgets instead of a widget itself.

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,127 @@
...
+        if self._show_visibility_button:
+            self.showHideAction = QAction(self.tr("Show password"), self)
+            self.showHideAction.setCheckable(True)
+            self.showHideAction.toggled.connect(self._toggle_password_visibility)
+            self.showHideAction.setIcon(get_colored_icon("eye"))
+            self.addAction(self.showHideAction, QLineEdit.TrailingPosition)
+
+    def get_password(self):
+        """Return password text"""
+        return self.text()
```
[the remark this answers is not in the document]
jetchirag: @real-yfprojects I added it so we can call get_password on single password field (PasswordLineEdit) as well as PasswordInput. Mainly a hack to call same method in 
https://github.com/borgbase/vorta/blob/ac7de01645b0ab064fd80eb2543092472a5adee9/src/vorta/views/repo_add_dialog.py#L71

Should I conditionally add `password` using `__class__`?

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def _toggle_password_visibility(self):
+        if self.echoMode() == QLineEdit.Password:
+            self.setEchoMode(QLineEdit.Normal)
+            self.showHideAction.setIcon(get_colored_icon("eye"))
```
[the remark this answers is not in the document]
jetchirag: Since it will never be connected if `_show_visibility_button` is false, it shouldn't raise an error

https://github.com/borgbase/vorta/blob/ac7de01645b0ab064fd80eb2543092472a5adee9/src/vorta/views/partials/password_input.py#L17-L20

Do you want to make `show_visibility_button` or `_toggle_password_visibility` public?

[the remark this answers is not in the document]
jetchirag: Ignore the above comment. I somehow missed the review above.

src/vorta/views/partials/password_input.py
jetchirag: Do we need this signal? I've removed it for now.

src/vorta/views/partials/password_input.py
real-yfprojects: :+!:

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if self._show_visibility_button:
+            self.showHideAction.setIcon(get_colored_icon("eye"))
+        else:
+            self.showHideAction.setIcon(get_colored_icon("eye-slash"))
+
+    @property
+    def visible(self):
+        """Return password visibility"""
+        return self._visible
```
real-yfprojects, line 41: Can add a setter?

src/vorta/views/partials/password_input.py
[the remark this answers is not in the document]
real-yfprojects: I see.

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,137 @@
...
+            self._label_password = QLabel(label[0])
+            self._label_confirm = QLabel(label[1])
+        else:
+            self._label_password = QLabel(self.tr("Enter passphrase:"))
+            self._label_confirm = QLabel(self.tr("Confirm passphrase:"))
+
+        # Create password line edits
+        self.passwordLineEdit = PasswordLineEdit()
+        self.confirmLineEdit = PasswordLineEdit()
+        self.password_label = QLabel("")
```
real-yfprojects, line 74: I think e.g. `validation_label` would fit better.

src/vorta/views/repo_add_dialog.py
```diff
@@ -27,28 +27,29 @@ def __init__(self, parent=None):
         self.saveButton = self.buttonBox.button(QDialogButtonBox.StandardButton.Ok)
         self.saveButton.setText(self.tr("Add"))
 
+        if self.__class__ == AddRepoWindow:
+            self.passwordInput = PasswordInput()
+            self.repoDataFormLayout.addRow(self.passwordInput.create_form_widget())
+            self.passwordInput.passwordLineEdit.textChanged.connect(self.password_listener)
+            self.passwordInput.confirmLineEdit.textChanged.connect(self.password_listener)
```
real-yfprojects, line 34: I think `password_listener` should be connected to the encryption combobox too.

src/vorta/views/repo_add_dialog.py
```diff
@@ -27,28 +27,29 @@ def __init__(self, parent=None):
         self.saveButton = self.buttonBox.button(QDialogButtonBox.StandardButton.Ok)
         self.saveButton.setText(self.tr("Add"))
 
+        if self.__class__ == AddRepoWindow:
```
real-yfprojects, line 30: This is hacky. Is it possible to remove the PasswordInput from the layout in a subclass?

src/vorta/views/partials/password_input.py
jetchirag: Yes

src/vorta/views/partials/password_input.py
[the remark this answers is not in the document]
real-yfprojects: This is going in the right direction. However you made a logical error. This still fails if `_show_visibility_button` is false.

src/vorta/views/partials/password_input.py
real-yfprojects: What do you think?

src/vorta/views/repo_add_dialog.py
```diff
@@ -27,28 +27,27 @@ def __init__(self, parent=None):
...
         self.repoURL.textChanged.connect(self.set_password)
-        self.passwordLineEdit.textChanged.connect(self.password_listener)
-        self.confirmLineEdit.textChanged.connect(self.password_listener)
-        self.encryptionComboBox.activated.connect(self.display_backend_warning)
 
-        # Add clickable icon to toggle password visibility to end of box
-        self.showHideAction = QAction(self.tr("Show my passwords"), self)
-        self.showHideAction.setCheckable(True)
-        self.showHideAction.toggled.connect(self.set_visibility)
+        if self.__class__ == AddRepoWindow:
```
real-yfprojects, line 39: This is not clean code conforming to the standards of OOP. However the current class hierachy doesn't allow for a different solution. I therefore suggest moving the code for adding the password widgets into a subclass so that `AddRepoWindow` and `ExistingRepoWindow` share a common ancestor with the common code.

jetchirag: That would make more sense. Do you want me to push these changes here or open a new PR after this?

src/vorta/views/partials/password_input.py
jetchirag: Yes, that sounds better.

src/vorta/views/repo_add_dialog.py
jetchirag: Regarding the logical error involving `_show_visibility_button`, I think you reviewed an older commit. _toggle_password_visibility it was replaced with `toggle_visibility` public method.

src/vorta/views/repo_add_dialog.py
```diff
@@ -196,43 +135,82 @@ def validate(self):
...
-            self.showHideAction.setText(self.tr("Hide my password"))
-        else:
-            self.showHideAction.setIcon(get_colored_icon("eye"))
-            self.showHideAction.setText(self.tr("Show my password"))
+
+        # Remove existing passwordinput from parent class
+        passwordInput_index = self.repoDataFormLayout.indexOf(self.passwordInput.create_form_widget())
+        if passwordInput_index != -1:
+            self.repoDataFormLayout.removeRow(passwordInput_index)
+        del self.passwordInput
```
real-yfprojects, line 202: You don't have to remove it if you don't add it in the parent class

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,146 @@
...
+        """Set password visibility"""
+        if not isinstance(value, bool):
+            raise TypeError("visible must be a boolean value")
+        self._visible = value
+        self.setEchoMode(QLineEdit.Normal if self._visible else QLineEdit.Password)
+
+        if self._show_visibility_button:
+            self.showHideAction.setIcon(get_colored_icon("eye"))
+        else:
+            self.showHideAction.setIcon(get_colored_icon("eye-slash"))
```
real-yfprojects, line 44: This is going in the right direction. However you made a logical error. This still fails if `_show_visibility_button` is false.

src/vorta/views/repo_add_dialog.py
real-yfprojects: > Regarding the logical error involving `_show_visibility_button`, I think you reviewed an older commit. _toggle_password_visibility it was replaced with `toggle_visibility` public method.

No, I only added the comment to the existing conversation but I copied the comment to the new code lines now.

src/vorta/views/partials/password_input.py
jetchirag: My bad. I mistook `_show_visibility_button` for `_visible`.

src/vorta/views/repo_add_dialog.py
```diff
@@ -196,43 +132,80 @@ def validate(self):
...
-            return not bool(msg)
+            self.passwordInput.set_validation_enabled(True)
+
+    def display_backend_warning(self):
+        '''Display password backend message based off current keyring'''
+        if self.encryptionComboBox.currentData() != 'none':
+            self.passwordInput.set_error_label(VortaKeyring.get_keyring().get_backend_warning())
+
+    def run(self):
+        if self.validate() and self.passwordInput.validate():
```
real-yfprojects, line 180: Wouldn't it make sense to include that check in `self.validate()`?

src/vorta/views/repo_add_dialog.py
```diff
@@ -246,3 +219,13 @@ def run(self):
...
+
+    @property
+    def values(self):
+        out = dict(
+            ssh_key=self.sshComboBox.currentData(),
+            repo_url=self.repoURL.text(),
+            password=self.passwordInput.get_password(),
+            extra_borg_arguments=self.extraBorgArgumentsLineEdit.text(),
+        )
+        return out
```
real-yfprojects, line 231: This could be moved to the superclass because it's common to both subclasses.

src/vorta/views/repo_add_dialog.py
```diff
@@ -196,43 +132,80 @@ def validate(self):
...
+            else:
+                self._set_status(params['message'])
 
 
-class ExistingRepoWindow(AddRepoWindow):
+class ExistingRepoWindow(RepoWindow):
     def __init__(self):
         super().__init__()
         self.encryptionComboBox.hide()
         self.encryptionLabel.hide()
```
real-yfprojects, line 196: These widgets aren't used in this subclass. Therefore it would make more sense not to add them in the superclass.

src/vorta/views/repo_add_dialog.py
jetchirag: I added them to subclass because of difference in dict items. `ExistingRepoWindow` does not use `encryption` value as there's no such widget. Would it more sense to include it conditionally using a private variable?

real-yfprojects: You can override the method in the subclass but still leverage the super class implementation through `super().values()`.

m3nu:
What's missing to move this forward? This would be useful to encapsulate this functionality better.

jetchirag:
@m3nu I wasn't able to complete above reviews due to academic exams. Will commit them today.

src/vorta/views/repo_add_dialog.py
jetchirag: I didn't add it because `passwordInput.validate()` is not available in `ExistingRepoWindow`. I can either a conditional check for type of passwordInput in base class to call it or implement a `validate()` method in `PasswordLineEdit` if you want to move it to `validate()`

src/vorta/views/repo_add_dialog.py
jetchirag: Thank you, that was helpful.

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,147 @@
...
+            self._label_password = QLabel(self.tr("Enter passphrase:"))
+            self._label_confirm = QLabel(self.tr("Confirm passphrase:"))
+
+        # Create password line edits
+        self.passwordLineEdit = PasswordLineEdit()
+        self.confirmLineEdit = PasswordLineEdit()
+        self.validation_label = QLabel("")
+
+        self.passwordLineEdit.editingFinished.connect(self.validate)
+        self.confirmLineEdit.textChanged.connect(self.validate)
```
real-yfprojects, line 85: Why did you decide to connect to `editingFinished` for the first entry?

jetchirag: So it won't show error and red borders while user is still typing first password. Would textChanged make more sense?

real-yfprojects: > it won't show error and red borders while user is still typing first password.

That makes sense. However after typing a password that is too short the red border won't go away (immediately) when extending the password to over 8 letters.

jetchirag: @real-yfprojects What do you propose? We can either connect `textChanged` instead or perhaps replace it after first edit?

```
self.passwordLineEdit.editingFinished.connect(self.on_editing_finished)

def on_editing_finished(self):
    self.passwordLineEdit.editingFinished.disconnect()
    self.passwordLineEdit.textChanged.connect(self.validate)
    self.validate()
```

real-yfprojects (commented):
Can you add unittests for `password_input.py` where viable?

jetchirag:
Where should I define the test? Do I create a new test file?

Also I assume you mean writing tests for just the class, unrelated to any dialog like AddRepoWindow, by setting up new instance and testing methods?

real-yfprojects:
> Where should I define the test? Do I create a new test file?

Yes, you do.

> Also I assume you mean writing tests for just the class, unrelated to any dialog like AddRepoWindow, by setting up new instance and testing methods? 

Exactly. Unittests try to test as little code lines as possible/viable, so that bugs can be easily tracked down to a specific unit.

jetchirag:
I've pushed the tests for PasswordLineEdit. Can you check if I'm going in right direction?

tests/test_password_input.py
```diff
@@ -0,0 +1,61 @@
...
+    password_line_edit = PasswordLineEdit()
+    qtbot.addWidget(password_line_edit)
+    assert password_line_edit.error_state is False
+
+    password_line_edit.error_state = True
+    assert password_line_edit.error_state
+    assert password_line_edit.styleSheet() == "border: 2px solid red;"
+
+
+def test_password_lint_visibility_button(qtbot):
```
diivi, line 37: `password_line_visibility_button`?

diivi: The tests are good; coverage is at 91% and will be easy to increase.

tests/test_password_input.py
```diff
@@ -0,0 +1,61 @@
...
+    qtbot.addWidget(password_line_edit)
+    assert password_line_edit is not None
+
+
+def test_password_line_get_password(qtbot):
+    password_line_edit = PasswordLineEdit()
+    qtbot.addWidget(password_line_edit)
+
+    password_line_edit.setText("test")
+    assert password_line_edit.get_password() == "test"
```
real-yfprojects, line 15: Let's also test whether the password is empty before setting it.

tests/test_password_input.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_password_line_visible(qtbot):
+    password_line_edit = PasswordLineEdit()
+    qtbot.addWidget(password_line_edit)
+    assert password_line_edit.visible is False
+
+    password_line_edit.toggle_visibility()
+    assert password_line_edit.visible
```
real-yfprojects, line 24: Why do you use `is` inconsistently here?

real-yfprojects (commented):
Yes this is right. Can you add some more tests for PasswordInput? We might be able to reach nearly 100% coverage for this file.

tests/test_password_input.py
jetchirag: Thank you, fixed.

tests/test_password_input.py
jetchirag: There were some pep warning regarding use of `x == True:` instead of `x:`. 

I've changed the code to
```
    assert not password_line_edit.visible

    password_line_edit.toggle_visibility()
    assert password_line_edit.visible
```

jetchirag:
I'm running into some error regarding PasswordInput

```
def test_password_input_get_password(qtbot):
    password_input = PasswordInput()
    qtbot.addWidget(password_input.create_form_widget())

    assert password_input.get_password() == ""

    password_input.passwordLineEdit.setText("test")
    assert password_input.get_password() == "test"
```

```
self = <vorta.views.partials.password_input.PasswordInput object at 0x7f1201f31510>

    def get_password(self) -> str:
>       return self.passwordLineEdit.text()
E       RuntimeError: wrapped C/C++ object of type PasswordLineEdit has been deleted       
```

Garbage collector seems to be deleting our object. Perhaps because they don't have set parent but I'm not sure how to go about it.

diivi:
btw, Dropping 
```python
qtbot.addWidget(password_input.create_form_widget())
```
from subsequent tests makes it work.

tested with
```python
def test_password_input_get_password(qtbot):
    password_input = PasswordInput()

    assert password_input.get_password() == ""

    password_input.passwordLineEdit.setText("test")
    assert password_input.get_password() == "test"

def test_password_input_set_label(qtbot):
    password_input = PasswordInput(label=["test", "test2"])
    
    assert password_input._label_password.text() == "test"
    assert password_input._label_confirm.text() == "test2"
```

jetchirag:
> btw, Dropping
> ```
> qtbot.addWidget(password_input.create_form_widget())
> ```

I think in that case, it won't be added to the application and qtbot wouldn't work.

I've set the parent to main_window now and added more tests

```
$ coverage run --source=vorta.views.partials.password_input -m pytest -v tests/test_password_input.py && coverage report -m
...

Name                                         Stmts   Miss  Cover   Missing
--------------------------------------------------------------------------
src/vorta/views/partials/password_input.py     114      3    97%   88-90
--------------------------------------------------------------------------
TOTAL                                          114      3    97%
                                                                   
```

`on_editing_finished()` is missing

jetchirag:
Updated for PyQt6.
@m3nu Can you please review this once?

src/vorta/views/partials/password_input.py
```diff
@@ -0,0 +1,166 @@
...
+
+        return not bool(self.validation_label.text())
+
+    def add_form_to_layout(self, form_layout: QFormLayout) -> None:
+        """Adds form to layout"""
+        form_layout.addRow(self._label_password, self.passwordLineEdit)
+        form_layout.addRow(self._label_confirm, self.confirmLineEdit)
+        form_layout.addRow(self.validation_label)
+
+    def create_form_widget(self, parent=None):
```
real-yfprojects, line 158:
```suggestion
    def create_form_widget(self, parent: Optional[QWidget] = None) -> QWidget:
```

real-yfprojects:
@jetchirag Can you resolve the merge conflicts?

jetchirag:
@real-yfprojects I pushed a commit but still don't see it in this PR. Perhaps Github is glitching (getting frequent 500 errors).

real-yfprojects (dismissed):

m3nu:
Hey, thanks for removing some duplicate code here. Just tried it locally and noticed this:

- The first time someone reveals the password, the icon doesn't match the state. (Icon and tooltip says "Show password", but password is already shown)
- For the case of *Existing Repo*, the spacing at the bottom could be more consistent with the *New Repo* case.

Screenshot 2023-06-09 at 12 28 53

jetchirag:
Fixed in new commits. Also resolved merge conflicts.

m3nu:
The show/hide icon state works as expected now.

Just for the "Add Existing Repo" dialog, there may be a fixed height set, which squeezes the lines a bit. I know we wanted to reduce the extra space a bit, but this needs to be done without settings a fixed height on the dialog window, but by editing the ?vertical space element within it. (I know Qt can be fickle).

Screenshot 2023-07-05 at 11 21 30

jetchirag:
Ah, it doesn't happen on OS and UI I'm testing on. I'll fix that.

jetchirag:
@m3nu Without changing fixed height, do you want to expand the inner widget?

image

I enabled `verstretch`  for this.

jetchirag:
Sorry I'm still unsure what you are proposing. I don't think inner widget has a fixed height but because main layout has fixed height, it's expanding. The fixed height seems to be based on Add new repo so can we use a lower fixed height for this dialog?

https://github.com/borgbase/vorta/commit/92608f9eaaa6ac9d4bf28f9fa59edaae6001fe07#diff-804dae53d409307e28853c1c80251c625c13f04bd7b0037ef30dc99fe347a77dR10

real-yfprojects:
If possible we shouldn't use fixed heights when creating a GUI.

jetchirag:
Okay so I've removed the fixed size property and added some margin. I've also set a min width otherwise existing repo dialog would be quite smaller than the new repo dialog. Let me know if I understood it correctly.

real-yfprojects:
Sounds good, I'll test it when I find the time.

real-yfprojects (commented):
Very nice! One slight problem though:
Screenshot_20230721_165455
This only happens when the border colour is altered.

jetchirag:
Ah, fixing it.

jetchirag:
Fixed the red border. Is that white text an issue with this widget or the theme?

real-yfprojects:
> Fixed the red border. Is that white text an issue with this widget or the theme?

Both is fixed now. On my desktop the text is white and the background black.

real-yfprojects (approved):

Issue #1435, opened 2022-10-16
FR: Quick Browse


Justinzobel:
#### The problem
The Mount option is good but requires the user to have already created an empty folder to mount to. This seems to be a little bit of extra work for the basic user.

#### Requested Solution
A Quick Browse button. This button would create a folder for the user, for example in /tmp or in a known Borg location e.g. ~/.borg and mount the selected Archive for the user. This way a user can quickly access the backup and browser for files. The button would then change to an Unmount button which would unmount the temporary directory and clean up.

real-yfprojects:
Thanks for opening the FR. We were already had the idea for such a feature #1281.

I think this feature should include opening the folder in the file explorer automatically as well as a setting to disable this behaviour.

diivi:
I have a few questions regarding this:
1. what should be the directory where Vorta does a "Quick Mount" and open it later?
2. Should quick mount be a new button or a setting, which when enabled, the user won't have to select a folder when pressing the mount button everytime.

Justinzobel:
1. It should likely be mounted in the user's home as vorta backups are done by the user (in comparison to mounting in /tmp or /mnt as they are publicly accessible). Vorta could create a temporary directory in the user's home folder e.g. ~/Vorta-Restore-{randomcharacters} and mount it there.

2. I think it should be a button, Vorta is a GUI application to make the backup and restore of user data as simple and easy as possible. Alternatively, the Mount button could default to a quick mount and a dropdown next to it could allow users to specify the location if they desire. For example like the GitHub button below the comment box:


![image](https://user-images.githubusercontent.com/1370117/226072262-aa3281cb-b940-40d0-9583-aa0aad5a9648.png)

real-yfprojects:
I like the dropdown solution.

diivi:
> I like the dropdown solution.

Could you point me to the right direction (qt docs?), If I were to implement something like that?
I would also like a quick mount feature.

jetchirag:
@diivi I think you can use [QMenu](https://doc.qt.io/qt-6/qmenu.html) for UI

https://github.com/borgbase/vorta/blob/1f1278270df5dd452ffaa3646912f2520387a9c7/src/vorta/views/repo_tab.py#L29-L34

https://github.com/borgbase/vorta/blob/1f1278270df5dd452ffaa3646912f2520387a9c7/src/vorta/assets/UI/repotab.ui#L144-L151

PR #1664, opened 2023-03-19
feat: quick mount

diivi:
### Description
Adds a Quick Mount feature to vorta so a user can browse their backups quickly. Improves UX as it will no longer be required to create a folder separately, open it in vorta, and then mount.

### Related Issue
#1435 

### Motivation and Context
UX Improvement and an old feature request.

### How Has This Been Tested?
Tested with repos for now. The old functionality is "Mount to Folder" and the new one is "Quick Mount". This option opens the created folder and deletes it when the user unmounts the repository.

Quick Mount

### What's Left:
- [ ] Quick mount for archives
- [ ] Updating/Adding Tests

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -1,9 +1,13 @@
 import logging
+import os
+import random
+import shutil
+import string
 import sys
 from datetime import timedelta
 from typing import Dict, Optional
 from PyQt5 import QtCore, uic
-from PyQt5.QtCore import QItemSelectionModel, QMimeData, QPoint, Qt, pyqtSlot
+from PyQt5.QtCore import QItemSelectionModel, QMimeData, QPoint, Qt, QUrl, pyqtSlot
 from PyQt5.QtGui import QDesktopServices, QKeySequence
 from PyQt5.QtWidgets import (
     QAction,
@@ -69,6 +73,8 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         self.app = app
         self.toolBox.setCurrentIndex(0)
         self.repoactions_enabled = True
+        self.first_refresh = True
+        self.quick_mount = False
 
         #: Tooltip dict to save the tooltips set in the designer
         self.tooltip_dict: Dict[QWidget, str] = {}
@@ -118,7 +124,12 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         self.bPrune.clicked.connect(self.prune_action)
         self.bCheck.clicked.connect(self.check_action)
         self.bDiff.clicked.connect(self.diff_action)
-        self.bMountRepo.clicked.connect(self.bmountrepo_clicked)
+        self.menuMountRepo = QMenu(self.bMountRepo)
+
+        self.menuMountRepo.addAction(translate("ArchiveTab", "Mount to Folder"), self.bmountrepo_clicked)
+        self.menuMountRepo.addAction(translate("ArchiveTab", "Quick Mount"), self.quick_mount_action)
+
+        self.bMountRepo.setMenu(self.menuMountRepo)
 
         self.archiveNameTemplate.textChanged.connect(
             lambda tpl, key='new_archive_name': self.save_archive_template(tpl, key)
@@ -248,6 +259,9 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
             self.mount_points, repo_mount_points = get_mount_points(profile.repo.url)
             if repo_mount_points:
                 self.repo_mount_point = repo_mount_points[0]
+                mount_point_filename = os.path.basename(self.repo_mount_point)
+                if mount_point_filename.startswith("vorta-quick-mount-"):
+                    self.quick_mount = True
 
             self.toolBox.setItemText(0, self.tr('Archives for %s') % profile.repo.url)
             archives = [s for s in profile.repo.archives.select().order_by(ArchiveModel.time.desc())]
@@ -519,6 +533,33 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         else:
             self.mount_action(archive_name=archive_name)
 
+    def quick_mount_action(self, archive_name=None):
+        """
+        create a temporary directory in the user's home folder ~/vorta-quick-mount-{randomcharacters} and
+        mount the selected archive  or repository to that directory.
+        """
+        profile = self.profile()
+        params = BorgMountJob.prepare(profile, archive=archive_name)
+        if not params['ok']:
+            self._set_status(params['message'])
+            return
+
+        mount_point = os.path.join(
+            os.path.expanduser('~'),
+            'vorta-quick-mount-' + ''.join(random.choices(string.ascii_lowercase + string.digits, k=6)),
+        )
+        os.mkdir(mount_point)
+
+        params['cmd'].append(mount_point)
+        params['mount_point'] = mount_point
+
+        if params['ok']:
+            self._toggle_all_buttons(False)
+            job = BorgMountJob(params['cmd'], params, self.profile().repo.id)
+            job.updated.connect(self.mountErrors.setText)
+            job.result.connect(lambda result: self.mount_result(result, quick=True))
+            self.app.jobs_manager.add_job(job)
+
     def bmountrepo_clicked(self):
         """
         Handle `bMountRepo` being clicked.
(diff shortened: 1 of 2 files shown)


src/vorta/views/archive_tab.py
```diff
@@ -248,6 +259,9 @@ def populate_from_profile(self):
             self.mount_points, repo_mount_points = get_mount_points(profile.repo.url)
             if repo_mount_points:
                 self.repo_mount_point = repo_mount_points[0]
+                mount_point_filename = os.path.basename(self.repo_mount_point)
+                if mount_point_filename.startswith("vorta-quick-mount-"):
```
diivi, line 263: Assuming this should be safe right?  The other option is to delete the quick mount folder when vorta exits.

src/vorta/views/archive_tab.py
```diff
@@ -517,7 +540,37 @@ def bmountarchive_clicked(self):
         if archive_name in self.mount_points:
             self.unmount_action(archive_name=archive_name)
         else:
-            self.mount_action(archive_name=archive_name)
+            if quick:
```
real-yfprojects, line 543:
```suggestion
        elif quick:
```

src/vorta/views/archive_tab.py
```diff
@@ -517,7 +540,37 @@ def bmountarchive_clicked(self):
...
+        params = BorgMountJob.prepare(profile, archive=archive_name)
+        if not params['ok']:
+            self._set_status(params['message'])
+            return
+
+        mount_point = os.path.join(
+            os.path.expanduser('~'),
+            'vorta-quick-mount-' + ''.join(random.choices(string.ascii_lowercase + string.digits, k=6)),
+        )
+        os.mkdir(mount_point)
```
real-yfprojects, line 563: You should check whether the folder already exists/is empty.

src/vorta/views/archive_tab.py
```diff
@@ -550,6 +603,25 @@ def bmountarchive_refresh(self, icon_only=False):
...
 
+    def bmountarchive_update_menu(self):
+        """
+        Update the menu of `bMountArchive`.
+        """
+        archive_name = self.selected_archive_name()
+        if archive_name in self.mount_points:
+            self.bMountArchive.setMenu(None)
+            try:
+                self.bMountArchive.clicked.disconnect()
```
real-yfprojects, line 614: Please only disconnect the specific slot.

src/vorta/views/archive_tab.py
```diff
@@ -550,6 +603,25 @@ def bmountarchive_refresh(self, icon_only=False):
...
+                self.bMountArchive.clicked.connect(self.bmountarchive_clicked)
+            except TypeError:
```
real-yfprojects, line 616: Can you add a small comment explaining why a type error might be raised?

src/vorta/views/archive_tab.py
```diff
@@ -550,6 +603,25 @@ def bmountarchive_refresh(self, icon_only=False):
                 self.bMountArchive.setText(self.tr("Mount…"))
                 self.bMountArchive.setToolTip(self.tr("Mount the selected archive " + "as a folder in the file system"))
 
+    def bmountarchive_update_menu(self):
```
real-yfprojects, line 606: Why didn't you include that code in `bmount_archive_refresh`?

src/vorta/views/archive_tab.py
```diff
@@ -600,12 +680,17 @@ def receive():
...
+    def mount_result(self, result, quick=False):
         if result['returncode'] == 0:
             self._set_status(self.tr('Mounted successfully.'))
 
             mount_point = result['params']['mount_point']
 
+            if quick:
+                # open the folder
+                QDesktopServices.openUrl(QUrl.fromLocalFile(mount_point))
+                self.quick_mount = True
```
real-yfprojects, line 692: This var is related to repo mounts only I think.

src/vorta/views/archive_tab.py
real-yfprojects: I'd consider this safe.

src/vorta/views/archive_tab.py
```diff
@@ -248,6 +263,13 @@ def populate_from_profile(self):
             self.mount_points, repo_mount_points = get_mount_points(profile.repo.url)
             if repo_mount_points:
                 self.repo_mount_point = repo_mount_points[0]
+                mount_point_filename = os.path.basename(self.repo_mount_point)
+                if mount_point_filename.startswith("vorta-quick-mount-"):
+                    self.quick_mount = True
+            for mount_point in self.mount_points:
+                mount_point_filename = os.path.basename(self.mount_points[mount_point])
+                if mount_point_filename.startswith("vorta-quick-mount-"):
+                    self.quick_mount_points[mount_point] = self.mount_points[mount_point]
```
real-yfprojects, line 272: How do you know which repo a mount point belongs to?

real-yfprojects (commented):
All actions that open another dialog e.g. a file dialog must end with an ellisis `…`

src/vorta/views/archive_tab.py
```diff
@@ -64,11 +68,13 @@ def __init__(self, parent=None, app=None):
         super().__init__(parent)
         self.setupUi(parent)
         self.mount_points = {}  # mapping of archive name to mount point
+        self.quick_mount_points = {}  # mapping of archive name to quick mount point
```
m3nu, line 71: Maybe the `self.mount_points` dict can be reused for this. Why a new dict for this? Not possible to add more than one archive from the same repo and the mounts need to be discovered during startup. So tracking them all in one place is better.

src/vorta/views/archive_tab.py
```diff
@@ -665,16 +753,23 @@ def umount_result(self, result):
...
                 item = QTableWidgetItem('')
                 self.archiveTable.setItem(row, 3, item)
 
                 # update button
+                self.bmountarchive_update_menu()
                 self.bmountarchive_refresh()
             else:
                 # unmount repo
                 self.repo_mount_point = None
+                if self.quick_mount:
```
m3nu, line 770: So this variable is only used to decide if the folder get deleted? Why not use the name for that? What if I quick mount from one profile and then normal mount from another? Will it remove that folder too?

src/vorta/views/archive_tab.py
diivi: If I trigger `bmountrepo_refresh` on load, I can handle mounting/unmounting different repos.

src/vorta/views/archive_tab.py
```diff
@@ -544,11 +595,22 @@ def bmountarchive_refresh(self, icon_only=False):
             if not icon_only:
                 self.bMountArchive.setText(self.tr("Unmount"))
                 self.bMountArchive.setToolTip(self.tr('Unmount the selected archive from the file system'))
+                self.bMountArchive.setMenu(None)
+                try:
+                    self.bMountArchive.clicked.disconnect(self.bmountarchive_clicked)  # avoid race condition
+                    self.bMountArchive.clicked.connect(self.bmountarchive_clicked)
+                except TypeError:
+                    self.bMountArchive.clicked.connect(self.bmountarchive_clicked)
```
real-yfprojects, line 603: Why are there two calls to `connect`?

src/vorta/views/archive_tab.py
```diff
@@ -544,11 +595,22 @@ def bmountarchive_refresh(self, icon_only=False):
             if not icon_only:
                 self.bMountArchive.setText(self.tr("Unmount"))
                 self.bMountArchive.setToolTip(self.tr('Unmount the selected archive from the file system'))
+                self.bMountArchive.setMenu(None)
+                try:
+                    self.bMountArchive.clicked.disconnect(self.bmountarchive_clicked)  # avoid race condition
```
real-yfprojects, line 600: Would `bmountarchive_clicked` be connected to the signal multiple times?
This also doesn't disconnect the slot `lambda: self.bmountarchive_clicked(quick=True)` afaik.

src/vorta/views/archive_tab.py
```diff
@@ -366,6 +379,7 @@ def on_selection_change(self, selected=None, deselected=None):
 
             # special treatment for dynamic mount/unmount button.
             self.bmountarchive_refresh()
+            self.bmountrepo_refresh()
```
real-yfprojects, line 382: Why is that needed? I guess it doesn't hurt either.

diivi: To support changing the repository from the dropdown in the`Repository` tab. Without this, the Unmount option is visible for all repositories when I restart Vorta and not just the one I mounted.

real-yfprojects: I think this would make more sense in `_toggle_all_buttons`

m3nu:
Already has pending feedback.

real-yfprojects:
What does assigning me mean @m3nu ?

m3nu:
It means that I'll try to make sure those get merged and don't stay in limbo.

diivi:
it means I'll work on it and you'll review it, probably.

m3nu:
> What does assigning me mean @m3nu ?

I misread, @real-yfprojects . Since you already commented a good bit here, I assigned you to push this over the finishing line. So it doesn't stay in limbo, as before.

PR #1666, opened 2023-03-21
ci: add ruff for print checks

diivi:
### Description
Recently committed a print statement, which reminded me of https://github.com/borgbase/vorta/pull/1612#issuecomment-1447782606

image

### Types of changes

### Checklist:


setup.cfg
diivi, line 1: line 42: `paramiko` should probably be removed here now?

pyproject.toml
```diff
@@ -15,3 +15,6 @@ src_paths = ["src", "tests"]
 [build-system]
 requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
+
+[tool.ruff]
+select = ["T"]
```
Hofer-Julian, line 20: In order to see if flake8 can replace flake8 and isort mid-term, let's activate all flake8 and isort features.

.github/workflows/test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       run: |
         pip install .
         pip install -r requirements.d/dev.txt
-    - name: Test formatting with Flake8, isort and Black
+    - name: Test formatting with Flake8, isort, ruff and Black
       run: make lint
```
Hofer-Julian, line 28: I feel like pre-commit should replace the `make lint` rule.
This can also be handled in a follow up PR.

Hofer-Julian (commented):
Thanks, looks good in general.

Makefile
```diff
@@ -60,6 +60,7 @@ lint:
 	flake8
 	isort --check-only .
```
Hofer-Julian, line 61: Even if we wouldn't want to immediately move away from flake8 as requested by @real-yfprojects, I think we can safely move from isort to ruff.

pyproject.toml
real-yfprojects: *ruff

Hofer-Julian: Ah yes :P

package/fix_app_qt_folder_names_for_codesign.py
real-yfprojects, line 1: Is it possible to limit the print checks to production code?

Hofer-Julian: Yes https://beta.ruff.rs/docs/settings/#exclude

Hofer-Julian: In that case, the status quo is the best option as far as I can see

diivi: https://beta.ruff.rs/docs/settings/#per-file-ignores ?

m3nu: Helper scripts don't need to be included in these checks. Just `src`.

pyproject.toml
diivi: Turns out ruff does not use the same options with isort as isort itself. I could replicate the functionality to limit the changes to a minimum. But I can't figure out what option is triggering changes like these:
image
here are all the options ruff allows configuring - https://beta.ruff.rs/docs/settings/#isort
Also, I had to add a rule to remove line breaks between import sections too, why did we suppress them in the first place? They look better imo:
image

Makefile
real-yfprojects: Having https://github.com/borgbase/vorta/pull/1666#discussion_r1143728743 in mind I suggest keeping this as a check.

pyproject.toml
real-yfprojects: I like them as well. But I guess @m3nu doesn't.

real-yfprojects: > But I can't figure out what option is triggering changes like these

Maybe `PyQt5` isn't recognized as a [third party library](https://beta.ruff.rs/docs/settings/#known-third-party)?

m3nu: Either way is good. If there are many imports, some line breaks can separate the various groups of imports.

m3nu:
What's missing to move this forward?

diivi:
> What's missing to move this forward?

Just a few minor changes:
* keeping isort and not using ruff's isort.
* include only src in checks

pyproject.toml
diivi: I removed isort from the codebase and replaced it with ruff. It added some lines between imports but I think this separation is better.

real-yfprojects:
It would be great if you could seperate your changes from the autogenerated ones. This allow excluding the latter from git blame.

real-yfprojects:
> You mean make all config changes, skip pre-commit and push directly?

Maybe run pre-commit on the config changes before committing those.

> How will the sorting changes be made then?

Are you talking about the changes by *rust*?

> shouldn't git blame just blame me for the empty lines, and the other stuff remains the same?

I didn't look into the exact changes prior to writing my comment. This is just the general process of changing autoformatter configs. However rust didn't only add new lines I think.

diivi:
> rust
ruff 😅 

So you mean I should make 2 commits right - one with my changes and the other one with ruff's changes?

diivi:
Config changes - https://github.com/borgbase/vorta/pull/1666/commits/5d8073b8947b85d9c1ed3ce3a6d707c99893b39c.
I'll handle the merge conflicts after this is checked.

real-yfprojects:
> I'll handle the merge conflicts after this is checked.

Looking good. After resolving merge conflicts (I think rebasing makes the most sense) I would squash the first two commits. Then a commit can be created that adds `run formatter` to `.git-blame-ignore-revs`.

diivi:
Tried doing that, but I think I messed something up here - https://github.com/borgbase/vorta/compare/master...diivi:test-ruff-commits?expand=1.
I'll try to redo the changes in order.

real-yfprojects:
> Tried doing that, but I think I messed something up

1. Make sure you local master is up-to-date with `upstream/master`.
2. Checkout this branch
3. `git rebase -i master`
4. In the file opened select to squash `ca90d7d373e4dfdd14ce2b10017516937bf43c9c` and to drop `a5d433fb252ddb8b0c97b803fc4eba58dd3e161d`
5. Save and close the file
6. Rerun the automatic formatters and commit the changes to the code base
7. Add the hash of the last commit to `.git-blame-ignore-revs` and commit it.

real-yfprojects:
Merging this without the commit containing the formatting changes will break ci afaik.

m3nu:
Wasn't this to find print statements when it started? 😀

And wasn't Black doing the formatting? Can't we just keep the same settings?

Anyways, if everything needs to be reformatted again, I suggest to just add your config changes and apply the formatting later when there aren't many pending PRs. Better to rerun the formatter than deal with all the conflicts.

diivi:
> Wasn't this to find print statements when it started? grinning

Yeah, now it's about replacing isort with ruff too 😅 https://github.com/borgbase/vorta/pull/1666#discussion_r1143116959

m3nu:
Understood. Then I suggest this:

- Please all help testing #1685 (Qt6 upgrade) so we can merge it today or tomorrow. This will start a new major version, since a major dependency changes.
- For this PR, please remove the formatting commits and only keep the commit that changes the tooling config.
- After Qt6 is merged, you can add the formatting changes without worrying about conflicts.
- Since this is a new major version, one could consider expanding the use of Ruff and replacing some more tools, if you like.

m3nu:
#1685 is merged. So good to proceed here and add the formatting changes needed.

real-yfprojects:
Does everyone approve this PR? Then I'll take care of merging this correctly.

m3nu:
Good to go! Formatting change separate, if possible. 🙏

real-yfprojects:
> Formatting change separate

I am planning on fast forwarding master with 3 commits corresponding to the commits of this PR.

Makefile
```diff
@@ -58,8 +58,8 @@ flatpak-install: translations-to-qm
 
 lint:
 	flake8
-	isort --check-only .
 	black --check .
+	ruff check . --fix
```
real-yfprojects, line 62: Why `--fix`? This is intended as a check only. That's why black is run with `--check`.

.pre-commit-config.yaml
```diff
@@ -62,6 +57,11 @@ repos:
     hooks:
       - id: flake8
 
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: 'v0.0.257'
+    hooks:
```
real-yfprojects, line 62: I'd prefer running the pre-commit hook with `--fix`.

Issue #1220, opened 2022-03-13
Add white icon for menu bar MacOS


ph00lt0:
**Describe the bug**
Please add a white icon to conform with the design standard of menu bar icons in Monterey / Big Sur

**Environment (please complete the following information):**
- OS: MacOS
- Vorta version: 0.8.3
- Installed from: Brew

Vorta and Borg versions can be found in Main Window > Misc Tab.

real-yfprojects:
@m3nu You are more familiar with MacOS.

Brawl345:
> Currently it uses an icon that works on dark and light themes

<img width="332" alt="image" src="https://user-images.githubusercontent.com/5049892/210285644-b5fb31de-e7e1-4ad5-97b5-3af39add6cd9.png">

This does not look good though :/

diivi:
I made two [alternatives](https://www.figma.com/file/WHe8SAB7vv8dpnPSC1tNpX/Untitled?node-id=0%3A1&t=L18gUWgduhrReYZP-0).
![image](https://user-images.githubusercontent.com/41837037/225361735-635cd78a-e899-4a18-9ce2-c1740f715bee.png)
I prefer the second one, which one do you think might be better? @m3nu

m3nu:
4 icons are needed in total: dark/light x normal/active

I'm also not sure if Qt gets the signal regarding the theme change, if it's in the background only. That's why I didn't implement this and went for a simple icon that works for all cases. If you can make it work, it would be a nice addition though.

diivi:
> 4 icons are needed in total: dark/light x normal/active
> I'm also not sure if Qt gets the signal regarding the theme change, if it's in the background only. 

I am on a Linux machine running Ubuntu 22.04 so I won't be able to test how Vorta responds to changes to the system's theme on MacOS. But I feel like a transparent body for the hdd icon might be more suitable than the current black body. Let me know what you think, otherwise maybe I'll try to get this running on a friend's Mac later.

real-yfprojects:
The other icons use solid shapes instead of contours. So maybe your first option is better @diivi.

Brawl345:
I would also prefer the first option.

diivi:
Feel free to use the designs I shared, just request access so I can allow you to export the icons!

i1sm3ky:
> I wondered why detecting via QPalette doesn't work. Or maybe it does...

[...]

PR #1676, opened 2023-04-02
Added white icons for GNOME and MacOS

i1sm3ky:
### Description

This PR changes tray icon color based on the system tray color which increases the contrast between the tray icon and the system tray.

### Related Issue

Issue: #1220 requesting for white icons for menu bar on MacOS

### Motivation and Context

This change provides white icons for dark themed Gnome based linux and MacOS which look better as solid white icons instead of the dark black ones.

### How Has This Been Tested?

I've run and tested these icons and they look more native than the previous one.
This change doesn't affect any other part of the code.

### Screenshots:
Screenshot 2023-04-02 at 6 02 31 AM
Screenshot 2023-04-02 at 6 05 23 AM

### Types of changes

### Checklist:


src/vorta/tray_menu.py
```diff
@@ -9,6 +10,11 @@ class TrayMenu(QSystemTrayIcon):
     def __init__(self, parent: QApplication = None):
         QSystemTrayIcon.__init__(self, parent)
         self.app = parent
+
+        self.forceWhite = False
+        if ("GNOME" in str(os.environ.get('XDG_CURRENT_DESKTOP')).upper()) or (platform.system() == "Darwin"):
+            self.force_white = True
```
real-yfprojects, line 16: I suggest moving this into `set_tray_icon` or a new function to determine the colour dynamically.

real-yfprojects (commented):
Can we use the following method for the tray icon instead of having separate files for black and white like we do with other icons?
https://github.com/borgbase/vorta/blob/8fed5a3324baa5ccfb53732879ad48ff8e877104/src/vorta/views/utils.py#L5

i1sm3ky:
> Can we use the following method for the tray icon instead of having separate files for black and white like we do with other icons?
> 
> https://github.com/borgbase/vorta/blob/8fed5a3324baa5ccfb53732879ad48ff8e877104/src/vorta/views/utils.py#L5

I tried the 'uses_dark_mode' function but it wasn't consistent, it returned 'False' even when in dark mode on MacOS.

real-yfprojects:
> I tried the 'uses_dark_mode' function but it wasn't consistent, it returned 'False' even when in dark mode on MacOS.

You can add another parameter to `get_colored_icon` to force a specific colour.

src/vorta/tray_menu.py
i1sm3ky: This only checks if the OS is Gnome based or MacOS and forces the ui to always use white tray icons. I think I'll be moving this to the set_tray_icon itself because some tests on linux gave attribute not found error due to not being able to find force_white attribute. I made this separate so that I could use it along with the uses_dark_mode method to determine the tray color but as the uses_dark_mode didn't work consistently I skipped it for now and only used force_white to determine tray icon color.

i1sm3ky:
> You can add another parameter to `get_colored_icon` to force a specific colour.

Will the `uses_dark_mode` work fine on other OS? When run on a separate python file for testing it's accurate in determining the os theme dynamically but when implemented in the `tray_menu.py` file it always gave false no matter what the os theme was, even copying the code from `uses_dark_mode` and using it inside the file gave the same result.

src/vorta/tray_menu.py
i1sm3ky: I wrote a code using code from `uses_dark_mode` which dynamically changed the tray icon colour even while running, it detected the theme change in the os, but when implemented inside the `tray_menu.py` it didn't seem to work like intended. It didn't detect theme changes and when forced to check the theme it always gave `False`.

real-yfprojects:
> Will the `uses_dark_mode` work fine on other OS?

I would have thought so but what you describe is rather surprising. I can confirm the method works correctly on KDE in `tray_menu.py` also.

i1sm3ky:
> I would have thought so but what you describe is rather surprising. I can confirm the method works correctly on KDE in `tray_menu.py` also.

So the ui will forcefully use white on gnome and macos so I think if it works on KDE, I could add the `uses_dark_mode` for other os to use white icons on dark mode.

src/vorta/tray_menu.py
```diff
@@ -75,8 +81,9 @@ def build_menu(self):
 
     def set_tray_icon(self, active=False):
         """
-        Use white tray icon, when on Gnome or in dark mode. Otherwise use dark icon.
+        It uses white tray icon when on Gnome or MacOS. Otherwise uses dark icon.
+        Implementation pending for dark mode to use white tray icon.
         """
-        icon_name = f"icons/hdd-o{'-active' if active else ''}.png"
+        icon_name = f"icons/hdd-o{'-white' if self.force_white or uses_dark_mode() else ''}{'-active' if active else ''}.png"
```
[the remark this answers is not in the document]
i1sm3ky: Opps! I could've done that too, thanks for pointing it out. I'll change that and also try use the `get_colored_icon` if possible.

i1sm3ky:
> Can we use the following method for the tray icon instead of having separate files for black and white like we do with other icons?
> 
> https://github.com/borgbase/vorta/blob/8fed5a3324baa5ccfb53732879ad48ff8e877104/src/vorta/views/utils.py#L5

We can't actually use it as the not active black and white icons are slightly different from each other. The white ones have transparent circles in place of white ones in the black icon and that would require significantly more work to get and would make the `get_colored_icon` more complex and not suitable to be used in other places.

real-yfprojects:
If we are able to use svg file for the tray icon. I would use your white icons as a basis, make the black and use `get_colored_icon` to get the white version.

i1sm3ky:
> If we are able to use svg file for the tray icon. I would use your white icons as a basis, make the black and use `get_colored_icon` to get the white version.

But the black ones with holes in them won't look good if the task bar is also dark colored.

i1sm3ky:
Also I'm having `black` lint error, any idea how should I go through fixing it?

diivi:
> Also I'm having `black` lint error, any idea how should I go through fixing it?

Run `black .`

real-yfprojects:
> Run `black .` , this shouldn't be happening though, since we have a pre-commit hook for black.

He probably didn't install pre-commit since it doesn't allow committing unformatted code.

real-yfprojects:
> But the black ones with holes in them won't look good if the task bar is also dark colored.

And the white ones don't look good when the task bar is light coloured. I thought this PR adjusts the colour so that it contrasts the taskbar.

i1sm3ky:
> > Run `black .` , this shouldn't be happening though, since we have a pre-commit hook for black.
> 
> He probably didn't install pre-commit since it doesn't allow committing unformatted code.

I've installed pre-commit.

real-yfprojects:
> I've installed pre-commit.

And it runs without complains?

i1sm3ky:
> And it runs without complains?

There weren't any issues in the VSC. Unlike other lints which tells there's an issue on line - .

i1sm3ky:
> And the white ones don't look good when the task bar is light coloured. I thought this PR adjusts the colour so that it contrasts the taskbar.

Using `PIL` we can do that, Check the color of the system tray and choose the tray icon's color accordingly.
But this would require including a new library hence a new dependency.
What should we do?

real-yfprojects:
Can we use SVG images for the tray icon?

i1sm3ky:
> Can we use SVG images for the tray icon?

Yes we can, PyQt just needs icons to be an object of QIcons.

real-yfprojects:
> Yes we can, PyQt just needs icons to be an object of QIcons.

Then we can use the existing code in `get_colored_icon` to change the colour. We'll only have to replace the corresponding icons with SVG files.

i1sm3ky:
> Then we can use the existing code in `get_colored_icon` to change the colour. We'll only have to replace the corresponding icons with SVG files.

Yes, I considered doing that. But that will require a little change in the original icon, I'll do that asap after completing the proposal and get back you!
Also I'll try to use the `PIL` logic for getting the system tray color to work using PyQt if possible.

i1sm3ky:
> I like the new icon you created. The old one was made with an outline to work in dark and light mode. However an icon without an outline looks better.

Yeah, I'll convert those to svg and then we could use them with the `get_colored_icon` function to get different colored icons depending on the system theme.

> Which PIL logic? We wouldn't want to add PIL as dependency just for this.

I know, that's why I was thinking of finding a similar way or function that uses PyQt to get the system tray color.

i1sm3ky:
Hey @real-yfprojects! Could you please check if this works?

i1sm3ky:
Hey @real-yfprojects! Could you please check if this works? It checks the system tray color and changes the tray icon color accordingly to keep a higher contrast.

i1sm3ky:
[45c924b](https://github.com/borgbase/vorta/pull/1676/commits/45c924bd80353c767b6ecd760db46074b60ae83e) Refactored and made it compatible with PyQt6.

m3nu:
This can be merged now, if it's ready and conflicts resolved.

i1sm3ky:
It can be merged now. Tested on PyQt6 with all the new additions and found no issues. Created a new PR (#1694) for it.

PR #1682, opened 2023-04-04
Add a dev mode that allows for local storing of config files and logs

### LINKED ISSUE #1162: FR: Local config mode for development.
When I start *vorta* from my local python environment, it will always use the global settings and config files. I feel very uneasy when sharing the config files of my real backup system with development versions and branches. This can get especially problematic when a new version uses a different config setup or file format. I also see a problem for manual (and possibly automatic) testing.

A solution would be the option to call *vorta* with an command-line argument like `--local` that starts it in a local config mode. In this mode vorta would not use the user wide location for config files but one nearby *(e.g. in the cwd)*.

ratchek:
### Description
Allows vorta to be called with a command-line flag that will make it use a directory in the project tree to store all the settings, logs, and cache.
Also allows for a custom directory path allowing for multiple "configuration" folders at once.

### Related Issue
Fixes #1162

### Motivation and Context
This can be used to prevent the vorta instance that a developer is working on from accessing the configuration files that they have set up for their personal backups.

### How Has This Been Tested?
Just been tested in the terminal by me.

### Types of changes

### Checklist:


real-yfprojects:
Note how you can write `Fixed #1162` to close the issue when merging the PR.

Fixes #1162.

src/vorta/__main__.py
```diff
@@ -1,14 +1,18 @@
 import os
 import signal
 import sys
+from pathlib import Path
 from peewee import SqliteDatabase
+
+# Need to import config as a whole module instead of individual variables
+# because we will be overriding the modules variables
```
real-yfprojects, line 8: :+1: Nice commentary!

src/vorta/__main__.py
```diff
@@ -45,6 +49,7 @@ def exception_handler(type, value, tb):
 
     want_version = getattr(args, 'version', False)
     want_background = getattr(args, 'daemonize', False)
+    want_development = getattr(args, 'development')
```
real-yfprojects, line 52:
```suggestion
    want_development = getattr(args, 'development', False)
```

I think an exception will be raised if the attribute doesn't exist and no default value is supplied.

src/vorta/__main__.py
```diff
@@ -54,11 +59,44 @@ def exception_handler(type, value, tb):
...
+            os.makedirs(config.CACHE_DIR)
+
+        config.TEMP_DIR = config.CACHE_DIR / 'tmp'
+        if not os.path.exists(config.TEMP_DIR):
+            os.makedirs(config.TEMP_DIR)
+
+        config.PROFILE_BOOTSTRAP_FILE = config.DEV_MODE_DIR / '.vorta-init.json'
+
+    # except Exception as exception:
+    #     print("That's an invalid path for a dir \n", exception)
```
real-yfprojects, line 93: It seems like you either forgot to remove these lines or reimplement the exception handling.

src/vorta/__main__.py
```diff
@@ -54,11 +59,44 @@ def exception_handler(type, value, tb):
...
             sys.exit()
 
+    # want_development is None when flag is not called
+    # it's set to the default directory when flag is called without an argument
+    # and it's set to the flag's argument if the flag has one
+    if want_development:
+        if want_development == DEFAULT_DIR_FLAG:
+            pass
+        else:
+            config.DEV_MODE_DIR = Path(want_development)
```
real-yfprojects, line 69:
```suggestion
        if want_development != DEFAULT_DIR_FLAG:
            config.DEV_MODE_DIR = Path(want_development)
```

src/vorta/utils.py
```diff
@@ -20,6 +20,10 @@
 from vorta.log import logger
 from vorta.network_status.abc import NetworkStatusMonitor
 
+# Used to store whether a user wanted to override the
+# default directory for the --development flag
+DEFAULT_DIR_FLAG = "USE_DEFAULT_DIR"
```
real-yfprojects, line 25:
```suggestion
DEFAULT_DIR_FLAG = object()
```

This is an actually unique value. The user could theoretically supply the path `USE_DEFAULT_DIR`.

src/vorta/__main__.py
```diff
@@ -54,11 +59,44 @@ def exception_handler(type, value, tb):
...
+
+        config.CACHE_DIR = config.DEV_MODE_DIR / 'cache'
+        if not os.path.exists(config.CACHE_DIR):
+            os.makedirs(config.CACHE_DIR)
+
+        config.TEMP_DIR = config.CACHE_DIR / 'tmp'
+        if not os.path.exists(config.TEMP_DIR):
+            os.makedirs(config.TEMP_DIR)
+
+        config.PROFILE_BOOTSTRAP_FILE = config.DEV_MODE_DIR / '.vorta-init.json'
```
real-yfprojects, line 90: Can this be moved into `vorta.config`? And can you then try to reduce redundancy that is duplicated code?

real-yfprojects (commented):
Thank you very much for taking on this issue. That will be a very useful development feature.

ratchek:
> There are some commits in the branch diff that do not belong to this PR. Can you fix that?

Could you elaborate on that please? This is the first time I'm using git in a collaborative manner. I take it you mean that there are commits arising from me merging in work that other people did? How do I go about fixing that?

src/vorta/utils.py
ratchek: Ah, that's clever. I wasn't sure how to go about this and just went for a string that was unlikely to be used but still somewhat informative.

src/vorta/__main__.py
ratchek: That's what I initially planned on doing, but the result was a circular dependency I wasn't quite sure how to fix:

https://github.com/borgbase/vorta/issues/1162#issuecomment-1445214054

real-yfprojects:
> I take it you mean that there are commits arising from me merging in work that other people did? How do I go about fixing that?

Yes, usually these commits aren't shown in the branch diff. Here they are because you somehow rewrote the commits making you the committer. I managed to fix this by running

```bash
git rebase -i master
```
And dropping the commits in question since they are already included in the master branch. I forced pushed the fixed branch, so you will have to update your local copy of it.

src/vorta/__main__.py
real-yfprojects: This is what I came up with. However it isn't perfect since our config dirs aren't real constants any more and can't be imported with `from vorta.config import LOG_DIR` because they won't be updated when imported like that.  Instead one always has to use `from vorta import config`.

```py
SETTINGS_DIR = None
LOG_DIR = None
CACHE_DIR = None
TEMP_DIR = None
PROFILE_BOOTSTRAP_FILE = None


def default_dev_dir():
    return Path(__file__).parent.parent.parent / '.dev_config'


def init_from_platformdirs():
    dirs = platformdirs.PlatformDirs(APP_NAME, APP_AUTHOR)
    init(dirs.user_data_path, dirs.user_log_path, dirs.user_cache_path, dirs.user_cache_path / 'temp', Path.home())


def init_dev_mode(dir):
    init(dir / 'settings', dir / 'logs', dir / 'cache', dir / 'temp', dir)


def init(settings, logs, cache, temp, bootstrap):
    global SETTINGS_DIR
    global LOG_DIR
    global CACHE_DIR
    global TEMP_DIR
    global PROFILE_BOOTSTRAP_FILE
    SETTINGS_DIR = settings
    LOG_DIR = logs
    CACHE_DIR = cache
    TEMP_DIR = temp
    PROFILE_BOOTSTRAP_FILE = bootstrap / '.vorta-init.json'
    ensure_dirs()


def ensure_dirs():
    # ensure directories exist
    for dir in (SETTINGS_DIR, LOG_DIR, CACHE_DIR, TEMP_DIR):
        dir.mkdir(parents=True, exist_ok=True)
```

ratchek: Ok, so I'm gonna do this and then search through the codebase for any imports from config and modify accordingly. Sounds good?

ratchek: Ok, so I'm gonna do this and then search through the codebase for any imports from config and modify accordingly. Sounds good?

real-yfprojects:
This is how I do it:

Keep master branch in sync with upstream
```
git checkout master
git pull upstream
```

Now you can either merge master into the feature branch. This what many people do. Its the easiest method:
```
git checkout feature-branch
git merge master
```

However I prefer keeping clean and linear git histories. I therefore rebase my feature branch onto master:

```
git checkout feature-branch
git rebase master
```

The rebase will also remove all merge commits of master that occur after the diverging.

I would be interested how you managed to get branch you had before. My only guess was that you might have merged `origin/master` or `upstream/master` and than rebased onto `master` which wasn't in sync with the remote branches.

ratchek:
Ok, that's pretty much exactly what I do most of the time. The problem is, I think whenever I get merge conflicts (and I think that's where I git those weird results last time - trying to resolve merge conflicts).

Sidenote:
When I resolve merge conflicts and commit the changes, what should I be putting in the commit message? Just a generic "merge changes from upstream"? That seems redundant and unhelpful, but I don't know what the alternative might be.

ratchek:
Don't merge this yet please. I didn't have a chance to test it out, because the new version with Qt6 is giving me issues (even when doing a fresh install from the main repo) and I need to fix that first.

src/vorta/__main__.py
ratchek: > Ok, so I'm gonna do this and then search through the codebase for any imports from config and modify accordingly. Sounds good?

I modified all the imports except the one in `./package/vorta.spec` on line 7. That one doesn't use any of the variables that are changing. Only APP_NAME and APP_ID_DARWIN, and those are both just assigned as constants. Should I leave it like that, or should I also import the whole module for consistency's sake?

real-yfprojects:
> Ok, that's pretty much exactly what I do most of the time. The problem is, I think whenever I get merge conflicts (and I think that's where I git those weird results last time - trying to resolve merge conflicts).

Merge conflicts are usually resolved by the merge commit itself without rewriting any previous commits. Only few commands rewrite git history.

> When I resolve merge conflicts and commit the changes, what should I be putting in the commit message? Just a generic "merge changes from upstream"?

That's fine. The merge commit will have the commits with helpful descriptions as a parent. Mostly we squash PRs anyway.

src/vorta/application.py
```diff
@@ -23,7 +23,12 @@
 
 logger = logging.getLogger(__name__)
 
-APP_ID = TEMP_DIR / "socket"
+# If config.TEMP_DIR is None, this means that the app got launched withoug going through main
```
real-yfprojects, line 26: Typo

src/vorta/application.py
```diff
@@ -23,7 +23,12 @@
...
+# (for example when running tests or when running on schedule.)
+# In any case, if the config dirs aren't set, assume you're runnning system wide initialize appropriately
```
real-yfprojects, line 28: Typo

src/vorta/application.py
```diff
@@ -23,7 +23,12 @@
...
+if not config.TEMP_DIR:
+    config.init_from_platformdirs()
```
real-yfprojects, line 30: I might be better to call `init_from_platformdirs` in `config` (every time) and override it in `__main__` when needed.

src/vorta/application.py
```diff
@@ -259,7 +264,7 @@ def break_lock(self, profile):
         job = BorgBreakJob(params['cmd'], params)
         self.jobs_manager.add_job(job)
 
-    def bootstrap_profile(self, bootstrap_file=PROFILE_BOOTSTRAP_FILE):
+    def bootstrap_profile(self, bootstrap_file=config.PROFILE_BOOTSTRAP_FILE):
```
real-yfprojects, line 267: When is `config.PROFILE_BOOTSTRAP_FILE` evaluated? 

I think it is evaluated when the module is imported. Therefore changes to `config` aren't respected.

src/vorta/application.py
ratchek: Sorry, what does this mean? I can't find the typo.

src/vorta/application.py
ratchek: I'm not sure what to do about this. Could you point me in the right direction?

src/vorta/application.py
real-yfprojects: It sais `withoug` instead of `without`.

src/vorta/application.py
real-yfprojects:
```suggestion
    def bootstrap_profile(self, bootstrap_file=None):
        bootstrap_file = bootstrap_file or config.PROFILE_BOOTSTRAP_FILE
```

src/vorta/application.py
ratchek: You mean calling `init_from_platformdirs` in `application`? Or calling that method in `config`  so that it runs every time the module is loaded? Wouldn't the latter risk overwriting the variables from time to time as a side effect?

real-yfprojects: > Or calling that method in config so that it runs every time the module is loaded?

That's what I was thinking of since it ensures the config values contain valid values.

> Wouldn't the latter risk overwriting the variables from time to time as a side effect?

Python modules are only loaded once (and cached) in a run afaik.

ratchek: Ohhh. I see. Thanks! I'll do that.

ratchek:
So, I just realized that now there's this issue, where everything works fine, unless you call `vorta` with the `--development` flag and pass it a custom directory using relative paths. So for example, if you do `vorta --development ./dev-mode`. In that particular scenario, the link to logs in the misc tab doesn't work. The logs exist and they're in the correct folder, but the link doesn't work.
I would suspect the call to `config` on line 37 in `views/misc_tab.py` is happening before the actual modification of the config values, however if you supply the full path when you're running the program (for example `vorta --development /home/username/path/to/project/folder/dev-mode`), then everything works fine. Is it possible that the string interpolation is not unpacking the full path correctly?

ratchek:
> Is it possible that the string interpolation is not unpacking the full path correctly?

Yup. That was it. Fixed it. Should be good now.

real-yfprojects (commented):
Nice! Can you add type hints and docstrings?

src/vorta/utils.py
```diff
@@ -353,7 +357,20 @@ def parse_args():
...
+    # will be set to that argument
+    parser.add_argument(
+        '--development',
+        '-D',
+        nargs='?',
+        const=DEFAULT_DIR_FLAG,
+        help='Start vorta in a local development environment. '
+        'All log, config, cache, and temp files will be stored within the project tree. '
+        'You can follow this flag with an optional path and it will store the files in the provided location.',
+    )
```
real-yfprojects, line 373: Can you specify a `metavar` so that it doesn't say 

```
usage: vorta [-h] [--version] [--daemonize] [--create PROFILE]
             [--development [DEVELOPMENT]]
```

when running `vorta --help`?

real-yfprojects (approved):

real-yfprojects:
Nice work, thanks!

ratchek:
> Nice work, thanks!

Thank you so so much for guiding me through this and being so patient.

m3nu:
That's fine. I see you mostly changed `XXX` to `config.XXX`. Reminds me a bit of Django now.

Updating the docs would be important too. So people actually use this feature as intended.

m3nu (approved):

real-yfprojects:
> Updating the docs would be important too. So people actually use this feature as intended.

Follow up issue: borgbase/vorta.borgbase.com#42

PR #1694, opened 2023-04-17
Change tray icon color based on system tray color

i1sm3ky:
### Description

This PR changes tray icon color based on the system tray color which increases the contrast between the tray icon and the system tray.

### Related Issue

Issue: #1220 requesting for white icons for menu bar on MacOS

### Motivation and Context

This change provides white icons for dark themed Gnome based linux and MacOS which look better as solid white icons instead of the dark black ones.

### How Has This Been Tested?

I've run and tested these icons and they look more native than the previous one.
This change doesn't affect any other part of the code.

### Screenshots:
Screenshot 2023-04-02 at 6 02 31 AM
Screenshot 2023-04-02 at 6 05 23 AM

### Types of changes

### Checklist:

### DIFF
--- a/tests/unit/test_import_export.py
+++ b/tests/unit/test_import_export.py
@@ -4,6 +4,7 @@ from pathlib import Path
 import pytest
 from PyQt6 import QtCore
 from PyQt6.QtWidgets import QDialogButtonBox, QFileDialog, QMessageBox
+from vorta.profile_export import VersionException
 from vorta.store.models import BackupProfileModel, SourceFileModel
 from vorta.views.import_window import ImportWindow
 
@@ -30,6 +31,41 @@ def test_import_success(qapp, qtbot, rootdir, monkeypatch):
     restored_repo = restored_profile.repo
     assert restored_repo is not None
     assert len(SourceFileModel.select().where(SourceFileModel.profile == restored_profile)) == 3
+
+
+@pytest.mark.parametrize(
+    "exception, error_message",
+    [
+        (AttributeError, "Schema upgrade failure"),
+        (VersionException, "Newer profile_export export files cannot be used on older versions"),
+        (PermissionError, "Cannot read profile_export export file due to permission error"),
+        (FileNotFoundError, "Profile export file not found"),
+    ],
+)
+def test_import_exceptions(qapp, qtbot, rootdir, monkeypatch, mocker, exception, error_message):
+    monkeypatch.setattr(QFileDialog, "getOpenFileName", lambda *args: [VALID_IMPORT_FILE])
+    monkeypatch.setattr(QMessageBox, 'information', lambda *args: None)
+
+    main = qapp.main_window
+    main.profile_import_action()
+    import_dialog: ImportWindow = main.window
+    import_dialog.overwriteExistingSettings.setChecked(True)
+
+    def raise_exception(*args, **kwargs):
+        raise exception
+
+    # force an exception and mock the error QMessageBox
+    monkeypatch.setattr(import_dialog.profile_export, 'to_db', raise_exception)
+    mock_messagebox = mocker.patch.object(QMessageBox, "critical")
+
+    qtbot.mouseClick(
+        import_dialog.buttonBox.button(QDialogButtonBox.StandardButton.Ok), QtCore.Qt.MouseButton.LeftButton
+    )
+
+    # assert the correct error appears, and the profile does not get added
+    mock_messagebox.assert_called_once()
+    assert error_message in mock_messagebox.call_args[0][2]
+    assert BackupProfileModel.get_or_none(name="Test Profile Restoration") is None
 
 
 def test_import_bootstrap_success(qapp, mocker):
@@ -92,7 +128,7 @@ def test_export_success(qapp, qtbot, tmpdir, monkeypatch):
     assert os.path.isfile(FILE_PATH)
 
 
-def test_export_fail_unwritable(qapp, qtbot, tmpdir, monkeypatch):
+def test_export_fail_unwritable(qapp, qtbot, monkeypatch):
     FILE_PATH = os.path.join(os.path.abspath(os.sep), "testresult.vortabackup")
 
     def getSaveFileName(*args, **kwargs):


i1sm3ky:
@real-yfprojects Could you please cross check and review these changes once.

m3nu:
No need to notify individual people. Someone will see it in time and review it.

i1sm3ky:
> No need to notify individual people. Someone will see it in time and review it.

Ok, Thanks!

src/vorta/views/utils.py
```diff
@@ -2,14 +2,17 @@
...
-def get_colored_icon(icon_name):
+def get_colored_icon(icon_name, force_color=''):
     """
     Return SVG icon in the correct color.
     """
     with open(get_asset(f"icons/{icon_name}.svg"), 'rb') as svg_file:
         svg_str = svg_file.read()
-    if uses_dark_mode():
-        svg_str = svg_str.replace(b'#000000', b'#ffffff')
+    if not force_color:
```
m3nu, line 11: It's generally better to avoid `if not` when one has a choice since `if-else` is easier to understand. So I suggest to swap this to be `if force_color:... else: ...`

src/vorta/tray_menu.py
```diff
@@ -73,10 +73,19 @@ def build_menu(self):
         exit_action = menu.addAction(self.tr('Quit'))
         exit_action.triggered.connect(self.app.quit)
 
+    def is_taskbar_dark(self):
+        app = QApplication.instance()
```
m3nu, line 77: `self.app` should already have the QApplication instance.

src/vorta/tray_menu.py
```diff
@@ -73,10 +73,19 @@ def build_menu(self):
...
+        system_tray_color = int(QScreen.grabWindow(app.primaryScreen()).toImage().pixel(0, 0))
```
m3nu, line 78: This only works if the task bar is on top. So it's true for Mac and Gnome, but not everywhere.

It also triggers a privacy warning on Mac.

Screenshot 2023-04-18 at 08 24 07

m3nu: Maybe it's possible to create the task bar icon and then record it's color only?

src/vorta/tray_menu.py
i1sm3ky: Changed it! It works now, earlier it gave error before PyQt6.

src/vorta/views/utils.py
i1sm3ky: Done!

src/vorta/tray_menu.py
i1sm3ky: > This only works if the task bar is on top. So it's true for Mac and Gnome, but not everywhere.

I can check which os it's using and check for tray color on either top or bottom based on it.
 
> It also triggers a privacy warning on Mac.

It didn't on mine for some reason.

> Maybe it's possible to create the task bar icon and then record it's color only?

Any ideas on how can I create the initial icon to have a contrasting color to the tray color?

real-yfprojects: > I can check which os it's using and check for tray color on either top or bottom based on it.

KDE Plasma allows the task bar to be on any edge of the screen and it can be shorter than the edge it sits on.

i1sm3ky: > KDE Plasma allows the task bar to be on any edge of the screen and it can be shorter than the edge it sits on.

Oh, Then I think I might have to find another way to locate and scan the tray color.

i1sm3ky: Found a way using `QSystemTrayIcon.geometry()`.

real-yfprojects: I think there is no real way as of now. Getting the following to work looks like the best solution: https://github.com/borgbase/vorta/blob/e0fe766051e521396d7d58f973d1f8002ff749ad/src/vorta/utils.py#L374-L379

i1sm3ky: > I think there is no real way as of now. Getting the following to work looks like the best solution: https://github.com/borgbase/vorta/blob/e0fe766051e521396d7d58f973d1f8002ff749ad/src/vorta/utils.py#L374-L379 

It returns if the system theme is dark or not, but on MacOS the system tray is themed according to the desktop background and is independent of system theme.

real-yfprojects: There is the concept/file type of *template* images on MacOS but I wasn't able to find much information how that could be used from python.

i1sm3ky: > There is the concept/file type of _template_ images on MacOS but I wasn't able to find much information how that could be used from python.

Thanks for the help! I looked into those a little and found a way of using it for our use.

i1sm3ky: > There is the concept/file type of _template_ images on MacOS but I wasn't able to find much information how that could be used from python.

Thanks for the help! I looked into those a little and found a way to use it.
But there's a slight issue, they only work as black and white so the coloured dots which represented the backing up of data in progress don't render in colours and instead just render as white or black. Also I would like to know if this works on linux?

real-yfprojects: I suppose template images do not work in Linux. Let's first add the code for MacOs and think about linux later.

real-yfprojects: > But there's a slight issue, they only work as black and white so the coloured dots which represented the backing up of data in progress don't render in colours and instead just render as white or black.

Can we somehow determine how a template tray icon was rendered and extract information about the tray through that?

real-yfprojects: Alternatively we could add a setting and let the user pick the tray colour manually.

i1sm3ky: > I suppose template images do not work in Linux. Let's first add the code for MacOs and think about linux later.

Are linux system tray's theme also based on the desktop wallpaper or they follow the system's theme? 

> Can we somehow determine how a template tray icon was rendered and extract information about the tray through that?

I tried that but the colour change is based on the system os as the template is a transparent image only having information about the dimensions and the shape and when trying to extract its colour by converting the current qicon to a pixmap and checking the colour of a pixel on there returns same for both light and dark icons.

> Alternatively we could add a setting and let the user pick the tray colour manually.

Yes we can do that as well, but that too will require some logic to check if the selected colour is in range of the dots indicating the working of vorta while backing and converting them to some different colour.

I'll check further into these and let you know if I find any alternatives.

real-yfprojects: > Are linux system tray's theme also based on the desktop wallpaper or they follow the system's theme?

Usually they follow the system theme, but on KDE you can use custom themes that can customize every single bit of the appearance.

> Yes we can do that as well, but that too will require some logic to check if the selected colour is in range of the dots indicating the working of vorta while backing and converting them to some different colour.

This might actually be the best solution.

i1sm3ky: > This might actually be the best solution.

I think so. I'll work on that after I get home in a few days.

real-yfprojects:
Are you still working on this @i1sm3ky?

i1sm3ky:
> Are you still working on this @i1sm3ky?

I was a little confused as per which solution to implement, also I was on a vacation and I'll be going back to my university in the first week of august then I could start implementing it.

Could you please confirm which solution should I implement?

i1sm3ky:
> Are you still working on this @i1sm3ky?

Hey there! Sorry for the delay. Actually I was a little confused as per which solution to implement, also I was on a vacation and I'll be heading to my university in the first week of august then I can start implementing the solution.

Could you please confirm which solution will be better to implement?

real-yfprojects:
I think this is were we left of:

> > Alternatively we could add a setting and let the user pick the tray colour manually.
> 
> Yes we can do that as well, but that too will require some logic to check if the selected colour is in range of the dots indicating the working of vorta while backing and converting them to some different colour.

That would be neat. However if you find that this is too complicated you can simply add a setting for toggling the icon between dark and light mode.

i1sm3ky:
> That would be neat. However if you find that this is too complicated you can simply add a setting for toggling the icon between dark and light mode.

I think I can do that, I'll start working on it as soon as possible.

i1sm3ky:
@real-yfprojects I've a few questions.

1. Where should we have the option to change the icon color?
2. Could you please point me towards where the settings are saved?

i1sm3ky:
@real-yfprojects I've a few questions.

Where should we have the option to change the icon color?
Could you please point me towards where the settings are saved?

i1sm3ky:
@real-yfprojects I've a few questions.

Where should we have the option to change the icon color?
Could you please point me towards where the settings are saved?

i1sm3ky:
@real-yfprojects I've a few questions.

Where should we have the option to change the icon color?
Could you please point me towards where the settings are saved?

real-yfprojects:
Settings are defined in https://github.com/borgbase/vorta/blob/f27c6536adc49b512725543def4e971e6dabe938/src/vorta/store/settings.py#L7 and stored in https://github.com/borgbase/vorta/blob/f27c6536adc49b512725543def4e971e6dabe938/src/vorta/store/models.py#L204, while https://github.com/borgbase/vorta/blob/f27c6536adc49b512725543def4e971e6dabe938/src/vorta/store/connection.py#L90-L102 ensures the settings model always contains the correct settings.
The widgets are generated in https://github.com/borgbase/vorta/blob/f27c6536adc49b512725543def4e971e6dabe938/src/vorta/views/misc_tab.py#L56.

real-yfprojects:
Since you still want to work on this I will reopen as a draft.

real-yfprojects:
> Should the option to change the icon color be added in the misc tab?

That would be great!

PR #1697, opened 2023-04-17
Make {hostname} name template consistent with Borg, add {fqdn} template option

### LINKED ISSUE #1696: hostname placeholder expansion
Vorta: misc 0.8.x versions and also 0.9.0 (on macOS M1)

In Vorta `{hostname}-...` is configured for creating the archive names and `{hostname}-` as prefix for pruning.

As a borg placeholder, `{hostname}` is documented to expand to the short hostname (not the fqdn).

But I see archive names in the repo with my mac hostname with misc. domains appended (sometimes `.local`, sometimes `.fritz.box`), so it seems to expand to the fqdn.

Especially since this does not always expand to the same fqdn, it creates a bit of a mess, especially considering `prune`.

m3nu:
### Description
Before, using `{hostname}` as archive or prune name template may include the network domain, like `.local` or `.fritzbox`. This can interfere with pruning and generate different archive names.

With this PR, only the short hostname will be used and `fqdn` is added as option.

### Related Issue
Fixes #1696

### DIFF
--- a/src/vorta/assets/UI/archivetab.ui
+++ b/src/vorta/assets/UI/archivetab.ui
@@ -618,7 +618,7 @@
             <item row="0" column="1">
              <widget class="QLineEdit" name="prunePrefixTemplate">
               <property name="toolTip">
-               <string>Available variables: hostname, profile_id, profile_slug, now, utc_now, user</string>
+               <string>Available variables: hostname, fqdn, profile_id, profile_slug, now, utc_now, user</string>
               </property>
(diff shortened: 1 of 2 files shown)


real-yfprojects:
> Maybe add a compatibility warning to changelog or docs?

Yeah, I'd consider this a breaking change in terms of semver.

m3nu:
Luckily we just made a breaking change to 0.9. Good timing.

PR #1710, opened 2023-05-14
Added '.log' suffix to log files

### LINKED ISSUE #1707: bad logfile names
Vorta creates log file names like:
- vorta.log
- vorta.log.2023-04-05
- vorta.log.2023-05-06

The "rotated" file names are bad: if you click on "logs" in the vorta UI, it will show you a file selector box and after selecting a log file, it will ask you to choose a program to view that log.

It will ask again and again because it considers `.log` (ok) and `.2023-04-05` (bad) as filename extensions ("file types").

So, the log file names should have the same extension, like e.g.: `vorta.2023-04-05.log`

## Solution

1. Subclass `logging.TimedRotatingFileHandler` inside [log.py](https://github.com/borgbase/vorta/blob/master/src/vorta/log.py)
2. Append `.log` to the suffix in the `__init__` of the created subclass
3. Use the subclass in [init_logger](https://github.com/borgbase/vorta/blob/4d65912d65e18e71462a03e903f6eb71c0edf9bd/src/vorta/log.py#L16)
4. Write unittests for the new subclass in a new test file.

bigtedde:
### Description
Currently, log files don't always end in ".log" due to the way the names get rotated by `TimedRotatingFileHandler`. In this PR, I ensure the files have the ".log" suffix.

### Related Issue
Fixes #1707 - Bad Logfile Names

### Motivation and Context
Log files will now be easier to open

### How Has This Been Tested?
Locally, and passed CI tests

### Types of changes

### Checklist:


bigtedde:
@ThomasWaldmann I used your recommended solution to #1707 as a guide for this PR. Any feedback is greatly appreciated :) 

Thanks!

src/vorta/log.py
```diff
@@ -14,6 +14,16 @@
...
 
 
+class LogTimedRotatingFileHandler(TimedRotatingFileHandler):
+    """Create subclass of TimedRotatingFileHandler that always appends '.log' to log files."""
+
+    def __init__(self, filename, when='d', interval=1, backupCount=5, encoding=None, delay=False, utc=False):
+        super().__init__(
+            filename, when=when, interval=interval, backupCount=backupCount, encoding=encoding, delay=delay, utc=utc
+        )
+        self.suffix += '.log'
```
ThomasWaldmann, line 24: so you append .log to whatever is already there.

what exactly is already there at init time? does that vary?

tests/test_log.py
```diff
@@ -0,0 +1,8 @@
+from vorta.log import LogTimedRotatingFileHandler
+
+
+def test_log_file_suffix():
+    """Ensure log files always have '.log' suffix."""
+
+    handler = LogTimedRotatingFileHandler('mylog.log', when='d', interval=1, backupCount=5)
+    assert handler.suffix.endswith('.log')
```
ThomasWaldmann, line 8: guess that is not enough to test the dynamic behaviour (which is hard).
maybe rather have no automated test than a pointless one?

src/vorta/log.py
real-yfprojects: https://github.com/python/cpython/blob/main/Lib/logging/handlers.py#L431-L432

This shows how the suffix is ueed. The suffix is constructed during init:

https://github.com/python/cpython/blob/b378d991f8cd41c33416e590cb83472cce1d6b98/Lib/logging/handlers.py#L210

While the solution I initially came up with and which was implemented in this PR works, I noticed that there is a documented way we should use instead:

https://docs.python.org/3.12/library/logging.handlers.html#logging.handlers.BaseRotatingHandler.namer

bigtedde: @ThomasWaldmann At init time the suffix is set to a year-month-day timestamp [here](https://github.com/python/cpython/blob/cb88ae635e96d7020ba6187bcfd45ace4dcd8395/Lib/logging/handlers.py#LL246C20-L246C20). This suffix then gets appended [here](https://github.com/python/cpython/blob/main/Lib/logging/handlers.py#L431-L432) to whichever filename (in this case, "vorta.log") is passed to the `TimedRotatingFileHandler`. I see now that editing the suffix directly may not be the best approach.

@real-yfprojects After reviewing the code and documentation you linked, maybe a better approach would be to create a function for `namer`. Currently the log files are getting named "vorta.log.[timestamp]". I was thinking of making a function that names them "vorta.[timestamp].log" and setting namer to this function. What do you think of this approach?

tests/test_log.py
bigtedde: @ThomasWaldmann  Agreed - I missed the mark here. 

Currently we are using a default filename format for the log (ex. vorta.log.[timestamp]). I am considering creating a function for `namer` that moves the '.log' to the end of filename (ex. vorta.[timestamp].log). 

To test this, I was considering testing the parts of the code that call `rotation_filename`, as this function uses `namer` to set the custom filenames. I'm curious what you think of this approach? What behavior would you recommend testing?

src/vorta/log.py
ThomasWaldmann: If you set `when="S"`, you could do a realistic test in a few seconds (instead of days).

bigtedde: Definitely! I've been doing that to test my local changes.

bigtedde:
Created a function for namer that appended '.log' to the filename.

Testing: Since '[namer](https://github.com/python/cpython/blob/b378d991f8cd41c33416e590cb83472cce1d6b98/Lib/logging/handlers.py#L51)' is used in '[rotation_filename()](https://github.com/python/cpython/blob/b378d991f8cd41c33416e590cb83472cce1d6b98/Lib/logging/handlers.py#L79)', testing was focused on the areas of code that called this; specifically, in [doRollover()](https://github.com/python/cpython/blob/b378d991f8cd41c33416e590cb83472cce1d6b98/Lib/logging/handlers.py#L431). 

The test involved creating a temp directory that called `doRollover()` a few more times than `backupCount` to ensure changing the filename did not mess with file creation and subsequent rotation/deletion. The behavior stayed the same as it was before, and all log files had the ".log" suffix.

tests/test_log.py
```diff
@@ -1,8 +1,39 @@
...
+                if i == 0:
+                    assert len(os.listdir(log_dir)) == 2  # Should keep initial backup + current log
+                else:
+                    if i < handler.backupCount:
+                        assert len(os.listdir(log_dir)) == i + 1  # Should keep i backups + current log
+                    else:
+                        assert len(os.listdir(log_dir)) == handler.backupCount + 1  # Keeps backupCount backups + log
+
+                for file in os.listdir(log_dir):
+                    assert file.endswith('.log')  # Checks that 'namer' is working as intended
```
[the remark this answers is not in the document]
bigtedde: The current log is 'vorta.log', and the backups are 'vorta.log.[timestamp].log'

src/vorta/log.py
```diff
@@ -24,6 +28,7 @@ def init_logger(background=False):
 
     # create handlers
     fh = TimedRotatingFileHandler(LOG_DIR / 'vorta.log', when='d', interval=1, backupCount=5)
```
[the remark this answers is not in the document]
bigtedde: Exactly, and I found it's important to keep it 'vorta.log' instead of just 'vorta' because it is set as name of the current log file, and does not go through the `namer` function to append '.log' to the end of it in the way the backup log files do.

tests/test_log.py
```diff
@@ -0,0 +1,39 @@
...
+
+                time.sleep(1)  # Need 1s between logs or they are overwritten
```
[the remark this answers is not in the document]
bigtedde: Point well taken :) I will remove the test and resubmit.

src/vorta/log.py
```diff
@@ -14,6 +14,11 @@
 logger = logging.getLogger()
 
 
+def log_namer(default_name):
+    # Backups will be named 'vorta.log.[timestamp].log'
```
ThomasWaldmann, line 18: maybe rather say "rotated log files" or "old log files".

tests/test_log.py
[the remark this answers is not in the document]
real-yfprojects: > but removing this in namer messed with other functionality.

What functionality do you mean?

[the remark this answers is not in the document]
bigtedde: The base filename is `vorta.log`, of which a timestamp gets appended on to. The `namer` function then appends `.log` to the end of this resulting in `vorta.log.[timestamp].log`

Unfortunately, if I implement functionality in the `namer` function to remove the superfluous `.log` , the `getFilesToDelete()` method in `TimedRotatingFileHandler` no longer properly identifies the files to delete, resulting in backups not rotating as intended. 

Additionally, if I set the base filename to simply 'vorta' instead of 'vorta.log', then the current log file will not get the '.log' suffix added since only rotated log files go through the `namer` function.

tests/test_log.py
ThomasWaldmann: Guess the code is now simple enough to not need a test.

ThomasWaldmann:
Guess this needs the conflict resolved and then should get merged.

bigtedde:
@m3nu Simplified change and works as intended. Ready for review.

m3nu (approved):

m3nu:
Great solution!

PR #1723, opened 2023-06-01
feat: refresh multiple archives

diivi:
### Description
First task for one of my GSoC Projects, "Enhancing the Archive Table actions". I listed these tasks out in the proposal, in total they should take me around 50 hours:

- [ ] #1734 
- [x] #1732 
- [x] #1727 
- [ ] #1723
- [ ] #1664

This will allow Vorta users to select multiple archives and refresh them at once, hence improving UX.

### Motivation and Context
Saves the user's time.

### How Has This Been Tested?
I tested this manually:

image

### Types of changes

### Checklist:


src/vorta/views/archive_tab.py
```diff
@@ -365,14 +367,18 @@ def on_selection_change(self, selected=None, deselected=None):
...
+            self.bRename.setEnabled(False)
 
             for index in range(layout.count()):
                 widget = layout.itemAt(index).widget()
                 tooltip = widget.toolTip()
 
                 tooltip = self.tooltip_dict.setdefault(widget, tooltip)
-                widget.setToolTip(tooltip + " " + reason)
+
+                if 'Refresh' not in tooltip:
```
diivi, line 380: Is this a dirty hack? I thought this was the only way to do this with the layout. Otherwise I could do something like https://github.com/borgbase/vorta/pull/1723/files#diff-1340facde0499b13accf612e9eab00d2c0909090fb7320e61ce2f0cf6282ca02R370-R372

m3nu: This link goes to the current page, but you're right. This can be solved better probably.

m3nu: `if 'Refresh' not in tooltip:`

This one doesn't look very stable if we ever rename the tooltip text. Maybe there is a better way to use some metadata or name of the button instead.

diivi: Thanks, Done.

Hofer-Julian (commented):
The PR works as expected.
I am a bit unsure about the implementation, a bit more discussion is needed here.

src/vorta/views/archive_tab.py
```diff
@@ -365,14 +367,18 @@ def on_selection_change(self, selected=None, deselected=None):
...
 
             for index in range(layout.count()):
                 widget = layout.itemAt(index).widget()
                 tooltip = widget.toolTip()
 
                 tooltip = self.tooltip_dict.setdefault(widget, tooltip)
-                widget.setToolTip(tooltip + " " + reason)
+
+                if widget != self.bRefreshArchive:
+                    widget.setToolTip(tooltip + " " + reason)
```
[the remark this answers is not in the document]
diivi: Because every button is disabled when multiple archives are selected except the refresh button. The "reason" variable here is - Only one archive should be selected (similar).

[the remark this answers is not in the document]
real-yfprojects: What about:

```
 # too few or too many selected.
for widget in [self.fArchiveActions, self.bMountArchive, ...]:
    tooltip = widget.toolTip()
    tooltip = self.tooltip_dict.setdefault(widget, tooltip)
    widget.setToolTip(tooltip + " " + reason)
    widget.setEnabled(False)

# special treatment for dynamic mount/unmount button.
self.bmountarchive_refresh()
tooltip = self.bMountArchive.toolTip()
self.bMountArchive.setToolTip(tooltip + " " + reason) 

```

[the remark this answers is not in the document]
Hofer-Julian: In understand, but that also applies for the diff button. Can't you treat them similarly?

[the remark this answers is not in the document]
diivi: > What about:
> 
> ```python
> # too few or too many selected.
> for widget in [self.fArchiveActions, self.bMountArchive, ...]:
>     tooltip = widget.toolTip()
>     tooltip = self.tooltip_dict.setdefault(widget, tooltip)
>     widget.setToolTip(tooltip + " " + reason)
>     widget.setEnabled(False)
> 
> # special treatment for dynamic mount/unmount button.
> self.bmountarchive_refresh()
> tooltip = self.bMountArchive.toolTip()
> self.bMountArchive.setToolTip(tooltip + " " + reason) 
> ```

self.fArchiveActions will include the refresh button though, so this snippet did not work.
But with the same idea, this works:

```python
        if self.repoactions_enabled and len(indexes) == 1:
            # Enable archive actions
            for widget in [self.bMountArchive, self.bExtract, self.bRename]:
                widget.setEnabled(True)

            for index in range(layout.count()):
                widget = layout.itemAt(index).widget()
                widget.setToolTip(self.tooltip_dict.get(widget, ""))

            # refresh bMountArchive for the selected archive
            self.bmountarchive_refresh()
        else:
            reason = reason or self.tr("(Select exactly one archive)")

            # too few or too many selected.
            for widget in [self.bMountArchive, self.bExtract, self.bRename]:
                tooltip = widget.toolTip()
                tooltip = self.tooltip_dict.setdefault(widget, tooltip)
                widget.setToolTip(tooltip + " " + reason)
                widget.setEnabled(False)
```

src/vorta/views/archive_tab.py
```diff
@@ -353,7 +355,8 @@ def on_selection_change(self, selected=None, deselected=None):
 
         if self.repoactions_enabled and len(indexes) == 1:
             # Enable archive actions
-            self.fArchiveActions.setEnabled(True)
+            for widget in [self.bMountArchive, self.bExtract, self.bRename]:
```
Hofer-Julian, line 358: I think this is a much nicer approach :)

One small thing: How about you extract `[self.bMountArchive, self.bExtract, self.bRename]` here and on line 371 into a variable with a descriptive name?

diivi: Good idea!

Hofer-Julian (approved):
Great job!

src/vorta/views/archive_tab.py
```diff
@@ -330,12 +332,14 @@ def on_selection_change(self, selected=None, deselected=None):
         if not self.repoactions_enabled:
             reason = self.tr("(borg already running)")
 
-        # toggle delete button
+        # Disable the delete and refresh buttons if no archive is selected
         if self.repoactions_enabled and len(indexes) > 0:
             self.bDelete.setEnabled(True)
+            self.bRefreshArchive.setEnabled(True)
             self.bDelete.setToolTip(self.tooltip_dict.get(self.bDelete, ""))
```
real-yfprojects, line 339: Wouldn't it be more readable if one grouped these three statements by button?

src/vorta/views/archive_tab.py
```diff
@@ -330,12 +332,14 @@ def on_selection_change(self, selected=None, deselected=None):
...
         else:
             self.bDelete.setEnabled(False)
+            self.bRefreshArchive.setEnabled(False)
```
real-yfprojects, line 342: According to the HIG disabled widgets should state a reason for being disabled when they are by enclosing the reason with brackets and appending it to the tooltip. The refresh button is missing that.

src/vorta/views/archive_tab.py
```diff
@@ -488,15 +490,28 @@ def list_result(self, result):
...
+
+        name_list = []
+        for index in selected_archives:
+            name_list.append(self.archiveTable.item(index.row(), 4).text())
+
+        archive_list = (
+            profile.repo.archives.select().where(ArchiveModel.name << name_list).order_by(ArchiveModel.time.desc())
+        )
+
+        archive_names = [archive.name for archive in archive_list]
```
real-yfprojects, line 504: Why is all of that needed?

real-yfprojects (commented):
Until now the buttons where visually grouped into single and multi archive options. This PR dismisses this grouping logic. However I am starting to think that grouping the buttons like that might not be so useful afterall. What do you think?

src/vorta/views/archive_tab.py
diivi: I pasted this stuff from other pieces of the code to get the names of the selected archives, do you have a better alternative in mind?

real-yfprojects: [`selected_archive_name`](https://github.com/borgbase/vorta/blob/2cb9afd4d551dfa5f251d6edafb29de020a4234e/src/vorta/views/archive_tab.py#LL507C24-L507C24) works with less code.

diivi: This does the same thing with multiple archives instead of one though, right?

real-yfprojects: Your code starts off the same as `selected_archive_name`. Then you query the database to sort the archives. This differs from the `selected_archive_name`. Why do you do that?

diivi:
```suggestion
        profile = self.profile()

        archive_names = []
        for index in selected_archives:
            archive_names.append(self.archiveTable.item(index.row(), 4).text())
```
I should be doing this right?

real-yfprojects: Yes, that should work.

src/vorta/views/archive_tab.py
```diff
@@ -488,15 +492,28 @@ def list_result(self, result):
...
+
+        for archive_name in archive_names:
+            if archive_name is not None:
+                params = BorgInfoArchiveJob.prepare(self.profile(), archive_name)
+                if params['ok']:
```
real-yfprojects, line 511: Can you log the problem when `not params['ok']` or maybe even display it in the GUI.

real-yfprojects: Similar to this:
https://github.com/borgbase/vorta/blob/2cb9afd4d551dfa5f251d6edafb29de020a4234e/src/vorta/views/archive_tab.py#L810-L813C19

src/vorta/views/archive_tab.py
```diff
@@ -488,15 +492,28 @@ def list_result(self, result):
...
+                    job = BorgInfoArchiveJob(params['cmd'], params, self.profile().repo.id)
+                    job.updated.connect(self._set_status)
+                    job.result.connect(self.info_result)
+                    self._toggle_all_buttons(False)
+                    self.app.jobs_manager.add_job(job)
 
     def info_result(self, result):
         self._toggle_all_buttons(True)
```
real-yfprojects, line 519: This will enable the buttons again although the other info jobs are still running.

diivi: So should I connect just the last job's result to this function?

real-yfprojects: Well, theoretically you don't know which job finishes first. Did you try to run the current code? I would think that only one borg jobs succeeds because of the repo lock.

diivi: Yeah I tried running it (I attached a GIF above).

diivi: Nevermind, I didn't attach the GIF. Here,
Peek 2023-06-15 02-22

They do blink.

m3nu:
The enabled/disabled state of buttons seems to go on and off. Or is that just the GIF?

And didn't we want to rename "Refresh" to something else? I recall a discussion here or elsewhere.

diivi:
> And didn't we want to rename "Refresh" to something else? I recall a discussion here or elsewhere.

Ah yes, done in the next commit.

> The enabled/disabled state of buttons seems to go on and off.

Yes, I am confused how to disable them until all jobs are completed though (so they don't get enabled after every single job), looking around.

m3nu:
> Yes, I am confused how to disable them until all jobs are completed though (so they don't get enabled after every single job), looking around.

Depends on how they are toggled right now. If it's from the BorgJob directly, there isn't much to be done without larger changes.

diivi:
> If it's from the BorgJob directly, there isn't much to be done without larger changes.

Yeah, it's connected to the BorgJob's result :/

diivi:
> Yeah, it's connected to the BorgJob's result :/

What can I do now?

m3nu:
OK for me. An isolated change, which doesn't add much new code.

Comments should be resolved, if any action is needed on them.

diivi:
> Comments should be resolved, if any action is needed on them.

I am only stuck with your question here:
https://github.com/borgbase/vorta/pull/1723#issuecomment-1591983364 

But you said 

> If it's from the BorgJob directly, there isn't much to be done without larger changes.

other issues are solved, so probably ready to merge.

real-yfprojects:
> I am only stuck with your question here:
> [#1723 (comment)](https://github.com/borgbase/vorta/pull/1723#issuecomment-1591983364)

Then we will need to open a follow-up issue for that problem. Or decide to fix the flashing problem first.

jetchirag:
Not sure if I'm missing something but they seem to toggled in `archive_tab.py`, instead of `BorgJob`

https://github.com/borgbase/vorta/blob/92608f9eaaa6ac9d4bf28f9fa59edaae6001fe07/src/vorta/views/archive_tab.py#L532C21-L536C48

So if you move `self._toggle_all_buttons(False)` outside of [for loop](https://github.com/borgbase/vorta/pull/1723/files#diff-1340facde0499b13accf612e9eab00d2c0909090fb7320e61ce2f0cf6282ca02R537), it should stop the constant state change. Perhaps also add a parameter to `info_result` to change state back only conditionally.

diivi:
> How about maintaining a variable with length of archive_names, passing it onto info_result which would substract one each time and enable buttons back when it's 0?

This seems like a reasonable approach, I'll try this and update the PR!

diivi:
It ended up being a little more complex than that 😅, but it works the way we want it to now:

Peek 2023-06-25 20-14

Of course I already had the sizes recalculated so it looks like there's less user feedback, but for no info it looks good.

m3nu (dismissed):

m3nu:
Minor issue, maybe due to rebasing just now: The "Start" button stays disabled after it's done.

Screenshot 2023-07-04 at 19 09 43

diivi:
Updated so that the start backup button gets enabled again after refreshing all archives.

m3nu (approved):

m3nu:
Works as expected locally. Merging. Thanks for your patience on polishing this. 👏

PR #1734, opened 2023-06-19
feat: inline edit for archive renaming

diivi:
### Description
This PR adds the ability to rename archives in the archive tab. When a user double-clicks on the name of an archive, they can edit the name. If the new name is valid, the archive is renamed. If the new name is invalid, an error message is displayed. This feature is implemented using the `cellChanged` signal of the `QTableWidget`. The `rename_action` method has been removed.

### Related Issue
Part of #1723

### Motivation and Context
Faster renaming of archives.

### How Has This Been Tested?
Peek 2023-06-20 02-31

### Types of changes

### Checklist:


tests/test_archives.py
```diff
@@ -168,31 +168,3 @@ def test_archive_delete(qapp, qtbot, mocker, borg_json_output):
     qtbot.waitUntil(lambda: 'Archive deleted.' in main.progressText.text(), **pytest._wait_defaults)
     assert ArchiveModel.select().count() == 1
     assert tab.archiveTable.rowCount() == 1
-
-
-def test_archive_rename(qapp, qtbot, mocker, borg_json_output):
```
diivi, line 173: Will look at some docs online to add a new test here, according to the changes. Any help will also be appreciated!

jetchirag: I tested using simple double mouse click event and it works

```
pos = tab.archiveTable.visualRect(tab.archiveTable.model().index(0, 4)).center()
 qtbot.mouseClick(tab.archiveTable.viewport(), QtCore.Qt.MouseButton.LeftButton, pos=pos)
 qtbot.mouseDClick(tab.archiveTable.viewport(), QtCore.Qt.MouseButton.LeftButton, pos=pos)
```

This will trigger the double click and rename action. You can then send new name and enter key.

diivi: Thanks!

m3nu:
Will review this after some of the other PRs are merged. Else you will have too many conflicts.

diivi:
I tried this:

```python
    pos = tab.archiveTable.visualRect(tab.archiveTable.model().index(0, 4)).center()
    qtbot.mouseClick(tab.archiveTable.viewport(), QtCore.Qt.MouseButton.LeftButton, pos=pos)
    qtbot.mouseDClick(tab.archiveTable.viewport(), QtCore.Qt.MouseButton.LeftButton, pos=pos)
    qtbot.keyClicks(tab.archiveTable.viewport().focusWidget(), new_archive_name)
    qtbot.keyClick(tab.archiveTable.viewport(), QtCore.Qt.Key.Key_Return)

    # Successful rename case
    print(tab.archiveTable.model().index(0, 4).data())
    print(tab.mountErrors.text())
```

but it did not work, it just printed the old name. Am I missing something?

real-yfprojects (commented):
Can you add back the rename button and the context menu action? 

The [KDE HIG](https://develop.kde.org/hig/components/navigation/contextmenu/) states:
> Don’t use context menus as the only way to access functionality. Every item in a context menu must be available via a method that is somehow visible by default

This also applies for this new feature. Inexperienced user might not discover that you can double click the entry. The old buttons should now enter the edit mode of the code cell.

jetchirag:
@diivi It needs some time to update the model. Add this after sending Return key.

```
qtbot.waitUntil(lambda:  tab.archiveTable.model().index(0, 4).data() == new_archive_name, **pytest._wait_defaults)
```

diivi:
Thanks, but I did that first and did not work. It just says 

`E       pytestqt.exceptions.TimeoutError: waitUntil timed out in 20000 milliseconds`

jetchirag:
You missed `focusWidget`:

```
qtbot.keyClick(tab.archiveTable.viewport().focusWidget(), QtCore.Qt.Key.Key_Return)
```

+You can add `main.show()` to see where it's stuck in testing.

diivi:
🤦🏼 it works now, also thanks for the `man.show()` tip, I was looking for something like that!

m3nu:
Worked as expected locally and doesn't add much new code by reusing the existing function. Good job!

m3nu (approved):

m3nu:
Minor comment: Better to make a feature branch in your personal fork of Vorta's repo. Else it can get messy when merging from the `master` branch more than once.

diivi:
Yeah, I do that everytime but missed it for this PR🥲

real-yfprojects (commented):
While your solution will work after some bugs are eliminated, I think this would be the perfect opportunity for implementing a QTableModel for the archive list which would be more elegant and probably easier to maintain.

diivi:
How much time will that take, and will I have to scrape this off completely? Also, I started this 20 days ago, if it was the perfect time to implement this you should've told me a little early (or mentioned it in the task in the first place) 😅

real-yfprojects:
> Also, I started this 20 days ago, if it was the perfect time to implement this you should've told me a little early (or mentioned it in the task in the first place)

You are right. However occupied with other things like mentoring my own contributor I didn't look into this PR in detail. Let's settle on the current approach then without a `QTableModel`.

src/vorta/views/archive_tab.py
```diff
@@ -831,6 +836,40 @@ def cell_double_clicked(self, row, column):
...
+        new_name = item.text()
+        profile = self.profile()
+
+        # if the name hasn't changed or if this slot is called when first repopulating the table, do nothing.
+        if new_name == self.renamed_archive_orginal_name or not self.renamed_archive_orginal_name:
+            return
+
+        if not new_name:
+            self._set_status(self.tr('Archive name cannot be blank.'))
+            return
```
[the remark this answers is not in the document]
diivi: Good point

m3nu:
Clear implementation and works as expected.

Just noticed that the archive duration and archive size get lost when renaming and then doing a "Refresh Archives". Possible that some row in the DB isn't updated?

diivi:
It was working the last time I checked, but now I can't understand why the Size and duration get lost :/

The error occurs here for me - https://github.com/diivi/vorta/blob/48204f8bfd087b0206421c515d3f11bdc803ddaa/src/vorta/borg/info_archive.py#L46

```bash
  File "/home/divyansh/Desktop/vorta/src/vorta/borg/info_archive.py", line 46, in process_result
    archive.name = remote_archive['name']  # incase name changed
AttributeError: 'NoneType' object has no attribute 'name'
```

diivi:
I deleted the database and tested this again, it worked.

Maybe some migration is missing? Can someone with the old database test it and let me know, I'm a bit confused.

src/vorta/views/archive_tab.py
```diff
@@ -831,6 +836,45 @@ def cell_double_clicked(self, row, column):
...
+
+        if not new_name:
+            item.setText(self.renamed_archive_original_name)
+            self._set_status(self.tr('Archive name cannot be blank.'))
+            return
+
+        new_name_exists = ArchiveModel.get_or_none(name=new_name, repo=profile.repo)
+        if new_name_exists is not None:
+            self._set_status(self.tr('An archive with this name already exists.'))
```
bigtedde, line 865: Should `item.setText(self.renamed_archive_original_name)` get called before this line? When I'm testing this, I find the status message is easily missed, and the name stays as the invalid name until an archive refresh.

bigtedde:
> It was working the last time I checked, but now I can't understand why the Size and duration get lost :/
> 
> The error occurs here for me - https://github.com/diivi/vorta/blob/48204f8bfd087b0206421c515d3f11bdc803ddaa/src/vorta/borg/info_archive.py#L46
> 
> ```shell
>   File "/home/divyansh/Desktop/vorta/src/vorta/borg/info_archive.py", line 46, in process_result
>     archive.name = remote_archive['name']  # incase name changed
> AttributeError: 'NoneType' object has no attribute 'name'
> ```

This error occurs during a "recalculate" archive action done after a name change. Separately, I am seeing the size and duration information get lost during the "refresh" archive action.

diivi:
Figured out what the issue is. The Borg [rename command](https://borgbackup.readthedocs.io/en/stable/usage/rename.html) changes the archive's ID too during renaming, so I need to update that here - https://github.com/diivi/vorta/blob/48204f8bfd087b0206421c515d3f11bdc803ddaa/src/vorta/borg/rename.py#L37. How can I get the archive's ID after the rename operation though?

real-yfprojects:
> How can I get the archive's ID after the rename operation though?

Since you know the new name I would think you can run `borg info` or `borg list` with the new name. I don't which of these is faster in this scenario.

src/vorta/views/archive_tab.py
real-yfprojects: Yes, this is still an issue.

diivi:
Changed the code so that the name reverts to the original one if the new name is already taken.
Also updated the `snapshot_id` in the `if` block.

real-yfprojects:
> This might solve the bug, where duration and size is lost when refreshing the archive list.

This is still the case. I think it is linked to `snapshot_id`.

#### Steps to reproduce

1. Rename archive
2. Refresh archive list
3. See that the archive appears but the columns `size`, `duration` and `trigger` are empty.

This doesn't happen when recalculating the archive size of the renamed archive between steps 1 and 2.

diivi:
So should I start an `info_archive` job everytime the rename job finishes? To update the snapshot id for the archive in the database

real-yfprojects:
> So should I start an `info_archive` job everytime the rename job finishes? To update the snapshot id for the archive in the database

If there is no other way to deal with this issue, I would accept running borg once more.

diivi:
I don't think there is - https://borgbackup.readthedocs.io/en/stable/usage/list.html

check `Keys available only when listing files in an archive:`

I can only get the dedup size and duration by listing the archive that was renamed 😞

diivi:
@real-yfprojects I added a call to `refresh_archive_info` after an archive is reamed, now I can't reproduce the error, seems like it's fixed.
Please check and let me know too.

m3nu (approved):
Works as expected, including keeping stats.

Issue #907, opened 2021-03-02
Exclude patterns GUI improvements


m3nu:
Just noticed that Nextcloud (desktop client) comes with a nice list of default exclusions. Maybe something for Vorta to assimilate.

<img width="500" src="https://user-images.githubusercontent.com/3916435/85429630-9745e300-b5b1-11ea-9792-8e376b7f7ed6.jpg" />

I agree that one simple kind of patterns would be enough.

_Originally posted by @m3nu in https://github.com/borgbase/vorta/discussions/838#discussioncomment-369487_

erAck:
BackInTime also has a sensible set of default excludes:

- .gvfs
- .cache/*
- .thumbnails*
- [Tt]rash*
- *.backup*
- *~
- .dropbox*
- /proc/*
- /sys/*
- /dev/*
- /run/*
- /etc/mtab
- /var/cache/apt/archives/*.deb
- lost+found/*
- /tmp/*
- /var/tmp/*
- /var/backups/*
- .Private

m3nu:
We disussed adding defaults before and the feedback wasn't good due to the large variety of OS and use cases.

So **if** we add defaults, they will be OS-specific and disabled by default.

m3nu:
Suggested by @Wikinaut in #970

> I cannot find that borg option --exclude-caches to be used by vorta. Was that intended?
> 
> See https://borgbackup.readthedocs.io/en/stable/usage/create.html?highlight=exclude%20caches
> 
> I suggest to add this as a vorta option which is active by default.

m3nu:
Getting close to starting this feature. Was thinking about this a bit more and would want those features:

- Maintain templates of exclusions and enable them for selected profiles. Like "Dev Cache Files" or "App Cache Files".
- Currently I use 2 groups of exclusions: specific paths (e.g. `~/Library/BlahBlah/Cache` and patterns that should be matched everywhere (e.g. `*/.git`)
- Maybe we also want "Exclude folders with this file"
- Add description to exclusion
- Import and export exclusions as CSV? (pattern, desc)

Anything to add? @Hofer-Julian @bastiencyr ?

Hofer-Julian:
Sounds good.
One idea:

> - Add description to exclusion
> - Import and export exclusions as CSV? (pattern, desc)

How about allowing comments with `#`?
In that case you don't need an extra field for descriptions and the export format would be a list of patterns and comments instead of a csv.

m3nu:
[...]

PR #1742, opened 2023-06-27
feat: exclude gui

diivi:
### Description
Progress updates for #907 will be added here. This is the base implementation which will allow the user to add exclude rules, select exclude presets to include, and edit/copy/paste the exclusions from a "Raw" tab.

TODO:
- [x] Presets Tab and a starting json file for presets.
- [x] Raw tab.
- [x] Actually pass the data to borg.
- [x] Exclude if Present Dialog.

### Related Issue
#907 

### Motivation and Context
Makes it simpler for the user to add excludes with a more intuitive interface.

### How Has This Been Tested?
Testing this manually for all scenarios that I can think of.

[Screencast from 21-08-23 01:12:55 AM IST.webm](https://github.com/borgbase/vorta/assets/41837037/f7bf556a-77a2-4708-a5ba-a5f6bda1f641)

### Types of changes

### Checklist:

***

### Follow ups

- [ ] Preview excludes (dialog)
- [ ] Edit custom patterns inline


src/vorta/views/exclude_dialog.py
```diff
@@ -0,0 +1,256 @@
...
+                    if preset_model:
+                        item.setCheckState(Qt.CheckState.Checked if preset_model.enabled else Qt.CheckState.Unchecked)
+                    else:
+                        item.setCheckState(Qt.CheckState.Unchecked)
+                    # add a link icon to the end of the item
+                    self.exclusionPresetsModel.appendRow(item)
+                    self.allPresets[preset['name']] = {
+                        'patterns': preset['patterns'],
+                        'tags': preset['tags'],
+                        'filename': preset_file,
```
diivi, line 146: this can be removed, I had some other thing in mind earlier (linking the row to a github permalink for the json file).

src/vorta/assets/exclusion_presets/browsers.json
```diff
@@ -0,0 +1,67 @@
...
+            "fm:/.config/chromium/*/Session Storage",
+            "fm:/.config/chromium/*/Service Worker/CacheStorage",
+            "fm:/.config/chromium/*/Application Cache",
+            "fm:/.config/chromium/*/History Index *",
+            "fm:/snap/chromium/common/.cache",
+            "fm:/snap/chromium/*/.config/chromium/*/Service Worker/CacheStorage",
+            "fm:/snap/chromium/*/.local/share/"
+        ],
+        "tags":["application:chromium", "type:browser", "os:linux"],
+        "author": "Divi"
```
Hofer-Julian, line 16: Is the `author` key needed for anything?

src/vorta/assets/exclusion_presets/browsers.json
```diff
@@ -0,0 +1,67 @@
+[
+    {
+        "name": "Chromium cache and config files",
+        "patterns":
+        [
+            "fm:/.config/chromium/*/Local Storage",
```
Hofer-Julian, line 6: What is the meaning of `fm`?

src/vorta/assets/exclusion_presets/browsers.json
diivi: Nope, only to attribute the author.

src/vorta/assets/exclusion_presets/dev.json
```diff
@@ -0,0 +1,35 @@
...
+            "fm:.pyc",
+            "fm:.pyo",
+            "fm:.env",
+            "fm:/.virtualenvs"
+        ],
+        "tags": ["type:dev", "lang:python", "os:linux", "os:macos"],
+        "author": "Divi"
+    },
+    {
+        "name": "Rust artefacts",
```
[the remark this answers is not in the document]
diivi: Ah, I follow British English so that's what google recommended.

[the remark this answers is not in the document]
Hofer-Julian: Wasn't aware that's how it's written in British English. Then let's keep it :)

Hofer-Julian (commented):
I've now tested the current WIP. I like the direction this is going.
My suggestions:
- Let's remove the `Raw` tab. The combination of `Custom` to enter and `Preview` to view covers all use cases of `Raw` as far as I can tell
- Reorder to `Presets`, `Custom`, `Preview`
- Stop adding the `fm` prefix and link to https://docs.python.org/3/library/fnmatch.html. That way we avoid overwhelming the users

diivi:
I am fine with the other points and will wait for other reviews, but I think the Raw rab might be useful for adding multiple patterns together, for example just pasting them in from somewhere, or sharing patterns between users.

I'd prefer the raw tab to quickly add patterns that I'll always need instead of adding patterns again and again through the custom tab.

But that's just one reason to support the Raw tab, I'm open to feedback.

src/vorta/assets/exclusion_presets/browsers.json
real-yfprojects: Does that make sense? Contributors might  modify presets in the future. Furthermore since the author isn't shown in the GUI.

diivi: @m3nu I had the same thought regarding edits and revisions.

diivi:
> That is a good idea GUI wise.

This is just shown in the Preview tab, which should show what gets passed to borg directly anyway.

> We should also support adding non-fm-style patterns.

It's supported, anything that you can put inside a Borg exclusion file is supported (it's visible for the custom presets right now too).

diivi:
>  It is the default, so the behavior doesn't change, it still allows other styles to be specified and we don't need logic to hide complexity from the user.

Oh okay, that sounds good. I thought I'll have to hide it from the user or something. But yes, I can remove it entirely and use prefixes only for the other types (I don't know many use cases so I'll probably just leave it for later).

real-yfprojects:
> It is the default, so the behavior doesn't change

Well borg changed the default style in the past afaik.

Hofer-Julian:
Maybe you can help here @ThomasWaldmann:
Is it safe to assume that `Fnmatch` will continue to be the default style for `--exclude` and `--exclude-from`?

ThomasWaldmann:
@Hofer-Julian There is currently no change planned. But I recently opened a ticket about "patterns docs and reality", so there might be some work on that in general for borg2.

Hofer-Julian:
> @Hofer-Julian There is currently no change planned. But I recently opened a ticket about "patterns docs and reality", so there might be some work on that in general for borg2.

Thank you, @ThomasWaldmann! (that's the link for the people who are curious too https://github.com/borgbackup/borg/issues/7634)

That sounds like @real-yfprojects suggestion to specify the `fm` prefix is indeed the best option.

> I thought I'll have to hide it from the user or something

@diivi I think you don't have to go out of your way to hide the prefix from the user

real-yfprojects:
How do we handle changes to the presets in the future? The user might wonder that something is suddenly excluded or included. Or he won't notice at all.

src/vorta/assets/exclusion_presets/browsers.json
```diff
@@ -0,0 +1,67 @@
...
+            "fm:/.config/chromium/*/Session Storage",
+            "fm:/.config/chromium/*/Service Worker/CacheStorage",
+            "fm:/.config/chromium/*/Application Cache",
+            "fm:/.config/chromium/*/History Index *",
+            "fm:/snap/chromium/common/.cache",
+            "fm:/snap/chromium/*/.config/chromium/*/Service Worker/CacheStorage",
+            "fm:/snap/chromium/*/.local/share/"
```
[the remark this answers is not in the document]
diivi: I haven't tested these paths yet (not started with that phase).

src/vorta/store/models.py
```diff
@@ -108,6 +108,36 @@ class Meta:
...
+    name = pw.CharField(unique=True)
+    enabled = pw.BooleanField(default=True)
+    source = pw.CharField(default='custom')  # custom or preset
+    date_added = pw.DateTimeField(default=datetime.now)
+
+    class Meta:
+        database = DB
+
+
+class RawExclusionModel(BaseModel):
```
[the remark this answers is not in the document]
diivi: Yeah, that makes more sense.

src/vorta/store/models.py
```diff
@@ -108,6 +108,36 @@ class Meta:
...
+    """
+    The raw exclusion patterns that a user adds to a profile will be added here as plaintext.
+    Each profile will have a single associated RawExclusionModel.
+    """
+
+    profile = pw.ForeignKeyField(BackupProfileModel, backref='raw_exclusions')
+    patterns = pw.TextField(default='')
+    date_added = pw.DateTimeField(default=datetime.now)
```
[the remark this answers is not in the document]
diivi: Nothing, will be removed when I move this to BackupProfileModel now.

src/vorta/store/models.py
```diff
@@ -108,6 +108,36 @@ class Meta:
...
+    """
+    If this is a user created exclusion, the name will be the same as the pattern added. For exclusions added from
+    presets, the name will be the same as the preset name. Duplicate patterns are already handled by Borg.
+    """
+
+    profile = pw.ForeignKeyField(BackupProfileModel, backref='exclusions')
+    name = pw.CharField(unique=True)
+    enabled = pw.BooleanField(default=True)
+    source = pw.CharField(default='custom')  # custom or preset
+    date_added = pw.DateTimeField(default=datetime.now)
```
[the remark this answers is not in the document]
diivi: Don't have a preference, but yes I can remove the date. I'll sort by id.

Hofer-Julian:
> How do we handle changes to the presets in the future? The user might wonder that something is suddenly excluded or included. Or he won't notice at all.

I would expect that all patterns are disabled per default.
So if we add new patterns, nothing will change for the user.

src/vorta/store/models.py
```diff
@@ -108,6 +108,36 @@ class Meta:
...
 
+class ExclusionModel(BaseModel):
+    """
+    If this is a user created exclusion, the name will be the same as the pattern added. For exclusions added from
+    presets, the name will be the same as the preset name. Duplicate patterns are already handled by Borg.
+    """
+
+    profile = pw.ForeignKeyField(BackupProfileModel, backref='exclusions')
+    name = pw.CharField(unique=True)
+    enabled = pw.BooleanField(default=True)
```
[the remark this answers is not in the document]
diivi: When I added presets, I found it annoying to manually enable them to see them in the preview tab, adding should ideally enable them by default?

diivi:
> I would expect that all patterns are disabled per default. So if we add new patterns, nothing will change for the user.

I think they mean the case where I enable a preset and then later someone changes that preset from Vorta source to add/remove a pattern. I won't know which files are excluded.

I think they can just check the preview tab for changes, as they should.

src/vorta/assets/exclusion_presets/browsers.json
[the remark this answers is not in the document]
real-yfprojects: :+1: one after another

src/vorta/store/models.py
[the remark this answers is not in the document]
real-yfprojects: Makes sense.

real-yfprojects:
> I think they can just check the preview tab for the final exclusions, as they should.

Requiring users to recheck their excludes after every update isn't very user friendly, is it?

diivi:
True, and for scheduled backups people might not even check the preview tab. I don't really have a solution in mind though :/

src/vorta/store/models.py
```diff
@@ -108,6 +108,36 @@ class Meta:
...
+    source = pw.CharField(default='custom')  # custom or preset
```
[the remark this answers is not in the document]
diivi: Something like this  - https://github.com/coleifer/peewee/issues/630#issuecomment-459404401?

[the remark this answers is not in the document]
real-yfprojects: My idea was using an Enum with String values that are then passed to peewee but your suggestion is also possible.

real-yfprojects:
> True, and for scheduled backups people might not even check the preview tab. I don't really have a solution in mind though :/
> 
> 
Possible solutions include:

-  including such changes only in breaking releases and having a section with breaking changes in the release message

- transferring the rules of the changed preset to the user defined rules when updating.

-  Showing a prominent warning on startup after update

- Only adding new presets (labelled `v2`) instead of changing existing ones

Combinations of these could make sense as well.

diivi:
This can now be manually tested and reviewed before I add tests.

real-yfprojects:
Can you implement the database migrations? I cannot test this at the moment: `sqlite3.OperationalError: no such column: t1.raw_exclusions`

diivi:
> Can you implement the database migrations? I cannot test this at the moment: `sqlite3.OperationalError: no such column: t1.raw_exclusions`

Oh, weird. How did it work for me? Let me check and add the migration.

diivi:
Nevermind, I deleted the database and created it again, that's why it worked. Adding the migration.

diivi:
Should work now.

real-yfprojects:
Looking good. In the following a few things I noticed:

There still is a placeholder text on top of every tab. (I guess you already know)

Can you add a QDialogButtonBox to the bottom of the dialog?

I suggest moving the button for the dialog like so:
grafik

In `Custom` tab hitting *Esc* while adding a new pattern will cancel the editing but the empty row remains. 

Can we allow deleting rows with the *Del* key? And can we allow users to edit existing patterns in `Custom` tab?

When adding an item to `Custom` tab, you can click on the plus again adding an empty item while remaining in the edit mode of the current item.

I think a context menu toggle, delete and copy actions would be useful in `Custom` tab and possibly also `Presets` tab.

A button to copy the contents of the `Preview` tab would be nice as well.

The heading for presets in raw tab is missing a space:
```
#Node Modules and package manager cache
*/node_modules
*.npm
```

Why do you have to save when editing patterns in `Raw` tab?

diivi:
> There still is a placeholder text on top of every tab. (I guess you already know)

Yup, left the trivial stuff for the end :P 

> Can you add a QDialogButtonBox to the bottom of the dialog?

What do you mean? QDialogButtonBox for what?

> I suggest moving the button for the dialog like so:

Done!

> In Custom tab hitting Esc while adding a new pattern will cancel the editing but the empty row remains.

Nice catch, I couldn't find anything online, but after some trial and error I got to a solution.

> Can we allow deleting rows with the Del key?

Done.

> When adding an item to Custom tab, you can click on the plus again adding an empty item while remaining in the edit mode of the current item.

Thanks for the intensive testing, totally missed this. Fixed now.

> I think a context menu toggle, delete and copy actions would be useful in Custom tab and possibly also Presets tab.

Done.

> A button to copy the contents of the Preview tab would be nice as well.

Done.

> The heading for presets in raw tab is missing a space:

Fixed.

> Why do you have to save when editing patterns in Raw tab?

I've changed it to a save on change.


> And can we allow users to edit existing patterns in Custom tab?

This is turning out to be a bit complex because of the empty editable row that we add during the addition of a pattern. The code also gets messier because even adding is technically updating. I'll keep trying for some time and let you know.

real-yfprojects:
> What do you mean? QDialogButtonBox for what?

For closing the dialog. I we wanted we could have a save button for the entire dialog here.

diivi:
> For closing the dialog. I we wanted we could have a save button for the entire dialog here.

I prefer auto save instead of doing something repeatedly😅, especially now that auto save is easier than adding a save button and removing auto save.
I can add a button for closing but will it be worth it when we have the close icon anyway?

real-yfprojects:
> I can add a button for closing but will it be worth it when we have the close icon anyway?

Good point. But I think it makes sense nonetheless.

- recognisable design pattern for modal dialogs
- bigger button, easier to use
- One reads the dialog from top to bottom, its natural to have the close button at the end.

diivi:
Starting with the tests now.

diivi:
Wrote an essential test for this, coverage is at 77%. Let me know other tests which could be necessary.

diivi:
https://github.com/borgbase/vorta/actions/runs/5648079022/job/15299508091#step:8:174

Is the Presets tab not working for MacOS? Julian, yfprojects and I have Linux, so maybe we missed this?

bigtedde:
@diivi and I discovered it was an issued relating to how the different OS's were loading the `exclusion_presets` json files in different orders. Diivi had the idea of loading them in sorted order, which fixes the issue!

diivi:
Thanks for your time @bigtedde!

diivi:
Reminder that this needs an approval (just the heading text for each tab is left), while I work on closing other PRs!

src/vorta/assets/UI/sourcetab.ui
```diff
@@ -112,6 +112,16 @@
           </property>
          </widget>
         </item>
+        <item>
+         <widget class="QPushButton" name="bExclude">
+          <property name="text">
+           <string>Select Exclude Patterns...</string>
```
real-yfprojects, line 118: Please use the unicode ellipsis here.

src/vorta/views/exclude_dialog.py
```diff
@@ -0,0 +1,376 @@
...
+            "Patterns that you add here will be used to exclude files and folders from the backup. For more info on how to use patterns, see the <a href=\"https://borgbackup.readthedocs.io/en/stable/usage/help.html#borg-patterns\">documentation</a>. To add multiple patterns at once, use the \"Raw\" tab."  # noqa: E501
+        )
+        self.exclusionPresetsHelpText.setText(
+            "These presets are provided by the community and are a good starting point for excluding certain types of files. You can enable or disable them as you see fit. To see the patterns that are used for each preset, switch to the \"Preview\" tab after enabling it."  # noqa: E501
+        )
+        self.rawExclusionsHelpText.setText(
+            "You can use this field to add multiple patterns at once. Each pattern should be on a separate line."
+        )
+        self.exclusionsPreviewHelpText.setText(
+            "This is a preview of the patterns that will be used to exclude files and folders from the backup."
```
real-yfprojects, line 130:
```suggestion
            "This is a preview of the patterns that will be passed to borg for excluding files and folders from the backup."
```

src/vorta/views/exclude_dialog.py
```diff
@@ -0,0 +1,376 @@
...
+        self.bAddPattern.clicked.connect(self.add_pattern)
+        self.bAddPattern.setIcon(get_colored_icon('plus'))
+
+        # help text
+        self.customPresetsHelpText.setOpenExternalLinks(True)
+        self.customPresetsHelpText.setText(
+            "Patterns that you add here will be used to exclude files and folders from the backup. For more info on how to use patterns, see the <a href=\"https://borgbackup.readthedocs.io/en/stable/usage/help.html#borg-patterns\">documentation</a>. To add multiple patterns at once, use the \"Raw\" tab."  # noqa: E501
+        )
+        self.exclusionPresetsHelpText.setText(
+            "These presets are provided by the community and are a good starting point for excluding certain types of files. You can enable or disable them as you see fit. To see the patterns that are used for each preset, switch to the \"Preview\" tab after enabling it."  # noqa: E501
```
real-yfprojects, line 124:
```suggestion
            "These presets are provided by the community and are a good starting point for excluding certain types of files. You can enable or disable them as you see fit. To see the patterns that comprise a preset, switch to the \"Preview\" tab after enabling it."  # noqa: E501
```

src/vorta/views/exclude_dialog.py
```diff
@@ -0,0 +1,376 @@
...
+        )
+        self.rawExclusionsHelpText.setText(
+            "You can use this field to add multiple patterns at once. Each pattern should be on a separate line."
+        )
+        self.exclusionsPreviewHelpText.setText(
+            "This is a preview of the patterns that will be used to exclude files and folders from the backup."
+        )
```
real-yfprojects, line 131: Please use `translate` to mark these texts for translation.

real-yfprojects (commented):
The dialog is coming along!

As a user I would expect the context menu actions in `Custom` tab to act on all selected entries.

You are able to open multiple instances of the exclude dialog since  interaction with the main window isn't disabled when opening it. You either have to make the dialog modal or handle this case.

The del key still doesn't work.

Can we allow the user to edit existing presets in `Custom` tab?

diivi:
> You are able to open multiple instances of the exclude dialog since interaction with the main window isn't disabled when opening it. You either have to make the dialog modal or handle this case.

How about this:

```python
        if hasattr(self, '_window') and self._window is not None:
            self._window.close()
```
to close any previous window when the button is clicked again?

m3nu:
Can you fix the merge conflict, so we can get closer to merging this, @diivi ?

diivi:
> What about raising and focussing the existing window when the button is clicked again?

I tried this:
```python
        if hasattr(self, '_window') and self._window is not None:
            self._window.activateWindow()
            self._window.raise_()
            return
```
but it doesn't seem to work as the raised window does not get focus.

diivi:
> As a user I would expect the context menu actions in Custom tab to act on all selected entries.

Done (for remove and toggle, not for copy, like in the archive tab).

> The del key still doesn't work.

Fixed.

> You are able to open multiple instances of the exclude dialog since interaction with the main window isn't disabled when opening it. You either have to make the dialog modal or handle this case.

I tried some solutions from the web, but none seemed to focus the window back, but the window stays on top anyway.
Currently, I've committed my earlier solution that closed and reopened the window, it could be bad UX, but changes are still saved on closing.

> Can we allow the user to edit existing presets in Custom tab?

The last time I tried this, I couldn't figure something out even after spending some time on the problem -https://github.com/borgbase/vorta/pull/1742#issuecomment-1646820298. I'll try again and ask here if I get stuck.

real-yfprojects:
> > Can we allow the user to edit existing presets in Custom tab?
> 
> The last time I tried this, I couldn't figure something out even after spending some time on the problem -[#1742 (comment)](https://github.com/borgbase/vorta/pull/1742#issuecomment-1646820298). I'll try again and ask here if I get stuck.

We can move this into a FR then.

real-yfprojects:
> I tried some solutions from the web, but none seemed to focus the window back, but the window stays on top anyway.
> Currently, I've committed my earlier solution that closed and reopened the window, it could be bad UX, but changes are still saved on closing.

I think you should be able to use

```py
self.setWindowFlags(Qt.WindowType.Window)
self.setWindowModality(Qt.WindowModality.NonModal)
```

in the exclude dialog if you want to prevent that the exclude dialog always stays on top.
Then you can use the following to raise it to the top.

```py
if hasattr(self, '_window') and self._window is not None:
    # Raise window if opened already
    self._window.show()
    self._window.activateWindow()
    self._window.raise_()
    return
```

Alternatively you can use

```py
self.setWindowModality(Qt.WindowModality.ApplicationModal)
```

to prevent input to all other windows while the exclude dialog is open.

diivi:
The first option you shared doesn't seem to work (even when I focus on the window underneath, the dialog doesn't go behind it).
image

But the second solution works flawlessly, adding

> self.setWindowModality(Qt.WindowModality.ApplicationModal)

to the `exclude_dialog` makes it impossible to interact with the main window until it is closed. So I went ahead with that.

m3nu:
Just ran this locally. Looks already very polished. Some thoughts for small changes:

- Not fully clear that an exclude pattern needs to be added in the *Custom* tab. Possible to add a placeholder with an example? Else one may enter a random name with the expectation of adding the actual patterns later.
- For _Presets_, adding a preview in the same line wasn't possible with the Qt elements we have, right? A reasonable tradeoff.
- What happens to existing exclusions? Shouldn't they show up in the *Custom* tab?
- Quitting the app via Cmd-Q (Ctrl-Q) doesn't work when this window is open.
- For macOS you may be able to set the "modal" (??) option for this window. Then it stays attached to the main window.

real-yfprojects:
> * For macOS you may be able to set the "modal" (??) option for this window. Then it stays attached to the main window.

The following resources might help with that:
https://doc.qt.io/qt-5/qdialog.html#open
https://doc.qt.io/qt-5/qt.html#WindowType-enum

For some dialogs we use the window type `Sheet`.

diivi:
> Not fully clear that an exclude pattern needs to be added in the Custom tab. Possible to add a placeholder with an example? Else one may enter a random name with the expectation of adding the actual patterns later.

Any suggestions for a pattern example? I thought of adding node_modules/ there.

> For Presets, adding a preview in the same line wasn't possible with the Qt elements we have, right?

Yeah.

> What happens to existing exclusions? Shouldn't they show up in the Custom tab?

Will a simple string.split("\n") give me all the exclusion patterns that I can then create Custom Exclusion records for? 

> Quitting the app via Cmd-Q (Ctrl-Q) doesn't work when this window is open.

Just checked, but the same thing happens with some other dailogs like the extract dialog, add backup profile dailog, add new repo dialog, add existing repo dialog (and maybe other places). So is this worth solving (I don't know how to yet)?

> For macOS you may be able to set the "modal" (??) option for this window. Then it stays attached to the main window.

I'll check what this is from yf-project's comment above.

diivi:
Made the changes that I mentioned, not sure how I can manually test if the old exclusions are converted to the new ones or not though.

real-yfprojects:
You can use the dev mode of Vorta to create a db with a previous version and then run the Pr from that db.

diivi:
> For some dialogs we use the window type Sheet.

With window type = sheet, I can only move the main application and not the dialog (it stays fixed). Let me know which one I should keep.

Without 
```py
self.setWindowFlags(Qt.WindowType.Sheet)
```
image

With
```py
self.setWindowFlags(Qt.WindowType.Sheet)
```
image

real-yfprojects:
So `Sheet` doesn't work correctly on some linux. In that case we could only set it on MacOS if `Sheet` is actually what @m3nu meant in his review.

m3nu:
We use the "sheet" option in many places and should only have an effect on macOS. So I'd be surprised if it causes an issue for you on Linux. Do you get the same problem with e.g. the *Add Repo* window or *Add SSH key* window? Those all use sheets.

https://doc.qt.io/qt-6/qt.html#WindowType-enum

diivi:
Yup, same thing.

I can't move the "Rename Profile" sheet modal. Only the window underneath. But now that every dialog has windowtype = sheet, maybe I should add it here too?

real-yfprojects:
I just checked on my machine running KDE: I don't see this issue.

src/vorta/store/migrations.py
```diff
@@ -250,6 +251,39 @@ def run_migrations(current_schema, db_connection):
...
+    # add a default exclusion to help the user understand how to use the new exclude GUI
+    if current_schema.version < 24:
+        for profile in BackupProfileModel:
+            previous_exclusions = profile.exclude_patterns.splitlines()
+            for pattern in previous_exclusions:
+                ExclusionModel.create(
+                    profile=profile,
+                    name=pattern,
+                    enabled=True,
+                )
```
real-yfprojects, line 275: This can raise `peewee.IntegrityError: UNIQUE constraint failed: exclusionmodel.name`

real-yfprojects (commented):
In Custom tab one cannot select multiple items at once using `Shift`. Can you fix that? Probably a different selection mode is needed.

src/vorta/store/migrations.py
diivi: Done

real-yfprojects (dismissed):

m3nu:
Looks good. Any way to reduce the line height here? Or could there be multiple lines?

Screenshot 2023-08-21 at 16 45 34

diivi:
[Line Height](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height), which is the padding in this case, or Height of the lines (separators)?

diivi:
How about 5 px (from 20)?
image


or 10:
image

diivi:
Went ahead with 10.

src/vorta/views/exclude_dialog.py
```diff
@@ -0,0 +1,410 @@
...
+                padding: 10px 0px;
+                border-bottom: .5px solid black;
+            }
+            QListView::item:selected {
+                background-color: palette(highlight);
+            }
+            QListView::item::icon {
+                padding-right: 10px;
+            }
+        '''
```
real-yfprojects, line 106: Why do we even need to customly style the widgets?
And when you use the same style for multiple lists you can store the style sheet in a variable and reuse it.

real-yfprojects (commented):
What are the plans with the `exclude if present` files?

diivi:
> What are the plans with the exclude if present files?

I think when we discussed this last time, it was decided that the same dialog will be copied for exclude_if_present later. But I guess if we merge this PR then we won't be able to migrate the old value to the new one.

We also had the option of keeping the exclude_if_present box as is for now, but that might look bad.

src/vorta/views/exclude_dialog.py
diivi: I didn't find a way to style these elements other than this one.

Yes, I can use the same string again.

real-yfprojects: > I didn't find a way to style these elements other than this one (border, icon, item).

Yeah, but why to you want to style them? You override the Qt style used on the different platforms. On KDE Qt apps will use the Breeze theme which results in very nice looking list items, on Gnome one might use Kvantum.

real-yfprojects:
> We also had the option of keeping the exclude_if_present box as is for now, but that might look bad.

For this PR their might not be a better option. The master branch shouldn't contain incomplete features or regressions.

src/vorta/views/exclude_dialog.py
diivi: image
It looked too packed for me, that's why I added the styles. Does it look good for you by default? Maybe then I can remove them.

diivi:
Doesn't look too bad:

image

m3nu:
We should have considered the "Exclude-if-present" box before. Keeping it on the main screen may imply that it's more important than normal exclusions, which isn't the case.

What would be the best option here? Could this be a dialog window with checkboxes, like our exclusions?

diivi:
> What would be the best option here? Could this be a dialog window with checkboxes, like our exclusions?

I can make a new exclude_if_present_dialog with 3 tabs - custom, raw & preview. I guess it won't take much time since I'll reuse most of the code.

src/vorta/views/exclude_dialog.py
real-yfprojects: > It looked too packed for me, that's why I added the styles.

That's true but that's how the cerators of the oxygen style have chosen to design it. Outside my dev environment Vorta's lists look similar to this:

grafik

real-yfprojects:
> I can make a new exclude_if_present_dialog with 3 tabs - custom, raw & preview. I guess it won't take much time since I'll reuse most of the code.

:+1: sounds good

diivi:
I reused the exclude GUI for the "Exclude if present" GUI. I went with using conditionals to change the exclude gui (and exclusions models) first, but that approach made the code way less readable. It looks simpler now. Let me know what you think.


Also, I think I made some mistake with migrations on my end, so I can only run the app when my

```py
SCHEMA_VERSION = 22
```

I pushed with schema_version = 24 though, because that would be the correct one.

diivi:
[Screencast from 25-08-23 02:17:33 AM IST.webm](https://github.com/borgbase/vorta/assets/41837037/0d8c1d09-feb3-44ce-8705-90052aa1fbb5)

Works on my end with correct distinction.

diivi:
@jetchirag can you help me figure out where and how I can move the 2 tests for this feaure? I put them in the tests folder to start, but after your PR got merged they should probably be changed a little and go inside one of /unit or /integration directories.

m3nu:
Just noticed that all my exclude rules went missing after testing this PR. I have some backups of my settings, so no problem. Let's just make sure this doesn't happen for real users.

jetchirag:
@diivi Since this is only testing the interface, you can simply move it to `tests/unit`. I don't think you need to change anything in the test itself.

diivi:
Simply moving these files to the `unit` folder does not work:

```bash

=============================================================== FAILURES ===============================================================
___________________________________________________ test_exclusion_preview_populated ___________________________________________________
CALL ERROR: Exceptions caught in Qt event loop:
________________________________________________________________________________
  File "/home/divyansh/Desktop/vorta/src/vorta/views/schedule_tab.py", line 106, in <lambda>
    self.app.scheduler.schedule_changed.connect(lambda pid: self.draw_next_scheduled_backup())
  File "/home/divyansh/Desktop/vorta/src/vorta/views/schedule_tab.py", line 175, in draw_next_scheduled_backup
    status = self.app.scheduler.next_job_for_profile(self.profile().id)
  File "/home/divyansh/Desktop/vorta/src/vorta/store/models.py", line 238, in profile
    return BackupProfileModel.get(id=self.window().current_profile.id)
RuntimeError: wrapped C/C++ object of type ScheduleTab has been deleted
__________________________________________________________________________
```

jetchirag:
@diivi Check this commit for fix

https://github.com/borgbase/vorta/pull/1749/commits/a56af3ab8279beeb0abd70f076f31651c1ede6fd

src/vorta/store/migrations.py
```diff
@@ -250,6 +252,74 @@ def run_migrations(current_schema, db_connection):
...
+            profile.exclude_if_present = ''
+            profile.exclude_patterns = ''
+            profile.save()
+
+        if ExclusionModel.select().count() == 0:
+            for profile in BackupProfileModel:
+                ExclusionModel.create(
+                    profile=profile,
+                    name='*/node_modules',
+                    enabled=True,
```
real-yfprojects, line 312: I suppose the example shouldn't be enabled by default.

src/vorta/store/migrations.py
```diff
@@ -250,6 +252,74 @@ def run_migrations(current_schema, db_connection):
...
+                )
+
+        if ExcludeIfPresentModel.select().count() == 0:
+            for profile in BackupProfileModel:
+                ExcludeIfPresentModel.create(
+                    profile=profile,
+                    name='.nobackup',
+                    enabled=True,
```
real-yfprojects, line 320: Same here.

src/vorta/store/migrations.py
```diff
@@ -250,6 +252,74 @@ def run_migrations(current_schema, db_connection):
...
+            previous_exclusions = profile.exclude_patterns.splitlines() if profile.exclude_patterns else []
+            for pattern in previous_exclusions:
+                try:
+                    ExclusionModel.create(
+                        profile=profile,
+                        name=pattern,
+                        enabled=True,
+                    )
+                except pw.IntegrityError:
+                    pass
```
real-yfprojects, line 290: Can you add a code comment exlaining what this catch is for?

real-yfprojects (commented):
Looking all good, but the implementation can be enhanced. Currently `exclude_dialog.py` and `exclude_if_present_dialog.py` share a lot of code -> redundancy. For example you don't have to maintain to implementations of `MandatoryInputItemModel`. Instead you can reference the same class in both files. Regarding the duplicate code in `ExcludeDialog` and `ExcludIfPresentDialog` you  can    inheritance to fix that. Move the duplicate code into a common super class and add the custom behaviour like different ui strings by overriding methods in the subclasses.

diivi:
> Move the duplicate code into a common super class

I tried doing this first, but got stuck with multiple inheritance - https://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance. In my case the exclusion dialogs had three classes to inherit from - (ExcludeDialogBase, ExcludeDialogUi, ExclusionDialogBase - which is the superclass). But I could not ininitialise all three with the same params. And qt forces me to use `super`.

Also, I tried moving the `setupUi` and `ExcludeDialogBase, ExcludeDialogUi` classes to the superclass, but I could not understand the order in which they are supposed to be called.

Do you have an example of Qt widget superclasses in the Vorta codebase?

real-yfprojects:
> Do you have an example of Qt widget superclasses in the Vorta codebase?

You can have a look how @jetchirag has done it in #1749 for `DiffResultDialog` and `ExtractDialog`.

m3nu:
Is this still being worked on or abandoned? If the latter, we would need to see if we can finish the work somehow.

m3nu:
Great! Any help, let us know here. Would love to see this included in the next release.

m3nu:
Hey Divi,

Can we merge a simplified version of this with the original scope? I know we changed the scope last minute by including managing those "Exclude if present" files and then asking for subclassing those (both are correct in the long-term). If this is proving too hard (it's also a very niche feature), I'd just skip it and do a new branch and PR with just file exclusions, as originally planned. Same as what you present in your results: https://github.com/diivi/GSoC23-PythonSoftwareFoundation#project-achievements

Let me know what you think. We can then add "Exclude if present" later or the few users who rely on it can use CLI args.

diivi:
Yup, sorry I've been a little busy with college and my internship hunt. I'll try to revert to the original approved version when I get a chance.

m3nu:
@diivi , can you revert this to the exclude GUI only state? With the "exclude if present" text area field removed for now (but data settings still there). I think this would be a good starting conclusion of the project you also worked hard on and probably want to see merged and maintained.

If you're busy and already left for new endeavors, I'll make the change some time after Wednesday next week.

Thanks!

m3nu:
Closing this, since a less ambitious version was merged via #1846. Thanks again to @diivi for contributing this!

PR #1749, opened 2023-07-11
Initial Search box

### LINKED ISSUE #1674: FR: Search a file in diff and extract view
#### The problem
For discussion regarding 
https://github.com/borgbase/vorta/wiki/Google-Summer-of-Code-2023-Ideas#search-a-file-in-diff-and-extract-view

TLDR; There's currently no option to filter/search items in diff or extract. This FR aims to add this function.

#### Requested Solution
Initially proposed GUI:
image

After suggestions from @m3nu and @real-yfprojects, removed the additional filter options as it gets too cluttered. 

> I too think that adding widgets for all filter options isn't viable. I was thinking of implementing something like the filter syntax for github issues.

New design:
image

and a dropdown which automatically adds supported search syntax.
image

Syntax for search pattern:
- change-type:
- starts-with:
- ends-with:

jetchirag:
### Description
WIP. Adds filtering option(s) to Vorta diff and extract view.

### Related Issue
Fixes #1674

image

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/views/diff_result.py
+++ b/src/vorta/views/diff_result.py
@@ -114,6 +114,8 @@ class DiffResultDialog(DiffResultBase, DiffResultUI):
         self.comboBoxDisplayMode.setCurrentIndex(int(diff_result_display_mode))
         self.bFoldersOnTop.toggled.connect(self.sortproxy.keepFoldersOnTop)
         self.bCollapseAll.clicked.connect(self.treeView.collapseAll)
+        # Search widget
+        self.searchWidget.textChanged.connect(self.sortproxy.setFilterFixedString)
 
         self.buttonBox.accepted.connect(self.accept)
         self.buttonBox.rejected.connect(self.reject)
--- a/src/vorta/views/partials/treemodel.py
+++ b/src/vorta/views/partials/treemodel.py
@@ -907,6 +907,7 @@ class FileTreeSortProxyModel(QSortFilterProxyModel):
         """Init."""
         super().__init__(parent)
         self.folders_on_top = False
+        self.searchPattern = ""
 
     @overload
     def keepFoldersOnTop(self) -> bool:
@@ -994,3 +995,32 @@ class FileTreeSortProxyModel(QSortFilterProxyModel):
         data1 = self.choose_data(left)
         data2 = self.choose_data(right)
         return data1 < data2
+
+    def setFilterFixedString(self, pattern: str):
+        """
+        Set the pattern to filter for.
+        """
+        self.searchPattern = pattern
+        self.invalidateRowsFilter()
+
+    def filterAcceptsRow(self, sourceRow: int, sourceParent: QModelIndex) -> bool:
+        """
+        Return whether the row should be accepted.
+        """
+
+        self.setRecursiveFilteringEnabled(True)
+        self.setAutoAcceptChildRows(True)
+
+        if self.searchPattern == "":
+            return True
+
+        sourceModel = self.sourceModel()
+        sourceIndex = sourceModel.index(sourceRow, 0, sourceParent)
+        name = self.extract_path(sourceIndex)
+
+        if self.searchPattern.lower() in name.lower():
+            return True
+
+        # TODO: Implement path: syntax which builds full path of current item and then compares using startWith
+
+        return False
--- a/src/vorta/assets/UI/diffresult.ui
+++ b/src/vorta/assets/UI/diffresult.ui
@@ -115,6 +115,30 @@
      </item>
     </layout>
    </item>
+
+    <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_2">
+     <property name="spacing">
+      <number>4</number>
+     </property>
+     <property name="topMargin">
+      <number>0</number>
+     </property>
+     <item>
+
+      <widget class="QLineEdit" name="searchWidget">
+       <property name="toolTip">
+        <string>Search</string>
+       </property>
+       <property name="placeholderText">
+        <string>Search</string>
+       </property>
+      </widget>
+
+     </item>
+     </layout>
+    </item>
+
    <item>
     <widget class="QTreeView" name="treeView"/>
    </item>


jetchirag:
Any suggestions how I can expand parent from `QSortFilterProxyModel`?

real-yfprojects:
> Any suggestions how I can expand parent from `QSortFilterProxyModel`? 
> 

What do you mean by 'expand' and 'parent'?

jetchirag:
@real-yfprojects When a match is found in an item, I want the parent items of that item to expand to reveal the matched child. 

I've also spent un-necessarily high time thinking about the search syntax. There are lot of good suggestions in the linked issue. I'm thinking this currently:

Search input must start with either a normal string or `re:` depending on which the name will be searched using either `in` or `re.search`. The search input will be split into dictionary afterwards which can contain `path:` and other other syntax as needed. All of this will be joined and filtered using `and` condition.

It will also be easier to create a dialog box using this pattern as suggested in the issue and include regex search. Just wanted to run this by you before I start working on this.

real-yfprojects:
> When a match is found in an item, I want the parent items of that item to expand to reveal the matched child.

You could use `QTreeView::scrollTo`. I like the idea but what about multiple matches?

real-yfprojects:
> Just wanted to run this by you before I start working on this.

Sounds good! Can you describe the syntax in detail and present examples before implementing it?

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +995,32 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
         data1 = self.choose_data(left)
         data2 = self.choose_data(right)
         return data1 < data2
+
+    def setFilterFixedString(self, pattern: str):
+        """
+        Set the pattern to filter for.
+        """
+        self.searchPattern = pattern
+        self.invalidateRowsFilter()
```
real-yfprojects, line 1004: Have you read the notice that `beginResetModel` might be more efficient?

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +995,32 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
         data1 = self.choose_data(left)
         data2 = self.choose_data(right)
         return data1 < data2
+
+    def setFilterFixedString(self, pattern: str):
```
real-yfprojects, line 999: Does overriding this existing method provide a benefit? Otherwise I would choose a more fitting name since we will filte with an elaborate syntax eventually.

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +995,32 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+
+    def filterAcceptsRow(self, sourceRow: int, sourceParent: QModelIndex) -> bool:
+        """
+        Return whether the row should be accepted.
+        """
+
+        self.setRecursiveFilteringEnabled(True)
+        self.setAutoAcceptChildRows(True)
+
+        if self.searchPattern == "":
```
real-yfprojects, line 1014:
```suggestion
        if not self.searchPattern:
```

jetchirag:
It should give you an idea what I'm thinking


Valid:
important.txt path:/home/chirag/  <--- `if search_pattern in name`
re:^doc_.*\.pdf$ path:/home/chirag/work size<:1MB <--- `if re.search(search_pattern, name)`

Invalid:
something re:^doc_.*\.pdf$
path:/home/chirag/ important.txt



I've not look at implementation of size though so I'll have to figure this out when I start implementing this.

src/vorta/views/partials/treemodel.py
jetchirag: I had not, thank you I'll check it.

real-yfprojects:
> It should give you an idea what I'm thinking
> 
> 
> Valid:
> important.txt path:/home/chirag/  <--- `if search_pattern in name`
> re:^doc_.*\.pdf$ path:/home/chirag/work size<:1MB <--- `if re.search(search_pattern, name)`
> 
> Invalid:
> something re:^doc_.*\.pdf$
> path:/home/chirag/ important.txt

Why is the last one invalid? And isn't `path:` the same as `re:^`?

> I've not look at implementation of size though so I'll have to figure this out when I start implementing this.

Yes, start with the path-based filters. Then you could proceed with implemeting the same for the extract view. In the process it would be great if you moved the common GUI and code into a superclass implementation to reduce redundancy. Afterwards you can think about implemeting filtering by other fields.

src/vorta/views/partials/treemodel.py
real-yfprojects: Last paragraph of https://doc.qt.io/qt-6/qsortfilterproxymodel.html#filtering

jetchirag: I'm not sure if it will be actually more efficient. It does mention it will require model to be repopulated and I want to invoke search through a button so it won't actually be called a lot of times (depends on the users though).
Let's leave this review open for now and once all filters have been added, I'll try running some basic tests to see which one performs better.

jetchirag:
> Why is the last one invalid? 

I was thinking allowing `path:` to come **only** after the name would give flexibility and not require us to use `name:` syntax. Otherwise it would be difficult to identify what part of search is path and what is name.

> And isn't path: the same as re:^?

I was thinking `path:` to search the path either using `in` operator or using `glob.glob` to allow more flexibility. While `re:` would just search the name. Reason being regex would require escaping special characters in path so knowledge of regex would be needed for just filtering by path.

I think we will have to construct the path by going up and getting name from parent and then joining to form a path. I'm trying to find if there's a more effecient way. Do you have any clue?

real-yfprojects:
Are you familiar with the concept of user stories?

jetchirag:
> I see, so by default only the `subpath` is matched. Does that mean when entering `a`, `path/a` will be shown but not `path/a/b` ?

No, if something is matched, its parents (`path/`) should be revealed but that item should be expandable to reveal child `/b` as too

> Are you familiar with the concept of user stories?

I'm not but I did a quick lookup.

real-yfprojects:
While finished user stories compose a full description of a (sub)feature from a user perspective, I like to work with their basis that is describing a situation in which our feature might help. This allows us to better understand possible needs for our software and to  design our software to better serve those needs. That's why I propose we think now think of some user (need) stories first before we continue thinking about the details of the search syntax. In the mean time you can work on this step (see previous comment):

> In the process it would be great if you moved the common GUI and code into a superclass implementation to reduce redundancy.

I will start with a story:

As someone that uses Vorta for accessing private backups I want to extract the old version of a specific file which I just modified erroneously. While I know the exact file path I find expanding each parent item in the path by clicking on it, very tedious. Expanding all items doesn't improve the situation since now there are so many elements that I need a lot of patience for finding the right path. I wish Vorta could help me with that.

This user need could be solved in several ways. This new filtering option might not be the best option to address this user story. However we should still describe this story although in the end we might decide that this isn't relevant enough to be addressed or needs a different feature entirely. Let's first write down these stories and think about how to solve them afterwards. Now its your turn!

jetchirag:
I recently had to recover a file `.configuration` from bor

jetchirag:
Pushed commit to move methods I could find similar to base class. 

### A story based on my last experience:

I was working on an app sometimes ago where I needed to restore `.configuration` file from backups. I provided the path as `app/.configuration` to borgmatic which was able to fetch the correct file for me despite not knowing/providing absolute path. I don't want to provide just the file name either as there maybe several files with same name in this project like `__init__.py`. So path option should be flexible.

### Another story: Size based filters

I have a cronjob which deletes all files from Downloads folder which haven't been modified for more than 14 days. I realised a month later I had some important documents I downloaded but never moved. I now want to recover them from my Vorta backup but there are lot of unnecessary stuff in Downloads folder like large application files which I don't need.

They were simple documents and spreadsheet files which I am sure were small, surely under 50MB. I'd want a simple filter in Vorta where I can enter the size range and set path to `/home/chirag/Downloads`. I also haven't read the documentation and don't know how to use correct size format. Is it `MB` or `M` or does it only support bytes? It would be great if there was a dropdown to select the size format and perhaps a path selector.

**Some small points**
- It would be even great if it supports unix like pathname (`/home/*/Documents`) but for personal use which is Vorta I personally don't need it.
- I am not sure how resource intensive / fast adding these features would be as they are essentially run on every row. As the backup grows it would get slower and slower.
- This made me realise, recovering a file requires knowing archive which has it or atleast the dates so you can filter through them. Can we have a feature to search all archives? Though this would be a bigger and seperate feature.

real-yfprojects:
> * I am not sure how resource intensive / fast adding these features would be as they are essentially run on every row. As the backup grows it would get slower and slower.

There is nothing we can do about that. That's part of the nature of filtering/searching. However Qt does implement some optimizations afaik. Those should include only calculating the filter for expanded items.

### Story 5: Name

As a user who has recently moved a file between backups. I want to find out where was located in earlier backups. Borg and Vorta do not detect the renaming or moving of a file (afaik). So filtering diff changes to only show the ones with a specific file name would be useful.

### Story 6: Removed files

As a normal user I want to get an overview of all the files I deleted from my harddrive (between two backups). However sorting by change type still clutters the list (in tree view) with added or modified files. This is because it is only sorted in the sublist of each parent.

### Story 7: Search for parent

As a user who wants to extract a document regarding GSoC I deleted previoiusly, I know it is located in a folder somewhere that contains GSoC in its name. Now I want to enter something simple like `GSoC` in the search bar and have the folder appear. Then I want to be able to expand the correct folder and look for the document in its children.

real-yfprojects:
Can you now lay out your syntax and describe in which way it satisfies the needs of each story? Please also think about how conveniently it solved each problem and whether the resulting order ov convenience is justified.

jetchirag:
```
[ string ] [ -m ] [ -i ] [ -p ] [ -c ] [ -s ]
Note: Search String and other options are all optional

Options/Flags:

	--match -m <type>
	Type of match query
	Supported values:
		in (default): Filename should contain search string
		exact: Filename must match the search string exactly
		re: Use regex to match filename                        <-- Do we need this?

	--ignore-case -i    <-- inspired by grep
	Ignore Case. Supported by `in` and `exact` match type only.	

	--path -p <path>
	Specify path to match. Supports wildcard. Parsed using glob (?)
	Example:
		--path /home/vorta/Documents
		--path /var/www/public/*
		--path */src/*
	
	--change -c <type>
	Only available in Diff View. Supported values:
		A: Match only Added files
		R: Match only Removed files
		...
	
	--size -s <relation[<|>]number[B|KB|MB|GB>]>
	Match by size. Can be combined once to create a range. Range must be closed.
	Example:
		--size <10MB
		--size >1GB
		--size >1GB --size <10GB
```

I am planning on using argparser for this. I was looking into implementing custom with `startsWith`, `in` and so but that would be just reinventing the wheel. I also think many users will be familiar with this type of syntax than the earlier proposed one.

Matching with the stories
### 1. exact path
```
importantfile.txt --path /home/chirag/work/
```
Or
```
--path /home/chirag/work/importantfile.txt
```

Satisfies the need with a clear syntax

### 2. A story based on my last experience
```
.configuration --path */app
# Or
--path */app/.configuration
```
Allows wildcards hence I can match relative path.

### 3. Size based filters
> They were simple documents and spreadsheet files which I am sure were small, surely under 50MB. I'd want a simple filter in Vorta where I can enter the size range and set path to `/home/chirag/Downloads`.

```
--size <50GB --path /home/chirag/Downloads
```

### 4. Name
```
mylogins.txt --match exact
```

### 5. Removed files
```
--change R
```

### 6. Search for parent
> I know it is located in a folder somewhere that contains GSoC in its name
```
GSoC

# was it GSoC, GSOC, gsoc, Gsoc?
gsoc -i
```

Overall I feel this fits every use case, is easier to read and organised. Took some time to write this, let me know what you think.

Some other special use specific options "may" add (from grep):
	-v, --invert-match
	--exclude-dir

real-yfprojects:
I will have a look at this tomorrow.



> I've pushed some code to match filenames and path to stay on the timeline. It's working well but can be easily modified shall we decide to change syntax.

You can implement the search button until we have settled on a syntax.

real-yfprojects:
Can we use an icon for the search button?

jetchirag:
Can we use this icon?

https://fonts.google.com/icons?selected=Material+Symbols+Outlined:search:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=search

real-yfprojects:
Yes, that works! @m3nu how do we have to handle licensing?

real-yfprojects:
I like the syntax you proposed since it can be (partly) implemented by using `argparse`.

> `--size >1GB --size <10GB`

Why not combine those to a single argument: `--size >1GB,<10GB` and allow `--size <=10GB`?

> Specify path to match. Supports wildcard. Parsed using glob (?)

Glob support is great for not so technical users but one should also be able to use a regex to match the path. 

> re: Use regex to match filename                        <-- Do we need this?

I don't think we need this if one can use a regex for `path`.

>```
> .configuration --path */app
> # Or
> --path */app/.configuration
> ```

Why have to ways of doing the same. If we make `name` and `path` exclusive we could add a flag to toggle between them and use `match` for choosing the pattern format for both options.
Like so:

```
[ <search_pattern> [ ( -m | --match ) ( ex | in | re | fm ) ] [ -p | --path ] [ -i | --ignore-case ] ] [ ( -c | --change ) <type> ] ...
```
The naming of the pattern types is inspired by borg exclude patterns since the user has to learn less when using the same design patterns as borg. The user would need to write less when using the borg pattern prefixes like `pf:`, `re:`, ... directly.

For the diff view there also should be an option to filter by balance. For extract view there should be options to filter by health, modification time and size.

I feel like when the path with a regex, vorta shouldn't show all children of a matched file even though they don't match as well. And I had another question do we only want to support joining the filters like `path` or `size` with AND or also OR?

jetchirag:
> And I wondered whether we only want to support joining the filters like path or size with AND or with OR also?

I actually pondered over this for quite a while but settled on AND because that seems more intuitive and I couldn't find personal use case where I would need to use OR.


> Why have to ways of doing the same. If we make name and path exclusive we could add a flag to toggle between them and use match for choosing the pattern format for both options.

I like it but there's a usecase I thought of. If I want to find a file in a given path, I thought we would require to use regex but the wildcard here seems to support nested directories so it should be fine. 

```
print(fnmatch.fnmatch("/home/chirag/docs/latest/test", "/home/chirag/*/test"))
# True
```

What do you think about this?

real-yfprojects:
> If I want to find a file in a given path, I thought we would require to use regex but the wildcard here seems to support nested directories so it should be fine.

Yes, exactly. This is the reasoning for my suggestion to make `path` and `name` mutually exclusive.

jetchirag:
> and allow --size <=10GB?

Since underline size comparison is in bytes, I thought this match would be a rare care.

> I feel like when the path with a regex, vorta shouldn't show all children of a matched file even though they don't match as well.

Any reason for that? By behavior, filter function will stop quering children if parent matches and just return them too. 

I added `path` function as you suggested and also implemented size filter. Pushing the code now.

real-yfprojects:
> Since underline size comparison is in bytes, I thought this match would be a rare care.

Oh, I didn't think of that. But It doesn't hurt either.

> Any reason for that? By behavior, filter function will stop quering children if parent matches and just return them too.

I guess we can leave it like that and change it in the future when users bother.

> > When a match is found in an item, I want the parent items of that item to expand to reveal the matched child.
> 
> You could use `QTreeView::scrollTo`. I like the idea but what about multiple matches?

Something like this would be very useful for Story 1 and also for 4 and 6.

***

I noticed our filters behave very differently in the different view modes of the tree view. However I haven't figured out yet whether this is an issue (or a feature).

jetchirag:
> I noticed our filters behave very differently in the different view modes of the tree view. However I haven't figured out yet whether this is an issue (or a feature).

It's a feature ofcourse 😃
Can you share any example? I found them to be similar but it's only my test data not very vast.

jetchirag:
I found that `changed_size=73746` corresponds to Size Column, and `size=57358` to Balance. So `changed_size` should be the overall new size of file, so should we be comparing size filter with it instead of `size`? 

Do you know if `changed_size` is indeed overall new size? Because it's different  from the size in extract view.

src/vorta/views/partials/treemodel.py
jetchirag: I think resetting model will also loose the checkbox state in extract view. So if someone needs to extract files from multiple searches, it wouldn't work.

jetchirag:
> For the diff view there also should be an option to filter by balance. For extract view there should be options to filter by health, modification time and size.

Only this should be pending now.

real-yfprojects:
> I found that `changed_size=73746` corresponds to Size Column, and `size=57358` to Balance. So `changed_size` should be the overall new size of file, so should we be comparing size filter with it instead of `size`? 
> 
> Do you know if `changed_size` is indeed overall new size? Because it's different  from the size in extract view.

Changed size tells you wether the file size grew or shrinke, since it is calculated by substracting the number of bytes removed from the number of bytes added to the file.

src/vorta/views/partials/treemodel.py
real-yfprojects: Ok, that rules out resetting the model.

real-yfprojects:
> Can you share any example? I found them to be similar but it's only my test data not very vast.

A search string like `word` will show all directories containing that word in tree mode. However in flat mode they usually do not appear because the directories themselves weren't changed but a file within them. The same happens in the simplified tree mode.

jetchirag:
> Changed size tells you wether the file size grew or shrinke, since it is calculated by substracting the number of bytes removed from the number of bytes added to the file.

Looks like `changed_size` is final size and `size` is the difference. On adding content to a file:
```
FileSystemItem<'('home', 'kali', 'vorta', 'source1', 'hello.txt')', 'hello.txt', DiffData(file_type=<FileType.FILE: 1>, change_type=<ChangeType.MODIFIED: 2>, changed_size=2783, size=2277, mode_change=None, owner_change=None, ctime_change=(PyQt6.QtCore.QDateTime(2023, 8, 8, 13, 29, 1, 574), PyQt6.QtCore.QDateTime(2023, 8, 8, 13, 29, 52, 874)), mtime_change=(PyQt6.QtCore.QDateTime(2023, 8, 8, 13, 29, 1, 570), PyQt6.QtCore.QDateTime(2023, 8, 8, 13, 29, 52, 870)), modified=(2530, 253)), []>
```

real-yfprojects:
Yes, basically.

jetchirag:
So should it be <,> DDMMYYYY format?

real-yfprojects:
> So should it be <,> DDMMYYYY format?

Maybe separators in the date? Most europeans write it like dd.mm.yyyy while Americans use mm/dd/yyyy. So you can choose anything that is easily readable (and writable).

jetchirag:
> So you can choose anything that is easily readable (and writable).

Decided to use YYYY-MM-DD since it won't confuse users if date comes first or month
https://www.iso.org/iso-8601-date-and-time-format.html

m3nu:
> Yes, that works! @m3nu how do we have to handle licensing?

Google Material icons are good to use. We already have the attribution and licenses in the project. Just grab the SVG, add the fill attribute, so it can change color and add the copyright header (similar to other Google Material icons). More here: https://vorta.borgbase.com/contributing/#icons

jetchirag:
@real-yfprojects I think only tests are pending now. Can you let me know if anything else is missing.

jetchirag:
I'm stuck a bit with tests. Do you know how I can get all elements from `treeView` Model which are present / accepted by `filterAcceptsRow()`? 

I tried to recursively create list of rows including children but it keeps running into `RuntimeError: wrapped C/C++ object of type ScheduleTab has been deleted` error but even then it would contains all elements filtered or not.

Also, where should tests go? `test_archives` or respective tests for diff and extract?

m3nu:
There should be some existing test modules for extraction and diff. Would look there first.

Regarding the PR itself, it looks like a great addition and worked as expected. Had those thoughts when trying it out:

- There could be a button to quickly clear the search input. That's the first thing I was looking for.
- The "Collapse" button doesn't do anything when "Flat" view is selected. Could be disabled in such cases
- If the icon isn't from Fork/FontAwesome, please add a `<-- comment -->` at the top, saying it's from Google. See some other icons we have from Google Material as example. This is to know where the icon came from originally.

real-yfprojects:
> I tried to recursively create list of rows including children but it keeps running into `RuntimeError: wrapped C/C++ object of type ScheduleTab has been deleted` error but even then it would contains all elements filtered or not.

This error may sometimes be caused in our testing suite by a signal that is still connected to a object from a previous test but I don't know whether that applies in this case too.

jetchirag:
I figured these two lines are causing the above exception (in `unit/conftest.py`)
```
qapp.main_window.deleteLater()
del qapp.main_window
```

I commented them out for now but any idea what is a better fix?

> There could be a button to quickly clear the search input. That's the first thing I was looking for.
Added

> The "Collapse" button doesn't do anything when "Flat" view is selected. Could be disabled in such cases
Since this is not related to this feature, should we open another issue and I'll create a PR for it later on?

real-yfprojects:
> I figured these two lines are causing the above exception (in `unit/conftest.py`)
> ```
> qapp.main_window.deleteLater()
> del qapp.main_window
> ```
> 
> I commented them out for now but any idea what is a better fix?

We need these two lines since we want a fresh mainwindow for each unittest (basically o reset its state). Optimally we would also create a new instance of `VortaApp` but Qt doesn't allow that. 
Since VortaApp references and instanciates other parts of the app this  makes things complicated and leads to above exception.
`deletLater` deletes the C++ and Qt objects while `del` deletes the python object wrapping the C++ one. However this doesn't always work correctly, e. g. in the case of signals. Sometimes a reference of a python object is kept while the C++ object is already deleted. Here this happens with `ScheduledTab`. So you could check whether there is a signal connected to a method of `ScheduledTab` that you can then disconnect in `conftest.py`.

> Since this is not related to this feature, should we open another issue and I'll create a PR for it later on?

:+1:

src/vorta/views/partials/file_dialog.py
```diff
@@ -0,0 +1,145 @@
...
+        if event.key() in [Qt.Key.Key_Return, Qt.Key.Key_Enter] and self.searchWidget.hasFocus():
+            self.submitSearchPattern()
+        else:
+            super().keyPressEvent(event)
+
+    def submitSearchPattern(self):
+        self.sortproxy.setSearchString(self.searchWidget.text())
+
+    def searchStringError(self, error: bool):
+        self.searchWidget.setStyleSheet("border: 2px solid red;" if error else "")
```
real-yfprojects, line 145: This leads to the same issue as in a previous PR: The tooltip also gets the border.

real-yfprojects (commented):
Now it would make sense to rename `FileTreeSortProxyModel` to `FileTreeSortFilterProxyModel`.

You current implementation works very well. However it isn't really modular or using all the powers of OOP. The following diagram proposes an architecture that will not only solve that but also allow the parser to differentiate between the diff and extract views while being easily extendible.
Vorta-Search-UML drawio

Additionally it would be useful when a parse error would be explained by a small message. Keep that in mind when implementing the architecture.

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +999,239 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+            elif len(date) == 2:
+                return [parse_date(date[0]), parse_date(date[1])]
+            else:
+                raise argparse.ArgumentTypeError("Invalid date format. Can only accept two values.")
+
+        parser = argparse.ArgumentParser(description="Search files and folders based on various options.")
+        parser.add_argument("search_string", nargs="*", default=[], help="String to search in the name.")
+        parser.add_argument(
+            "-m", "--match", choices=["in", "ex", "re", "fm"], default=None, help="Type of match query."
+        )  # TODO: Type "regex"
```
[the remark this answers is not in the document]
jetchirag: This was implemented. I'll remove the comment it in next PR.

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +999,239 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+
+        def valid_date_range(value):
+            date = value.split(',')
+
+            if len(date) == 1:
+                return [parse_date(date[0])]
+            elif len(date) == 2:
+                return [parse_date(date[0]), parse_date(date[1])]
+            else:
+                raise argparse.ArgumentTypeError("Invalid date format. Can only accept two values.")
```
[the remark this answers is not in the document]
jetchirag: Since these are very use specific, moving these to static methods

jetchirag:
@real-yfprojects Thank you for the very detailed approach on modularizing this. I just pushed a commit implementing everything. Checking rest of reviews now.

jetchirag:
> Additionally it would be useful when a parse error would be explained by a small message. Keep that in mind when implementing the architecture.

I didn't implement this initially because I ran into issue with error catching

Reference: [argparse exits on error even when exit_on_error=False](https://github.com/python/cpython/issues/103498)

So it's raising `SystemExit` which doesn't contain error description, just exit code.

src/vorta/views/partials/file_dialog.py
```diff
@@ -0,0 +1,145 @@
...
+    Qt,
+    QUrl,
+)
+from PyQt6.QtGui import QKeySequence, QShortcut
+from PyQt6.QtWidgets import QApplication, QDialog, QMenu, QTreeView
+
+from vorta.views.utils import get_colored_icon
+
+
+class BaseFileDialog(QDialog):
```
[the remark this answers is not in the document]
jetchirag: Added.

jetchirag:
> We need these two lines since we want a fresh mainwindow for each unittest (basically o reset its state). Optimally we would also create a new instance of `VortaApp` but Qt doesn't allow that. Since VortaApp references and instanciates other parts of the app this makes things complicated and leads to above exception. `deletLater` deletes the C++ and Qt objects while `del` deletes the python object wrapping the C++ one. However this doesn't always work correctly, e. g. in the case of signals. Sometimes a reference of a python object is kept while the C++ object is already deleted. Here this happens with `ScheduledTab`. So you could check whether there is a signal connected to a method of `ScheduledTab` that you can then disconnect in `conftest.py`.

I figured it out in my Integration PR but it slipped my mind. Thank you for the description.

tests/unit/conftest.py
```diff
@@ -64,8 +64,8 @@ def init_db(qapp, qtbot, tmpdir_factory):
     source_dir = SourceFileModel(dir='/tmp/another', repo=new_repo, dir_size=100, dir_files_count=18, path_isdir=True)
     source_dir.save()
 
-    qapp.main_window.deleteLater()
-    del qapp.main_window
+    # qapp.main_window.deleteLater()
+    # del qapp.main_window
```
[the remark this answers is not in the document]
jetchirag: Added

real-yfprojects:
> I didn't implement this initially because I ran into issue with error catching
> 
> Reference: [argparse exits on error even when exit_on_error=False](https://github.com/python/cpython/issues/103498)
> 
> So it's raising `SystemExit` which doesn't contain error description, just exit code.

That's bad but nothing we can solve in this PR.

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1042,246 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+        if self.searchPattern.change:
+            item_change = item.data.change_type.short()
+
+            if item_change != self.searchPattern.change:
+                return False
+
+        return True
+
+
+class DiffTreeSortFilterProxyModel(FileTreeSortFilterProxyModel):
```
real-yfprojects, line 1158: I think you don't need to have `DiffTreeSortProxyModel` and `DiffTreeSortFilterProxyModel`. You can merge the methods into one class. Or am I overlooking an advantage?

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1042,246 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+
+        return parser
+
+    def filterAcceptsRow(self, sourceRow: int, sourceParent: QModelIndex) -> bool:
+        """
+        Return whether the row should be accepted.
+        """
+
+        if not self.searchPattern:
+            return True
```
real-yfprojects, line 1181: We don't need this here, since super already implements the check.

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1042,246 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+            raise argparse.ArgumentTypeError("Invalid date format. Can only accept two values.")
+
+    def get_parser(self):
+        """Add Extract view specific arguments to the parser."""
+        parser = super().get_parser()
+        parser.add_argument("--healthy", action="store_true", help="Match only healthy items.")
+        parser.add_argument("--unhealthy", action="store_true", help="Match only unhealthy items.")
+        parser.add_argument(
+            "--last-modified",
+            type=ExtractTreeSortFilterProxyModel.valid_date_range,
```
real-yfprojects, line 1254:
```suggestion
            type=self.valid_date_range,
```

This is more pythonic I think + allows overiding

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1042,246 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+                ExtractTreeSortFilterProxyModel.parse_date(date[1]),
+            ]
+        else:
+            raise argparse.ArgumentTypeError("Invalid date format. Can only accept two values.")
+
+    def get_parser(self):
+        """Add Extract view specific arguments to the parser."""
+        parser = super().get_parser()
+        parser.add_argument("--healthy", action="store_true", help="Match only healthy items.")
+        parser.add_argument("--unhealthy", action="store_true", help="Match only unhealthy items.")
```
real-yfprojects, line 1251: I guess, those should be mutually exclusive which is possible with argparse.
https://docs.python.org/3.9/library/argparse.html?#mutual-exclusion

And you can specify a the attribute to store the value in using `dest`. This allows storing the value for both arguments in the same attribute slightly simplifying code below.

src/vorta/views/partials/treemodel.py
```diff
@@ -896,17 +901,60 @@ def headerData(
...
         super().__init__(parent)
+
+        self.setRecursiveFilteringEnabled(True)
+        self.setAutoAcceptChildRows(True)
+
         self.folders_on_top = False
+        self.searchPattern = None
+
+    @staticmethod
+    def parse_size(size):
```
real-yfprojects, line 923:
```suggestion
    def parse_size(size: str):
```

src/vorta/views/partials/treemodel.py
```diff
@@ -896,17 +901,60 @@ def headerData(
...
+            unit = size[-2:]
+            if size[1] == '=':
+                return (size[:2], int(size[2:-2]) * 1024 ** (size_units.index(unit) + 1))
+            else:
+                return (size[0], int(size[1:-2]) * 1024 ** (size_units.index(unit) + 1))
+        except ValueError:
+            raise argparse.ArgumentTypeError("Invalid size format. Must be a number.")
+
+    @staticmethod
+    def valid_size(value):
```
real-yfprojects, line 948:
```suggestion
    def valid_size(value: str):
```

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1042,246 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+class ExtractTreeSortFilterProxyModel(FileTreeSortFilterProxyModel):
+    """
+    Tree model for diff view.
+    """
+
+    def __init__(self, parent=None) -> None:
+        super().__init__(parent)
+
+    @staticmethod
+    def parse_date(value):
```
real-yfprojects, line 1208:
```suggestion
    def parse_date(value: str):
```

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1042,246 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+
+        try:
+            date = datetime.strptime(date, '%Y-%m-%d')
+        except ValueError:
+            raise argparse.ArgumentTypeError("Invalid date format. Must be YYYY-MM-DD.")
+
+        return (sign, date)
+
+    @staticmethod
+    def valid_date_range(value):
```
real-yfprojects, line 1233:
```suggestion
    def valid_date_range(value: str):
```

src/vorta/views/partials/treemodel.py
```diff
@@ -896,17 +901,60 @@ def headerData(
...
+
+    @staticmethod
+    def valid_size(value):
+        """Validate the size string."""
+        size = value.split(',')
+
+        if len(size) == 1:
+            return [FileTreeSortFilterProxyModel.parse_size(size[0])]
+        elif len(size) == 2:
+            return [FileTreeSortFilterProxyModel.parse_size(size[0]), FileTreeSortFilterProxyModel.parse_size(size[1])]
```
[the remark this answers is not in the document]
jetchirag: Makes sense!

src/vorta/views/partials/treemodel.py
jetchirag: I added this because if super returns True (if search is empty) it will still go through child class's custom checks (balance in this case). What do you think?

src/vorta/views/partials/treemodel.py
jetchirag: @real-yfprojects I couldn't find where `DiffTreeSortProxyModel` is defined. I only see `DiffTreeSortFilterProxyModel`

src/vorta/views/partials/treemodel.py
jetchirag: Added mutual exclusion. I am not sure about the second suggestion. Do you mean to use `store_const` and then passing const value so it can be compared like

```
if args.health == "healthy":
```

src/vorta/views/partials/treemodel.py
real-yfprojects: In diff_result.py

src/vorta/views/partials/treemodel.py
real-yfprojects: Makes sense.

src/vorta/views/partials/treemodel.py
real-yfprojects: No, you can use `store_true` and `store _false` while the absence of a health filter will be encoded in the default value `None`

src/vorta/views/partials/treemodel.py
jetchirag: @real-yfprojects I still can't find any mention. Can you check the PR changes once again?

real-yfprojects: https://github.com/borgbase/vorta/blob/7ebc36ee733934a38f7ab6405f3287d12f3908ee/src/vorta/views/diff_result.py#L396

jetchirag: I'm getting circular import issue for `ChangeType` class import in `treemodel.py` from `diff_result.py`. Should I move it inside treemodel and import in the latter one?

jetchirag:
Thanks for patience. 

Added tests for both views. We can add a info icon beside search button that directs user to a syntax page on vorta documentation which explain it. Unfortunately argparse issue is still open so we can't add error info yet.

real-yfprojects:
> We can add a info icon beside search button that directs user to a syntax page on vorta documentation which explain it. Unfortunately argparse issue is still open so we can't add error info yet.

Yes, that would be good!

src/vorta/views/partials/treemodel.py
real-yfprojects: If you only keep https://github.com/borgbase/vorta/blob/7ebc36ee733934a38f7ab6405f3287d12f3908ee/src/vorta/views/diff_result.py#L396
and add the code from https://github.com/borgbase/vorta/blob/8950eb2c94f8abb1a3b14b1413ebd78e40263b01/src/vorta/views/partials/treemodel.py#L1160 to it. You shouldn't get any circular imports.

jetchirag: Please check the implementation in next commit.

jetchirag:
I didn't notice macos tests are failing. Couldn't reproduce locally, still looking.

real-yfprojects:
> I didn't notice macos tests are failing. Couldn't reproduce locally, still looking.

They time out on github. I think I saw the same behaviour in another PR.

m3nu:
The timeout could be a new macos popup. I need to connect to the CI instance and take screenshots to debug this. It's pretty annoying.

src/vorta/views/extract_dialog.py
```diff
@@ -254,11 +176,103 @@ def parse_json_lines(lines, model: "ExtractTree"):
...
+            return False
+
+        model = self.sourceModel()
+        item = model.index(sourceRow, 0, sourceParent).internalPointer()
+
+        if self.searchPattern.healthy is True and not item.data.health:
+            return False
+
+        if self.searchPattern.healthy is False and item.data.health:
+            return False
```
real-yfprojects, line 265:
```suggestion
        if self.searchPattern.healthy is not None
            return item.data.health == self.searchPattern.healthy
```

This is an optional simplification if I am not mistaken.

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1041,116 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+        if self.searchPattern.match is None:
+            self.searchPattern.match = "fm" if self.searchPattern.path else "in"
+
+        if self.searchPattern.path:
+            search_item = item_path
+        else:
+            search_item = item_name
+
+        if self.searchPattern.search_string:
+            # TODO: Move operations on search string to setSearchString method
```
real-yfprojects, line 1120: Is this still open?

real-yfprojects, line 1120: Is this still open?

src/vorta/views/extract_dialog.py
```diff
@@ -254,11 +176,103 @@ def parse_json_lines(lines, model: "ExtractTree"):
 # ---- Sorting ---------------------------------------------------------------
 
 
-class ExtractSortProxyModel(FileTreeSortProxyModel):
+class ExtractSortProxyModel(FileTreeSortFilterProxyModel):
```
real-yfprojects, line 179:
```suggestion
class ExtractSortFilterProxyModel(FileTreeSortFilterProxyModel):
```

src/vorta/views/diff_result.py
```diff
@@ -486,11 +397,46 @@ def size_to_byte(significand: str, unit: str) -> int:
 # ---- Sorting ---------------------------------------------------------------
 
 
-class DiffSortProxyModel(FileTreeSortProxyModel):
+class DiffSortProxyModel(FileTreeSortFilterProxyModel):
```
real-yfprojects, line 400:
```suggestion
class DiffSortFilterProxyModel(FileTreeSortFilterProxyModel):
```

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1041,116 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+        Creates and returns the parser for the search string.
+        """
+        parser = argparse.ArgumentParser(description="Search files and folders based on various options.")
+        parser.add_argument("search_string", nargs="*", default=[], help="String to search in the name.")
+        parser.add_argument(
+            "-m", "--match", choices=["in", "ex", "re", "fm"], default=None, help="Type of match query."
+        )
+        parser.add_argument("-i", "--ignore-case", action="store_true", help="Ignore case.")
+        parser.add_argument("-p", "--path", action="store_true", help="Match by path.")
+        parser.add_argument("-c", "--change", choices=["A", "D", "M"], help="Only available in Diff View.")
```
real-yfprojects, line 1070: This should be moved to `DiffSortProxyModel`.

src/vorta/views/partials/treemodel.py
```diff
@@ -994,3 +1041,116 @@ def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
...
+
+            for filter_size in self.searchPattern.size:
+                if not compare_values_with_sign(item_size, filter_size[1], filter_size[0]):
+                    return False
+
+        if self.searchPattern.change:
+            item_change = item.data.change_type.short()
+
+            if item_change != self.searchPattern.change:
+                return False
```
real-yfprojects, line 1154: This should be moved to `DiffSortProxyModel` too.

real-yfprojects (commented):
It's almost done!

jetchirag:
Hi, thanks for review. I'm having my final exams currently, will update after that.

real-yfprojects:
Thanks and good luck!

src/vorta/views/extract_dialog.py
```diff
@@ -64,70 +62,44 @@ def run(self) -> None:
...
-
-        # add sort proxy model
-        self.sortproxy = ExtractSortProxyModel(self)
-        self.sortproxy.setSourceModel(self.model)
-        view.setModel(self.sortproxy)
-        self.sortproxy.sorted.connect(self.slot_sorted)
-
-        view.setSortingEnabled(True)
+        super().__init__(model)
+        # TODO: disable self.treeView.setTextElideMode(Qt.TextElideMode.ElideMiddle)
```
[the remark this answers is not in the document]
jetchirag: I can't recall correctly but I think I was worried it won't be able to search elided text but it seems to be working fine, only visible text is modified.
image

real-yfprojects:
@jetchirag Nice to see you active again. Did the exams go well?

jetchirag:
@real-yfprojects Thanks, yes. They went pretty well!

jetchirag:
@m3nu @real-yfprojects Wanted to see if we can merge this up before new contributor(s) start their project so there's less conflict.

m3nu:
Can you add some examples as placeholder text in the search bar? I find this feature overly complicated to use considering what it does. This may help a little.

real-yfprojects:
I noticed two bugs:

1. The `fm` match doesn't seem to work. For some reason `test -m fm` is rejected as invalid syntax.
2. When entering an invalid regex with `-m re` the `re` lib will through an uncached exception.

jetchirag:
> I noticed two bugs:
> 
> 1. The `fm` match doesn't seem to work. For some reason `test -m fm` is rejected as invalid syntax.
> 2. When entering an invalid regex with `-m re` the `re` lib will through an uncached exception.

I am not sure why I added condition that fnmatch must only work with `--path`. I've removed it now. Also added check for invalid regex along with a test case for it.

Any idea on how to make it easier to use? I think the default search works well without any flag which most people can use like searching for file name without having to read docs. I can't decide on simple placeholder text.

m3nu:
> How would design it to be less complicated?

I would focus on filtering by (partial) filename mainly. That's intuitive and doesn't need a separate manual. Maybe maybe the size too.

I realize my feedback is late though. 😬 So never mind. I just hope we will be able to maintain this feature in the future.

Do add some examples as placeholder, please. Since nobody will read the manual to filter a file list.

PR #1751, opened 2023-07-12
fix repository compression size

### LINKED ISSUE #1750: Removal of compression field causes stale size being displayed in UI
#### Description

`unique_csize` field was removed in commit https://github.com/borgbase/vorta/commit/f9d12603164bc663263a86abd17502198059a006
I didn't find any reasoning behind this decision and would therefore like someone to clarify.

The _Compressed Size_ field was not removed from the repository view and this causes the following:
- For old repositories before this change, the _Compressed Size_ will show an outdated size of the repository before Vorta was upgraded past this commit and is never updated.
Screenshot 2023-07-11 at 18 39 55

- For new repositories after this change, the _Compressed Size_ displays N/A.
Screenshot 2023-07-11 at 18 40 39

#### Proposed Change
I'd either like to see the _Compressed Size_ field being completely removed or add it again.

#### Environment

- OS: Mac
- Vorta version: 0.8.11
- Installed from: Homebrew
- Borg version: 1.2.4

henry-spanka:
### Description

`unique_csize` field was removed in commit https://github.com/borgbase/vorta/commit/f9d12603164bc663263a86abd17502198059a006 as the field was dropped in Borg V2.

This PR adds a check to only display the field in Borg V1.

### Related Issue

Closes #1750

### Motivation and Context

Currently the Compressed Size field displays N/A for new repos and a stale size for old repos which is never updated.

### How Has This Been Tested?

- Compression size updated again on Borg V1.
- Field hidden on Borg V2.

### Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [X ] I have read the [CONTRIBUTING](https://vorta.borgbase.com/contributing/) guide.
- [ X] My code follows the code style of this project.
- [X] All new and existing tests passed.

### DIFF
--- a/src/vorta/assets/UI/repotab.ui
+++ b/src/vorta/assets/UI/repotab.ui
@@ -421,7 +421,7 @@
           </widget>
          </item>
          <item row="4" column="0">
-          <widget class="QLabel" name="label_12">
+          <widget class="QLabel" name="sizeCompressedLabel">
            <property name="text">
             <string>Compressed Size:</string>
            </property>
--- a/src/vorta/borg/create.py
+++ b/src/vorta/borg/create.py
@@ -36,7 +36,11 @@ class BorgCreateJob(BorgJob):
                 stats = result['data']['cache']['stats']
                 repo = RepoModel.get(id=result['params']['repo_id'])
                 repo.total_size = stats['total_size']
-                # repo.unique_csize = stats['unique_csize']
+
+                # Compressed Repository size only supported in borg v1.
+                if not borg_compat.check('V2'):
+                    repo.unique_csize = stats['unique_csize']
+
                 repo.unique_size = stats['unique_size']
                 repo.total_unique_chunks = stats['total_unique_chunks']
                 repo.save()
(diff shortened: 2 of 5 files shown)


m3nu:
Hi @henry-spanka,

Thanks for this PR! Appreciate the work. Just ran it locally and I think it needs some additional discussion:

- First, the version is set async to avoid blocking the interface. So it's not known when the repo tab is first rendered.
- Then you put the `hide()` functions into `RepoTab.__init__()` which is only run *once* when the version isn't known yet. To make it at least work for subsequent changes, it should go into `RepoTab.populate_from_profile()`

The other checks you added while processing the data should be fine.

In your linked issue, you also suggest removing the field completely. Given the timing issue we have here, maybe that's better? Any replacement we could use instead?

shivansh02:
> Any replacement we could use instead?

Maybe we can display `Compression Size: Calculating` till the version is known and then populate/hide it accordingly?

real-yfprojects:
> Maybe we can display `Compression Size: Calculating` till the version is known and then populate/hide it accordingly?

I like that.

Issue #1778, opened 2023-08-17
FR: Reduce test workflow for each PR/commit


bigtedde:
#### The problem
Our current Github test workflow has 38 checks being ran for each PR/commit. This uses a lot of resources and takes some time to complete. 

#### Requested Solution
Our previous matrix strategy was 4 python versions (3.8 to 3.11) on 2 OS's (linux, mac), with one version of Borg (1.2.4). If we went back to this it would run these 8 checks for unit tests and 8 for integration tests, so 16 total (+ lint). This would reduce the workflow by about 58%. There would remain the option to run all 39 checks manually. 

#### Alternatives
Open to feedback and alternatives 😀
@m3nu @real-yfprojects @jetchirag 

#### Additional context
Addresses [this comment](https://github.com/borgbase/vorta/pull/1716#issuecomment-1676435290).

m3nu:
In favor of reducing default tests and running the rest only when merging into master or manually.

For the latest 2 PRs I've looked at, all had a test time out.

real-yfprojects:
Also we can enable github's merge queue features which helps with merging multiple PRs after another while having to wait for ci each time.

> For the latest 2 PRs I've looked at, all had a test time out.

Will reducing the test matric solve that? I would think some step in the workflow is taking too long if that problem occurs.

Issue #1754, opened 2023-07-18
FR: Improve Test Coverage: A Test Plan


bigtedde:
#### The problem
Vorta's test coverage is currently 72% according to the `coverage` tool. I would like to increase that coverage to 80% or higher by parameterizing some of the current tests, as well as creating additional unit tests.

#### Requested Solution
```
pytest --cov=vorta --cov-report term-missing
```
This command yields a coverage report detailing the current code coverage of 72% as well as the lines in each file that are being missed by the current set of tests. Using this coverage report, I created a simple spreadsheet detailing all files with a code coverage of <70% and the areas of opportunity for additional tests. These files offer the greatest opportunity to increase code coverage, specifically through unit testing of currently untested functions/methods. 

![image](https://github.com/borgbase/vorta/assets/13988300/5762d19d-9cd1-4efd-a2de-5b0cff48a264)

#### Additional context
@jetchirag is currently working with the test suite on issues #1716 and #1711, so I will wait until after he is finished before progressing to far with this project.

m3nu:
Our testing expert has entered the room. He honed his expertise on thousands of Borg tests. So this is easy for him. 😄

diivi:
@bigtedde I might need your help with the tests for #1742 too 😉, and thanks for the great work!

jetchirag:
> @jetchirag is currently working with the test suite on issues https://github.com/borgbase/vorta/pull/1716 and https://github.com/borgbase/vorta/issues/1711, so I will wait until after he is finished before progressing to far with this project.

I am mostly done with the task. Currently all tests are passing too but they are mostly identical to the unit tests so coverage won't really be effected much except some functions like local folder selector, extract etc. It would be great if you can have a look them though and we can cover the part which can't be included in the unit tests there.

bigtedde:
> I might need your help with the tests for #1742 too 😉, and thanks for the great work!

I would be happy to help with testing your exclude GUI. I took a look at your PR - maybe there is a need for path/pattern testing?

diivi:
> maybe there is a need for path/pattern testing? 

I'll probably just text all 3 tabs and if they alter the preview tab correctly.

And what gets passed to Borg.

After that everything should be handled by Borg, so I don't think testinf if a particular pattern works or not should be necessary.

bigtedde:
[...]

PR #1780, opened 2023-08-17
PR: Reduce test workflow for each PR/commit

bigtedde:
### Description
#1778 
#1754 
@m3nu , @jetchirag , @real-yfprojects 

### Types of changes

### Checklist:

### DIFF
--- a/.github/scripts/generate-matrix.sh
+++ b/.github/scripts/generate-matrix.sh
@@ -6,25 +6,25 @@ if [[ "$event_name" == "workflow_dispatch" ]] || [[ "$branch_name" == "master" ]
       "python-version": ["3.8", "3.9", "3.10", "3.11"],
       "os": ["ubuntu-latest", "macos-latest"],
       "borg-version": ["1.2.4"]
-    }' > matrix-unit.json
+    }' | jq -c . > matrix-unit.json
 
     echo '{
       "python-version": ["3.8", "3.9", "3.10", "3.11"],
       "os": ["ubuntu-latest", "macos-latest"],
       "borg-version": ["1.1.18", "1.2.2", "1.2.4", "2.0.0b5"],
       "exclude": [{"borg-version": "2.0.0b5", "python-version": "3.8"}]
-    }' > matrix-integration.json
+    }' | jq -c . > matrix-integration.json
 
 elif [[ "$event_name" == "push" ]] || [[ "$event_name" == "pull_request" ]]; then
     echo '{
       "python-version": ["3.8", "3.9", "3.10", "3.11"],
       "os": ["ubuntu-latest", "macos-latest"],
       "borg-version": ["1.2.4"]
-    }' > matrix-unit.json
+    }' | jq -c . > matrix-unit.json
 
     echo '{
       "python-version": ["3.10"],
       "os": ["ubuntu-latest", "macos-latest"],
       "borg-version": ["1.2.4"]
-    }' > matrix-integration.json
+    }' | jq -c . > matrix-integration.json
 fi
--- a/.github/scripts/generate-matrix.sh
+++ b/.github/scripts/generate-matrix.sh
@@ -24,7 +24,7 @@ elif [[ "$event_name" == "push" ]] || [[ "$event_name" == "pull_request" ]]; the
 
     echo '{
       "python-version": ["3.10"],
-      "os": ["ubuntu-latest", "macos-latest"],
+      "os": ["ubuntu-latest"],
       "borg-version": ["1.2.4"]
     }' | jq -c . > matrix-integration.json
 fi
--- a/.github/scripts/generate-matrix.sh
+++ b/.github/scripts/generate-matrix.sh
@@ -1,0 +1,11 @@
+event_name="$1"
+branch_name="$2"
+
+if [[ "$event_name" == "workflow_dispatch" ]] || [[ "$branch_name" == "master" ]]; then
+    echo '{"python-version": ["3.8", "3.9", "3.10", "3.11"], "os": ["ubuntu-latest", "macos-latest"], "borg-version": ["1.2.4"]}' > matrix-unit.json
+    echo '{"python-version": ["3.8", "3.9", "3.10", "3.11"], "os": ["ubuntu-latest", "macos-latest"], "borg-version": ["1.1.18", "1.2.2", "1.2.4", "2.0.0b5"]}' > matrix-integration.json
+
+elif [[ "$event_name" == "push" ]] || [[ "$event_name" == "pull_request" ]]; then
+    echo '{"python-version": ["3.8", "3.9", "3.10", "3.11"], "os": ["ubuntu-latest", "macos-latest"], "borg-version": ["1.2.4"]}' > matrix-unit.json
+    echo '{"python-version": ["3.10"], "os": ["ubuntu-latest", "macos-latest"], "borg-version": ["1.2.4"]}' > matrix-integration.json
+fi
--- a/.github/scripts/generate-matrix.sh
+++ b/.github/scripts/generate-matrix.sh
@@ -2,10 +2,29 @@ event_name="$1"
 branch_name="$2"
 
 if [[ "$event_name" == "workflow_dispatch" ]] || [[ "$branch_name" == "master" ]]; then
-    echo '{"python-version": ["3.8", "3.9", "3.10", "3.11"], "os": ["ubuntu-latest", "macos-latest"], "borg-version": ["1.2.4"]}' > matrix-unit.json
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -26,16 +26,59 @@ jobs:
         shell: bash
         run: make lint
 
+  prepare-unit-test-matrix:
```
bigtedde, line 29: Prepares matrix's based on push/pull request vs workflow dispatch. Note: matrix's are the same for unit tests, but defined separately in case changes need to be made in the future.

.github/workflows/test.yml
```diff
@@ -78,18 +121,12 @@ jobs:
...
     strategy:
       fail-fast: false
-
-      matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
-        os: [ubuntu-latest, macos-latest]
-        borg-version: ["1.1.18", "1.2.2", "1.2.4", "2.0.0b5"]
-        exclude:
-          - borg-version: "2.0.0b5"
-            python-version: "3.8"
```
bigtedde, line 92: Biggest hurdle I ran into with the integration tests is excluding the combination of Borg 2.0.0b5 with python 3.8. Skipped this for now.

bigtedde:
Overall, this works as intended with 2 exceptions:

1) excluding combination of Borg 2.0.0b5 with python 3.8
2) received a deprecation warning:
```
The `set-output` command is deprecated and will be disabled soon. 
Please upgrade to using Environment Files. 
For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

bigtedde:
With this matrix, all jobs completed in 7m 57s

.github/workflows/test.yml
```diff
@@ -26,16 +26,59 @@ jobs:
...
+    outputs:
+      matrix: ${{ steps.set-unit-test-matrix-push-pr.outputs.matrix || steps.set-unit-test-matrix-workflow-dispatch.outputs.matrix }}
+    steps:
+      - name: Set unit test matrix for push or pull_request
+        if: github.event_name == 'push' || github.event_name == 'pull_request'
+        id: set-unit-test-matrix-push-pr
+        run: |
+          echo "::set-output name=matrix::{ \
+                  \"python-version\": [\"3.8\", \"3.9\", \"3.10\", \"3.11\"], \
+                  \"os\": [\"ubuntu-latest\", \"macos-latest\"], \
```
m3nu, line 40: I guess this workaround is needed and this can't be more readable?

Could we have 2 separate files for this and import common tasks? I recall Chirag did it like this?

bigtedde:
Welp, a *few* trial and errors later, and I think this is in a good spot. I made a script that generates the matrices for unit and integration tests based on the type of request (PR, push, workflow dispatch) and this script can be changed easily if we want to tinker in the future. Furthermore, I got past the deprecation warning that I was having before.

The only remaining issue is that I am unable to find a way to exclude the combination of Borg v2.0.05b with python 3.8

.github/workflows/test.yml
bigtedde: I created a separate script for the creation of these matrices and imported in the unit and integration test matrix when needed Hopefully this will be much more readable and easy to adjust if needed in the future.

.github/workflows/test.yml
real-yfprojects: Doesn't adding the exclude to the matrix work?

.github/scripts/generate-matrix.sh
```diff
@@ -0,0 +1,10 @@
+event_name="$1"
+
+if [[ "$event_name" == "push" ]] || [[ "$event_name" == "pull_request" ]]; then
```
real-yfprojects, line 3: For a push to master I would run all tests.

m3nu: Agree. That would mainly automate it.

- only run a few select tests for PR commits.
- run all tests when merging into master.

bigtedde: Okay, I have updated the matrix script to first check "is it a manual workflow dispatch OR a request from master branch" and if so fun all tests. In all other cases (push/PR/non master) run the subset of tests.

.github/workflows/test.yml
bigtedde: Unfortunately it does not work with the way I'm dynamically creating the matrix, this only works when the matrix is static I believe. However, I have included a new 'if' conditional that I believe accomplishes the same thing.

.github/workflows/test.yml
```diff
@@ -78,18 +100,17 @@ jobs:
...
 
   test-integration:
+    needs: prepare-matrix
     timeout-minutes: 20
     runs-on: ${{ matrix.os }}
+    if: >-
+      !(
+        fromJson(needs.prepare-matrix.outputs.matrix-integration).python-version == '3.8' &&
+        fromJson(needs.prepare-matrix.outputs.matrix-integration).borg-version == '2.0.0b5'
+      )
```
bigtedde, line 110: I theory this will work the same way as an `exclude` but for dynamically generated matrices, the only difference being that the combinations will still show up as a job, they should just not execute.

.github/workflows/test.yml
```diff
@@ -49,9 +70,17 @@ jobs:
...
+          if [ "$github.event.inputs.debug_enabled" == "true" ]; then
+            echo "Debugging is enabled"
+          else
+            echo "Debugging is not enabled"
+          fi
+
       - name: Setup tmate session
         uses: mxschmitt/action-tmate@v3
-        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled }}
+        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled == 'true' }}
```
bigtedde, line 83: This corrects a pretty big error I found while testing. In previous conditional debugging would always evaluate to true:

```
if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled }}
```

This is because the call returns a string which is truthy. With this, a manual workflow would always be ran in debugging mode because that check was always evaluating to true.

bigtedde: Another thing I found when running in debug mode is that the tests would always timeout (execution time > 20 minutes). Should we consider raising the execution timeout if debug mode is enabled?

bigtedde:
Okay, everything now works as expected and without any deprecation warnings. I implemented a new way of doing the "exclude" by creating a step that validates the matrix combinations. Although it may appear slightly clunky now, if we need to add future excluded combinations this step will be easy to edit/maintain.

.github/workflows/test.yml
real-yfprojects: I think that's because of the remote shell created. It stops execution and allows you to connect to the box.

.github/scripts/generate-matrix.sh
```diff
@@ -0,0 +1,11 @@
+event_name="$1"
+branch_name="$2"
+
+if [[ "$event_name" == "workflow_dispatch" ]] || [[ "$branch_name" == "master" ]]; then
+    echo '{"python-version": ["3.8", "3.9", "3.10", "3.11"], "os": ["ubuntu-latest", "macos-latest"], "borg-version": ["1.2.4"]}' > matrix-unit.json
+    echo '{"python-version": ["3.8", "3.9", "3.10", "3.11"], "os": ["ubuntu-latest", "macos-latest"], "borg-version": ["1.1.18", "1.2.2", "1.2.4", "2.0.0b5"]}' > matrix-integration.json
```
real-yfprojects, line 6: Did you write the exclude like that when testing:

```suggestion
    echo '{"python-version": ["3.8", "3.9", "3.10", "3.11"], "os": ["ubuntu-latest", "macos-latest"], "borg-version": ["1.1.18", "1.2.2", "1.2.4", "2.0.0b5"], "exclude": [{"borg-version":"2.0.0b5, "python-version":"3.8"}]}' > matrix-integration.json
```

.github/workflows/test.yml
bigtedde: Okay, guess the timeout is fine as is then.

.github/scripts/generate-matrix.sh
bigtedde: @real-yfprojects This worked like a charm :)

m3nu:
Anything missing here before merging? Could we limit the "short" integration tests to Linux only?

bigtedde:
> Anything missing here before merging? Could we limit the "short" integration tests to Linux only?

I just made that change to the matrix script, and marked this PR as ready for review.

m3nu (approved):

m3nu:
Thanks a lot for taking care, @bigtedde . Will merge this and then we see how it goes in practice.

Issue #1792, opened 2023-08-31
Archive 'delete' action allowed via context menu during background operations


bigtedde:
#### Description
When testing the archive tab context menu, I found the "delete" action is allowed via the context menu, even when it is disabled due to an ongoing background operation. This can lead to out of bound exceptions if the same archive is deleted twice, as show below:

![delete_issue](https://github.com/borgbase/vorta/assets/13988300/efb0426b-b6ee-47d8-8f5e-e2879934be74)

### Reproduction steps:

1. Open Vorta and navigate tot he Archive tab
2. Select the first archive, and click 'delete'. Confirm that you wish to delete.
3. While the archive delete action is occurring, observe that the delete button is now disabled.
4. While the archive delete action is occurring, right click the archive being deleted and observe that the delete action is available via this menu.
5. While the archive delete action is occurring, right click the archive being deleted and select "delete". Confirm that you wish to delete.
6. Wait for archive delete actions to complete
7. Observe that an out of bounds exception has occurred, causes by attempting to delete an archive that no longer exists.

#### Environment

- OS: Mac OS Ventura 13.5.1
- Vorta version: 0.9.0
- Installed from: Source
- Borg version: 1.2.4

#### Logs
```
023-08-31 13:42:56,458 - vorta.borg.borg_job - INFO - Running command /opt/homebrew/bin/borg delete --info --log-json ssh://vwxe9eus@vwxe9eus.repo.borgbase.com/./repo::This is archive 6
2023-08-31 13:42:56,626 - root - CRITICAL - Uncaught exception, file a report at https://github.com/borgbase/vorta/issues/new/choose
Traceback (most recent call last):
  File "/Users/ted/Desktop/vorta_workspace/vorta/env/lib/python3.11/site-packages/peewee.py", line 7117, in get
    return clone.execute(database)[0]
           ~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/Users/ted/Desktop/vorta_workspace/vorta/env/lib/python3.11/site-packages/peewee.py", line 4481, in __getitem__
    return self.row_cache[item]
           ~~~~~~~~~~~~~~^^^^^^
IndexError: list index out of range

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/ted/Desktop/vorta_workspace/vorta/src/vorta/views/archive_tab.py", line 945, in delete_result
    ArchiveModel.get(name=archive).delete_instance()
[...]

PR #1793, opened 2023-09-01
Bound archive buttons enable/disable to context menu actions

bigtedde:
### Description
The context menu that appears when you right click on an archive has its own rules for when actions are enabled or disabled, separate from the rules for the buttons in the archive tab, even though they perform the same actions. I do not believe there is a reason for these buttons to ever be enabled in the context menu and disabled on the archive tab or vise versa.

In this PR, I have bound the "enabled" or "disabled" state of the actions in the archive right-click context menu to the state of their button equivalent in the archive tab. This will improve consistency, and prevent issues like the one I encountered [here](https://github.com/borgbase/vorta/issues/1792).

### Related Issue
#1792 

### Motivation and Context
Improve consistency and prevent errors

### How Has This Been Tested?
Tested locally, and passes Github test workflow

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -382,7 +382,8 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
 
             for index in range(layout.count()):
                 widget = layout.itemAt(index).widget()
-                widget.setToolTip(self.tooltip_dict.get(widget, ""))
+                if widget is not None:
+                    widget.setToolTip(self.tooltip_dict.get(widget, ""))
 
             # refresh bMountArchive for the selected archive
             self.bmountarchive_refresh()
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -154,7 +154,7 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         self.app.paletteChanged.connect(lambda p: self.set_icons())
 
     def set_icons(self):
-        "Used when changing between light- and dark mode"
+        """Used when changing between light- and dark mode"""
         self.bCheck.setIcon(get_colored_icon('check-circle'))
         self.bDiff.setIcon(get_colored_icon('stream-solid'))
         self.bPrune.setIcon(get_colored_icon('cut'))
@@ -189,11 +189,11 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         # archive actions
         button_connection_pairs = [
             (self.bRefreshArchive, self.refresh_archive_info),
+            (self.bDiff, self.diff_action),
             (self.bMountArchive, self.bmountarchive_clicked),
             (self.bExtract, self.extract_action),
             (self.bRename, self.cell_double_clicked),
             (self.bDelete, self.delete_action),
-            (self.bDiff, self.diff_action),
         ]
 
         for button, connection in button_connection_pairs:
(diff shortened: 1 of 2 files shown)


src/vorta/views/archive_tab.py
```diff
@@ -183,46 +183,22 @@ def archiveitem_contextmenu(self, pos: QPoint):
...
-
-        if not (self.repoactions_enabled and len(selected_rows) <= 1):
-            for action in archive_actions:
-                action.setEnabled(False)
-
-        # diff action
-        menu.addSeparator()
-        diff_action = QAction(self.bDiff.icon(), self.bDiff.text(), menu)
-        diff_action.triggered.connect(self.diff_action)
-        menu.addAction(diff_action)
```
bigtedde, line 222: I guess, ideally, we would have the action menu mimic the layout of the button menu, with a separator before 'diff' and moving 'delete' to the end?

real-yfprojects: That would make sense although we should review the current button layout. I initialls designed it so that it distinguishes between single and multi archive options. But maybe there is a better button order.

bigtedde: > I initially designed it so that it distinguishes between single and multi archive options

Let's run with that idea. Maybe list them by number of archives required for the action?
- Single archive only actions: 'mount', 'extract', 'rename'
- two archive only actions: 'diff'
- any number archive actions: 'recalculate', 'delete'

Separators between each of these sections, maybe?

Thoughts?

real-yfprojects: Categorizing by impact would also be possible.

1. Recalculate
2. Diff

Those only extract data of vorta.

3. mount
4. extract

Those do not alter the repo but modify the filesystem

5. rename

This is reversible.

6. delete

This deletes data.

bigtedde: I can wrap these changes into this PR. Would we like separators between those categories, or just ordered in a list as you outlined?

Same list order for the context menu as well?

real-yfprojects: Which of these two options do you prefer?



> Would we like separators between those categories, or just ordered in a list as you outlined?

I feel like separators or increased spacing would clutter the ui too much.

> Same list order for the context menu as well?

Recognisable patterns repeated throughout the application usually improves the UX.

bigtedde: > Which of these two options do you prefer?

Categorizing by impact as you suggest makes logical sense, and I like 'delete' being the last option as that is intuitive.

I will remove the serparators and match the ordering for both button and context menu 👍

bigtedde: Okay, the requested changes have been made. The only difference now between the context menu and the buttons is the 'copy' context menu action. Would you like there to be a 'copy' button as well, or leave it as it is?

bigtedde:
I have tested this locally and everything works as intended. This solves the issue of mismatched available actions in the context menu vs the archive buttons. The buttons are categorized by impact as discussed, and the separators have have been cleaned up.

src/vorta/views/archive_tab.py
```diff
@@ -183,46 +183,22 @@ def archiveitem_contextmenu(self, pos: QPoint):
...
+            (self.bDiff, self.diff_action),
+            (self.bMountArchive, self.bmountarchive_clicked),
+            (self.bExtract, self.extract_action),
+            (self.bRename, self.cell_double_clicked),
+            (self.bDelete, self.delete_action),
+        ]
+
+        for button, connection in button_connection_pairs:
+            action = menu.addAction(button.icon(), button.text(), connection)
+            action.setEnabled(button.isEnabled())
```
real-yfprojects, line 201: Clever rewrite!

real-yfprojects (commented):
Can you align the archive actions button to the top?

m3nu (approved):

Issue #1801, opened 2023-09-01
SSH key combobox doesn't update when new keys are added


bigtedde:
#### Description

When testing the SSH key combobox, I found that the list of keys does not get updated to include newly added keys until after a Vorta restart. 

![ssh_not_added_gif](https://github.com/borgbase/vorta/assets/13988300/7ada7907-1873-4709-ab2e-e73158a73084)

#### Reproduction Steps
1. Open Vorta
2. Locate the "SSH Key" box in the "Repository" tab
3. Click the "+" button to add a new key
4. Enter a name for the output file, and select "Generate"
5. Close the dialogue box by clicking "cancel"
6. Click dropdown next to the "SSH Key" box to show all SSH keys
7. Observe that the newly generated key is not listed in the drop down

#### Environment

- OS: 13.5.1
- Vorta version: 0.9.0
- Installed from: Source
- Borg version: 1.2.4

#### Logs

No log output related to this issue.

bigtedde:
issue resoled by #1802

PR #1802, opened 2023-09-02
'Add SSH Key' Improvements

bigtedde:
### Description
Issue described here: #1801 

In addition to fixing the issue of adding keys to the dropdown right away, I tweaked the process for adding a key such that now when the user adds a new key, they receive a QMessageBox alert informing them and the SSH dialogue box closes. Previously, the SSH dialogue box would not close unless the user clicked 'cancel', which seems counterintuitive. Here is a demo of the changes:

ssh

### Related Issue
#1801 

### Motivation and Context
Greater consistency and (hopefully) improved user experience.

### How Has This Been Tested?
Locally

### Types of changes

### Checklist:

- [X] I have read the [CONTRIBUTING](https://vorta.borgbase.com/contributing/) guide.

### DIFF
--- a/src/vorta/views/ssh_dialog.py
+++ b/src/vorta/views/ssh_dialog.py
@@ -1,6 +1,6 @@
 import os
 
-from PyQt6 import uic
+from PyQt6 import QtCore, uic
 from PyQt6.QtCore import QProcess, Qt
 from PyQt6.QtWidgets import QApplication, QDialogButtonBox
 
@@ -11,6 +11,8 @@ SSHAddUI, SSHAddBase = uic.loadUiType(uifile)
 
 
 class SSHAddWindow(SSHAddBase, SSHAddUI):
+    create_ssh_key_message = QtCore.pyqtSignal(int, str)
+
     def __init__(self):
         super().__init__()
         self.setupUi(self)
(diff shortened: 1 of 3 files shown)


bigtedde:
@m3nu @real-yfprojects What do you think of these changes? Is this a better experience for the user than the current behavior?

For now, the these changes break a couple tests. If we like the changes then I will be happy to refactor those tests accordingly.

m3nu:
Shouldn't it close after the key is added.

But yes, the SSH stuff can be a barrier for new users, so any improvement is welcome.

Still need to try this locally.

bigtedde:
> Shouldn't it close after the key is added.

Yes, I believe it should. However, in the current implementation [it does not](https://github.com/borgbase/vorta/issues/1801)

> Still need to try this locally.

Please do - I'd appreciate hearing how to further improve it.

real-yfprojects:
> they receive a QMessageBox alert informing them

I don't think we need that since the button already states what is going to happen.

bigtedde:
> I don't think we need that since the button already states what is going to happen.

Good point. Maybe only show a message if the key creation fails?

real-yfprojects:
Sounds reasonable 👍

bigtedde:
@real-yfprojects I made those changes. Here is a demo of the latest changes below. What do you think? Once you approve the final design, I will refactor the tests 👍 

ssh

src/vorta/views/ssh_dialog.py
```diff
@@ -11,6 +11,8 @@
 
 
 class SSHAddWindow(SSHAddBase, SSHAddUI):
+    create_ssh_key_failure = QtCore.pyqtSignal(int)
```
real-yfprojects, line 14:
```suggestion
    failure = QtCore.pyqtSignal(int)
```

src/vorta/views/ssh_dialog.py
```diff
@@ -68,15 +70,16 @@ def generate_key(self):
             self.sshproc.finished.connect(self.generate_key_result)
             self.sshproc.start('ssh-keygen', ['-t', format, '-b', length, '-f', output_path, '-N', ''])
 
-    def generate_key_result(self, exitCode, exitStatus):
-        if exitCode == 0:
+    def generate_key_result(self, exit_code):
```
real-yfprojects, line 73:
```suggestion
    @pyqtSlot(int)
    def generate_key_result(self, exit_code):
```

Not sure whether that is needed. But it might be since the finished signal can be emitted with two values.

bigtedde:
@real-yfprojects @m3nu I included the changes for the signal, and added tests for a ssh key add success and failure. PR is ready for review 👍

m3nu (approved):

m3nu:
Thanks, @bigtedde! Works as promised locally. 👏

Issue #1763, opened 2023-07-27
add a setting to start vorta minimized to tray


sysdbugfactory:
#### The problem

The way to choose the settings to start vorta minimized to tray is confusing. 

#### Requested Solution

A clear wording such as "start vorta minimized as an icon to tray" would be bettre IMHO

#### Additional context

***

## Steps to solve this issue

1. Open https://github.com/borgbase/vorta/tree/master/src/vorta/store/settings.py
2. Edit the label and the tooltip of the setting as needed: https://github.com/borgbase/vorta/blob/b58ffb6aed0254c565788e912a7cbc13db826275/src/vorta/store/settings.py#L51-L52

real-yfprojects:
This is a good idea!

PR #1804, opened 2023-09-03
changed label and tooltip for startup of Vorta

SAMAD101:
### Description

For Issue #1763 
### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ x] I have read the [CONTRIBUTING](https://vorta.borgbase.com/contributing/) guide.
- [x ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

*I provide my contribution under the terms of the [license](./../../LICENSE.txt) of this repository and I affirm the [Developer Certificate of Origin][dco].*

[dco]: https://developercertificate.org/

PR #1820, opened 2023-09-14
Update settings.py

samarthmhetre:
#1763  adding setting to start vorta minimized to tray

### Editing the label and the tooltip of the setting as needed
Previously in settings.py it had 
51: 'label': trans_late('settings', 'Show main window of Vorta on launch'),
52 :           'tooltip': trans_late(
on lines 51 and 52
where in 52 line it the code wasn't fully written 
According to the suggestions I have fixed the code by changing the lines 51 and 52 to the ones mentioned in the issue
So I guess the issue is fixed.
You can merge the code

### DIFF
--- a/src/vorta/store/settings.py
+++ b/src/vorta/store/settings.py
@@ -48,8 +48,8 @@ def get_misc_settings() -> List[Dict[str, str]]:
             'value': True,
             'type': 'checkbox',
             'group': startup,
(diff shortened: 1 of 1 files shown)


real-yfprojects:
Thank you for taking the time to write a PR @samarthmhetre. The issue in question was already fixed by #1804. I am sorry about that. Feel free to tackle another issue.

PR #1846, opened 2023-11-16
Exclude GUI II (without exclude-if-present)

m3nu:
Continues #1742 with original scope only.

All based on @diivi work, who couldn't finish this GSoC subproject.


m3nu:
Removed the "Exclude if present" and only left the new Exclusion dialog.

Screenshot 2023-11-16 at 14 06 12

diivi:
Hey, thanks! I'm out for Diwali holidays so I couldn't get back to you. Happy to see this move forward though. I'll try to help with the exclude if present dialog in the future.

m3nu:
Needed a few edits. If anyone has time for testing, that would be great and speed up the next beta release.

- Only load presets for current OS
- Keep expanding home folder for exclusions
- Build exclusions when running a backup instead of caching them when editing settings only.
- Remove unnecessary DB migrations

src/vorta/assets/exclusion_presets/browsers.json
```diff
@@ -0,0 +1,71 @@
...
+            "*.config/chromium/*/Session Storage",
+            "*.config/chromium/*/Service Worker/CacheStorage",
+            "*.config/chromium/*/Application Cache",
+            "*.config/chromium/*/History Index *",
+            "snap/chromium/common/.cache",
+            "snap/chromium/*/.config/chromium/*/Service Worker/CacheStorage",
+            "snap/chromium/*/.local/share/"
+        ],
+        "tags":["application:chromium", "type:browser", "os:linux"],
+        "author": "Divi"
```
real-yfprojects, line 17: Why did we keep the author field? Authorship is already tracked in git, we don't need to add an unused field that has to be updated when changing a preset.

src/vorta/assets/exclusion_presets/browsers.json
```diff
@@ -0,0 +1,71 @@
+[
+    {
+        "name": "Chromium cache and config files",
+        "slug": "chromium-cache",
+        "patterns":
+        [
+            "*.config/chromium/*/Local Storage",
```
real-yfprojects, line 7: Why did we drop the format syntax specifier? Prepending `fm:` to fnmatch patterns makes sense for those build-in strings because they will continue to work when borg changes the default pattern syntax.

src/vorta/store/models.py
```diff
@@ -105,6 +107,68 @@ def refresh(self):
...
+            excludes += "\n"
+
+        # go through all source=='preset' exclusions, find the name in the allPresets dict, and add the patterns
+        for exclude in ExclusionModel.select().where(
+            ExclusionModel.profile == self,
+            ExclusionModel.enabled,
+            ExclusionModel.source == ExclusionModel.SourceFieldOptions.PRESET.value,
+        ):
+            if exclude.name not in allPresets:
+                continue
```
real-yfprojects, line 146: I suggest logging that case.

src/vorta/views/exclude_dialog.py
```diff
@@ -0,0 +1,330 @@
...
+        if role == Qt.ItemDataRole.EditRole and value == '':
+            self.removeRow(index.row())
+            return True
+        if role == Qt.ItemDataRole.EditRole and ExclusionModel.get_or_none(name=value, profile=self.profile):
+            QMessageBox.critical(
+                self.parent(),
+                'Error',
+                'This exclusion already exists.',
+            )
+            self.removeRow(index.row())
```
real-yfprojects, line 41: I think switching the order of the MessageBox and removing the row would be nicer.

.coverage.nyx2.66373.469600
real-yfprojects, line 1: Why is that file included in the PR?

setup.cfg
real-yfprojects, line 1: I think this change doesn't belong into this PR either.

real-yfprojects (commented):
The issue with dropping the exclude-if-present widget is that its contents still apply although the user is neither aware of it nor can he edit those entries.

The rest looks good and seems to be  working.

src/vorta/assets/exclusion_presets/browsers.json
m3nu: Maybe this change was added after I forked the previous PR. Will double-check and add it back.

m3nu:
> The issue with dropping the exclude-if-present widget is that its contents still apply although the user is neither aware of it nor can he edit those entries.
> 
> The rest looks good and seems to be working.

Thanks for looking into this! I've also disabled processing this setting by Borg. Until we may add it back in some other way.

src/vorta/assets/exclusion_presets/browsers.json
m3nu: I don't mind having the ownership outside Github/git. My hope is that people will add their presets and maybe maintain them.

real-yfprojects:
MacOS tests seem to work again 🎉

src/vorta/assets/exclusion_presets/browsers.json
m3nu: Also not in the [original PR](https://github.com/borgbase/vorta/pull/1742/files#diff-5276b23f1b5068ea8466dc6e08b4db034498e24e00214e62bff2b676988789d7). Let's add this here.

m3nu: I've added the matcher and made some exclusions more specific, so they will match a whole file name and not a partial one. E.g. we want to match `.config/blah` but not `blah.config`.

real-yfprojects (approved):

PR #1871, opened 2023-12-14
Multiselection of folders in sources tab

SAMAD101:
### Description

In the source, multiple folders cant be selected, due to some limitations in `QFileDialog` .

### Motivation and Context

It would be more convenient if one can select multiple directories in one go, not having to open add folder dialog window multiple time, saving previous time and clicks.

### How Has This Been Tested?

By running them in my system. No actual tests edited so far.

### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:


real-yfprojects:
If it doesn't get to complicated go ahead!

SAMAD101:
Committed the changes, multiselection and addition of directories to sources is working now.

SAMAD101:
This needs some changes to the tests, will commit them soon.

SAMAD101:
So apparently, some tests are failing due to the changes I've made to the `source_add`.
I'm not quite getting how this work:
`src/vorta/tests/unit/test_source.py` (line 40-41
```python
mocker.patch('os.access', return_value=False)
 tab.source_add(want_folder=True)
```
We are getting a `TypeError:  'MockFileDialog' object is not iterable'`
So can you explain a bit how these tests work? @real-yfprojects

src/vorta/utils.py
```diff
@@ -171,8 +177,10 @@ def choose_file_dialog(parent, title, want_folder=True):
     dialog.setFileMode(QFileDialog.FileMode.Directory if want_folder else QFileDialog.FileMode.ExistingFiles)
     dialog.setParent(parent, QtCore.Qt.WindowType.Sheet)
     if want_folder:
-        dialog.setOption(QFileDialog.Option.ShowDirsOnly)
-    return dialog
+        list_view = dialog.findChild(QTreeView)
+        list_view.setSelectionMode(QAbstractItemView.SelectionMode.MultiSelection)
+    dialog.exec()
+    return dialog.selectedFiles()
```
real-yfprojects, line 183: Why do you change that? Using the non-blocking `open `with a slot is probably better then using `exec`.

real-yfprojects:
> So can you explain a bit how these tests work?

Have a look at [pytest fixtures](https://docs.pytest.org/en/stable/explanation/fixtures.html). 
The failing tests use the `source_env` fixture which monkeypatches (that is replaces) the `choose_file_dialog` method with a different implementation which is needed for the tests to work without user interaction.
https://github.com/borgbase/vorta/blob/5b36faba8034baf5f8923553c471ba02d894f026/tests/integration/conftest.py#L183-L201

This mocking function still returns a dialog while the code expects a list of selected files resulting in the exception.

src/vorta/utils.py
SAMAD101: > Why do you change that? Using the non-blocking `open `with a slot is probably better then using `exec`.

For some reason that way was not working out, the directories (if selected multiple) were not getting added to the sources.

SAMAD101:
@real-yfprojects now I understand how this tests, when the `tab.source_add` is called inside `test_source_add_remove` and also in other tests, `choose_file_dialog` 's mocking fixture is used instead of the real one. So, if I change the returning value of that fixture to simply a list of directories, which are now used by the `receive` function. 
But for some reason this way is not working, Now its showing 
```
>       paths = choose_file_dialog(self, msg, want_folder=want_folder)
E       TypeError: 'list' object is not callable
```

I'm quite lost of what I could be still be missing atm.

real-yfprojects:
> So, if I change the returning value of that fixture to simply a list of directories

Before you change the fixture returned a function, now its a list. The return value of the fixture is available in tests using the fixture through the `choose_file_dialog` parameter. Before this parameter consequently contained a function that was then called. Now it contains a list. However you still try to call like the function that it contained before, raising an error.

SAMAD101:
> > So, if I change the returning value of that fixture to simply a list of directories
> 
> Before you change the fixture returned a function, now its a list. The return value of the fixture is available in tests using the fixture through the `choose_file_dialog` parameter. Before this parameter consequently contained a function that was then called. Now it contains a list. However you still try to call like the function that it contained before, raising an error.

Thank you for the assistance @real-yfprojects 

Finally done, multiselection working and tests pass now.
Here's a video:

https://github.com/borgbase/vorta/assets/71956678/7b153015-ad80-4288-9d0a-99bfaef92dc2

sayem314:
PR #1938 provides a more detailed implementation that accounts for different types of file dialog views (QListView and QTreeView).

SAMAD101:
> PR #1938 provides a more detailed implementation that accounts for different types of file dialog views (QListView and QTreeView).

Just checked it out, it does account for two different types of view. 
But is that needed for this issue? 
I've used the `NativeDialog` one, which is already in use for the purpose. Just by using a `QTreeView`, we can do this pretty easily in just 3 lines of code.

m3nu:
This works to select multiple files.

For the folder dialog I get this error.

Of course having just one dialog for files and folders including multiple files and folders would be even better.

Screenshot 2024-04-08 at 12 53 15

SAMAD101:
fixed this in this little commit.
However, the intended feature of this PR is not working anymore, perhaps due to some update to PyQT. 
The `QFileDialog` doesn't seem to have any child of `QTreeView` or `QListView`.
I'll try to find a workaround.

Issue #1870, opened 2023-12-13
Making the codebase typesafe.


SAMAD101:
#### The problem

I plan to make the codebase typesafe as discussed in this [discussion](https://github.com/borgbase/vorta/discussions/1868). This will require a lot of refactoring and changes. I'll be using this issue thread to give updates, and for some help if it needs.

m3nu:
I recommend to do this step-by-step with smaller PRs. And start very small before getting larger over time. That's to get timely feedback and also to avoid conflicts with other open PRs.

SAMAD101:
> I recommend to do this step-by-step with smaller PRs. And start very small before getting larger over time. That's to get timely feedback and also to avoid conflicts with other open PRs.

Sure sounds good!

PR #1872, opened 2023-12-21
Refactored some parts of `src/vorta/store`

SAMAD101:
### Description

To improve readability a bit, and towards making the codebase typesafe.

### Related Issue

#1870 

### Types of changes

### Checklist:


requirements.d/dev.txt
```diff
@@ -3,6 +3,7 @@ coverage
 flake8
 macholib
 nox
+peewee
```
real-yfprojects, line 6: Since peewee is a production dep it shouldn't be needed to specify it here as well.

src/vorta/store/connection.py
```diff
@@ -62,18 +62,18 @@ def init_db(con=None):
...
     )
-    last_scheduled_backups_per_profile = (
+    last_scheduled_backups_per_profile = Tuple(
         EventLogModel.select(EventLogModel.profile, fn.MAX(EventLogModel.start_time))
         .where(EventLogModel.subcommand == 'create', EventLogModel.category == 'scheduled')
         .group_by(EventLogModel.profile)
     )
 
-    three_months_ago = datetime.now() - timedelta(days=6 * 30)
+    three_months_ago: datetime = datetime.now() - timedelta(days=6 * 30)
```
real-yfprojects, line 76: For local variables static type linters are usually able to infer the type themselves.

src/vorta/store/connection.py
```diff
@@ -105,11 +105,11 @@ def init_db(con=None):
...
 
 
-def backup_current_db(schema_version):
+def backup_current_db(schema_version) -> None:
     """
     Creates a backup copy of settings.db
     """
 
-    timestamp = datetime.now().strftime('%Y-%m-%d-%H%M%S')
+    timestamp: str = datetime.now().strftime('%Y-%m-%d-%H%M%S')
```
real-yfprojects, line 113: Same here

src/vorta/store/models.py
```diff
@@ -26,11 +27,11 @@ class JSONField(pw.TextField):
     From: https://gist.github.com/rosscdh/f4f26758b0228f475b132c688f15af2b
     """
 
-    def db_value(self, value):
+    def db_value(self, value) -> (None or str):
```
real-yfprojects, line 30: Is that a valid type hint?

Hofer-Julian: I don’t think so. It should be `None | str`

real-yfprojects: And for that one needs to add `from __future__ import annotions` afaik.

Hofer-Julian: Only for Python < 3.10

real-yfprojects: ...which we support

src/vorta/store/connection.py
SAMAD101: Yup. Those changes might be a bit redundant, but I did those just to improve readability bit.

src/vorta/store/models.py
SAMAD101: Hmm.. I think `Union[str, None]` would be better

real-yfprojects: Even better: `Optional[str]`

src/vorta/store/connection.py
SAMAD101: So, should I remove these type annotations or, remove them? I thought that adding them to the whole codebase might make it messy so, I went for only local variables of some methods and functions, and leaving variables that variables that hold some `Field` (as inside modals).

src/vorta/store/connection.py
```diff
@@ -62,12 +63,12 @@ def init_db(con=None):
     # Delete old log entries after 6 months.
     # The last `create` command of each profile must not be deleted
     # since the scheduler uses it to determine the last backup time.
-    last_backups_per_profile = (
+    last_backups_per_profile = Tuple(
```
real-yfprojects, line 66: I don't think this is proper use of type hints. Shouldn't it be the following instead?
```suggestion
    last_backups_per_profile : Tuple = 
```

src/vorta/store/connection.py
```diff
@@ -105,7 +106,7 @@ def init_db(con=None):
         s.save()
 
 
-def backup_current_db(schema_version):
+def backup_current_db(schema_version: Any) -> None:
```
real-yfprojects, line 109: This should be `int` afaik.

src/vorta/store/migrations.py
```diff
@@ -15,7 +17,7 @@
 )
 
 
-def run_migrations(current_schema, db_connection):
+def run_migrations(current_schema: Any, db_connection: SqliteDatabase):
```
real-yfprojects, line 20: Do we have to specify `Any`? Isn't it the default, implicit type hint?

real-yfprojects (commented):
Good progress. Typing is hard but valuable!

src/vorta/store/connection.py
SAMAD101: True

src/vorta/store/migrations.py
SAMAD101: I think I left that one undone. It should be `SchemVersion` from `vorta.store.models`

PR #1879, opened 2024-01-08
random cleanups / fixes

ThomasWaldmann:
most stuff was found by pycharm code checker.

commit 48af51f should fix some potential malfunction, that function is called without params rather often, thus site=None -> bug.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_form.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_form.yaml
@@ -1,5 +1,5 @@
 name: "Bug Report Form"
-description: "Report a bug or a similiar issue."
+description: "Report a bug or a similar issue."
 body:
   - type: markdown
     attributes:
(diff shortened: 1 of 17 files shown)


Hofer-Julian (approved):
Looks good to me.
The pycharm code checker seems to be quite useful.

ThomasWaldmann:
@Hofer-Julian It's a bit of a mixed experience:
- sometimes it finds something useful (like here)
- sometimes it has false positives (valid code, but the checker didn't get it)
- lots of useless stuff (like unknown words)
- the sheer volume of warnings can be rather high (vorta: > 50000)

Luckily, the result is somehow structured, so one does not need to look at all of the warnings.

Hofer-Julian:
> @Hofer-Julian It's a bit of a mixed experience:
> 
>     * sometimes it finds something useful (like here)
> 
>     * sometimes it has false positives (valid code, but the checker didn't get it)
> 
>     * lots of useless stuff (like unknown words)
> 
>     * the sheer volume of warnings can be rather high (vorta: > 50000)
> 
> 
> Luckily, the result is somehow structured, so one does not need to look at all of the warnings.

Interesting.

I guess, we prefer tools anyway that have a CLI and can be easily run on the CI like `typos` and `ruff`.

m3nu (approved):

PR #1885, opened 2024-01-10
Update appdata.xml

Hofer-Julian:
The appdata.xml doesn't pass validation of flathub

1. The `launchable` tag is nowadays required
2. Flatpak doesn't like the beta releases. In the end, it only made sense to remove them from the xml

I have a PR open that patches the xml: https://github.com/flathub/com.borgbase.Vorta/pull/150#issuecomment-1885331084
It would still be good to get a patch release out, if possible :)


Hofer-Julian:
Any objections to this @m3nu? :)

m3nu:
All good. Just didn't see it.

m3nu (approved):

PR #1922, opened 2024-02-05
Add developer name to appdata

Hofer-Julian:
Flathub is getting more and more strict when it comes to metadata. 
I've added "Vorta developers", I can also be more specific if people prefer that.


real-yfprojects:
`Vorta Contributors` would be better but it doesn't really matter.

Hofer-Julian:
> `Vorta Contributors` would be better but it doesn't really matter.

Fine by me

sayem314 (approved):

Hofer-Julian:
@real-yfprojects or @m3nu can one of you merge this?
Would be good not to forget this before the next release.

m3nu (approved):

PR #1934, opened 2024-02-12
Prevent exception while running BorgVersionJob with unexpected lines on stderr

ntninja:
### Description
The borg job logic expects the `profile_name` key to be unconditionally present in the params if any data is received on stderr. This causes the BorgVersionJob (which didn’t set this param) to fail if any unexpected output – such as LD_PRELOAD warnings – in generated while running `borg --version`. This in turn causes Vorta to silently (other than an unrelated looking exception in the console output) fall back to assuming borg 1.1.0 as the borg version.

### Related Issue
Please don’t make me create an issue for this…

### Motivation and Context
Borg version detection currently silently fails in environments that have LD_PRELOAD set in a way to causes warnings to be emitted.

### How Has This Been Tested?
The change is pretty trivial, so I just tested with and without.

### Types of changes

### Checklist:

- [ ] I have read the [CONTRIBUTING](https://vorta.borgbase.com/contributing/) guide.
- [X] My code follows the code style of this project.

### DIFF
--- a/src/vorta/borg/version.py
+++ b/src/vorta/borg/version.py
@@ -22,6 +22,7 @@ class BorgVersionJob(BorgJob):
             return ret
 
         ret['cmd'] = ['borg', '--version']
+        ret['profile_name'] = 'default'
         ret['ok'] = True
         return ret
 


src/vorta/borg/version.py
```diff
@@ -22,6 +22,7 @@ def prepare(cls):
             return ret
 
         ret['cmd'] = ['borg', '--version']
+        ret['profile_name'] = 'default'
```
real-yfprojects, line 25: Did you try setting it to `None`? `'default'` might be the name of an existing profile but such a profile doesn't have to exist possibly resulting in unexpected sideeffects.

real-yfprojects (commented):
Thank you for taking the time to create this PR. You can find my commentary below.

m3nu:
Was also looking at this. Currently wondering about the test timeouts on Linux. A Github glitch?

src/vorta/borg/version.py
ntninja: Setting it to `None` also seems to work, yes. I think I chose `'default'` because the constructor of `BorgJob` mentions it, but I change it if `None` is better.

real-yfprojects: I see. Then let's stick to 'default'. However this seems like an unreasonable design choice to me.

real-yfprojects (approved):

PR #1976, opened 2024-03-24
terminate process

hasrat17:
Recalculating sizes
### Description

Missing UI element may trigger crash when recalculating folder sizes


real-yfprojects:
Which issue are you trying to fix? Is it #930?

Your PR description definetely doesn't match the PRs contents.

PR #1986, opened 2024-04-07
Keep the profile list sorted when creating/renaming profiles

### LINKED ISSUE #1983: Profile list in the main window are not kept sorted when adding/renaming entries
#### Description

Creating a new profile adds the new profile at the bottom of the list in the main window. Similarly, editing an existing profile doesn't change its position in the list. I think that in both cases the list should be kept sorted. Restarting Vorta correctly updates the positions of the profiles to be sorted alphabetically.

This was recently noticed during the review of https://github.com/borgbase/vorta/pull/1899#issuecomment-2029914025

1. Create a new profile using the menu in the main window
2. The profile is added last on the list, even if it should sort before other entries

Alternatively:
1. Rename an already existing profile
2. Its position in the list doesn't change, even if the new alphabetical ordering is no longer valid

#### Environment

- OS: Linux
- Vorta version: latest commit (9b8dbcecfbab7e7492e05f5e0ae22a2631335718)
- Installed from: pip
- Borg version: N/A

I see two possible solutions for this:

1. Add some additional handling in the `profile_add_edit_result` function to keep the list sorted. This is a quick commit that does that: https://github.com/Parnassius/vorta/commit/59661f937b25670e0bb40362a696693aa01bbf90

2. Use the [sortItems method of QListWidget](https://doc.qt.io/qt-6/qlistwidget.html#sortItems) to sort the list after every update, thus removing the need to use the `order_by` sql method.

The two solutions cannot co-exist since the sql ordering is case-insensitive while QListWidget.sortItems is case-sensitive. In the second case additional care should be taken to ensure the use of `sortItems` doesn't break anything. From a quick glance I think the way `current_profile` is set during the initialization of the main window should be updated, and maybe some additional changes are needed as well. https://github.com/borgbase/vorta/blob/9b8dbcecfbab7e7492e05f5e0ae22a2631335718/src/vorta/views/main_window.py#L66-L68

What do you think? Let me know if I should create a small pull request with the commit linked above or if a solution based on option 2 should be preferred

Parnassius:
### Description

Creating a new profile adds the new profile at the bottom of the list in the main window. Similarly, editing an existing profile doesn't change its position in the list. This change keeps the list sorted alphabetically in both cases.

### Related Issue

Fixes #1983

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/tray_menu.py
+++ b/src/vorta/tray_menu.py
@@ -59,10 +59,10 @@ class TrayMenu(QSystemTrayIcon):
             cancel_action.triggered.connect(self.app.backup_cancelled_event.emit)
         else:
             status.setText(self.tr('Next Task: %s') % next_task_time)
-            profiles = BackupProfileModel.select().order_by(BackupProfileModel.name)
+            profiles = BackupProfileModel.select()
             if profiles.count() > 1:
                 profile_menu = menu.addMenu(self.tr('Backup Now'))
-                for profile in profiles:
+                for profile in sorted(profiles, key=lambda p: (p.name.casefold(), p.name)):
                     new_item = profile_menu.addAction(profile.name)
                     new_item.triggered.connect(lambda state, i=profile.id: self.app.create_backup_action(i))
             else:
--- a/src/vorta/views/main_window.py
+++ b/src/vorta/views/main_window.py
@@ -284,13 +284,18 @@ class MainWindow(MainWindowBase, MainWindowUI):
     def profile_add_edit_result(self, profile_name, profile_id):
         # Profile is renamed
         if self.profileSelector.currentItem().data(Qt.ItemDataRole.UserRole) == profile_id:
-            self.profileSelector.currentItem().setText(profile_name)
+            profile = self.profileSelector.takeItem(self.profileSelector.currentRow())
+            profile.setText(profile_name)
         # Profile is added
         else:
             profile = QListWidgetItem(profile_name)
             profile.setData(Qt.ItemDataRole.UserRole, profile_id)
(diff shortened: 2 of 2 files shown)


shivansh02:
Screenshot from 2024-04-12 12-28-37
Works fine for me, assuming we want to keep the sorting case sensitive.

m3nu:
Thanks for testing, @shivansh02! I don't have a strong preference for the case-sensitivity, but would lean towards **case-insensitive** sorting.

Parnassius:
The latest commit adds case-insensitive sorting to the list in the main window, as well as the list in the `Backup Now` tray menu.

The solution I used was to ignore the `ORDER BY` sql directive and simply sort the profiles with python, using both the casefolded and the normal name as the sorting key. Adding the normal name as well is required to keep the sorting stable, otherwise two profiles called `test` and `TEST` would have no fixed ordering.

In my comment on #1983 I talked about the `sortItems` method of `QListWidget`, which would be perfect for the list in the main window. However the tray menu doesn't support anything similar, so I decided against it to keep the logic the same in both places.

One final option I thought about was the [`COLLATE NOCASE`](https://sqlite.org/datatype3.html#collation) sqlite function, but that only converts the 26 ascii uppercase letters to the lowercase equivalents, so the sorting would have remained case-sensitive for non-ascii characters.

shivansh02:
The profile list is now sorted case-insensitively, couldn't test the Backup Now list because my distro doesn't have a tray menu.

m3nu (approved):

m3nu:
Thanks for the contribution, @Parnassius and thanks to @shivansh02 for reviewing.

PR #2078, opened 2024-09-16
l10n - make string translatable (syntax fix)

p-bo:
String is not appearing on Transifex, so tried to modify its form based on ones, which does.


m3nu (approved):

m3nu:
Extra trailing comma, which makes it a tuple. Common error in Python that's hard to catch.

PR #2102, opened 2024-10-18
borg2: support >= b10, drop support for < b10

ThomasWaldmann:
- new repo-* commands (previously: r*)
- some stats are gone

I did some manual testing with a local fs repository.


m3nu:
Good stuff. That was a small change. Eventually we should just put version 2 and not support any older beta.

ThomasWaldmann:
Just to be clear: this only supports recent betas, I don't think it makes sense to keep support for older betas, that just makes things more complicated.

BTW, vorta could do things differently with borg2 (e.g. use "archive series"), but guess here it is the same problem: while vorta supports borg 1.x also, it would make things more difficult, so guess that needs to wait for quite a while still.

m3nu (approved):

PR #2103, opened 2024-10-20
Move populate_from_profile() to end of init

### LINKED ISSUE #2101: validation and compaction interval not changeable
### Description

See picture:
- the compaction interval is disabled, so one can not change the value of it. i did not touch anything there, so this is a bug: it should be changeable.
- the validation interval is enabled, so one can change the value of it. this is because i toggled the checkmark to off and to on again to get it into that state. this is a workaround for this bug.

Screenshot 2024-10-18 at 19 45 12

### Reproduction

### OS

macOS 15

### Version of Vorta

0.10.0 beta1

Binary

### Version of Borg

1.4.0

### Logs

_No response_

m3nu:
Fixes #2101

### DIFF
--- a/src/vorta/views/schedule_page.py
+++ b/src/vorta/views/schedule_page.py
@@ -23,8 +23,6 @@ class SchedulePage(SchedulePageBase, SchedulePageUI, BackupProfileMixin):
             'fixed': self.scheduleFixedRadio,
         }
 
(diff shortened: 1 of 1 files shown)


ThomasWaldmann (approved):
I practically tested this and it fixes #2101 for me.

Issue #2100, opened 2024-10-18
bundled borg is preferred over system borg (macOS)


ThomasWaldmann:
### Description

After stumbling over #2099 I also noticed that vorta seems to prefer the `borg` that comes bundled with it, even if the user has installed another borg on their system.

In my case, I have:
```
tw@MacBook-Pro ~ % which borg       
/opt/homebrew/bin/borg
tw@MacBook-Pro ~ % borg -V
borg 1.4.0
```

And that borg is the Apple Silicon binary, which would be better than the currently bundled Intel binary inside vorta.

### Reproduction

- [X] I tried to reproduce the issue.
- [X] I was able to reproduce the issue.

### OS

macOS 15

### Version of Vorta

0.10.0 beta1

### What did you install Vorta with?

Binary

### Version of Borg

1.4.0

ThomasWaldmann:
I just restarted vorta, still the same: it wants to use the bundled borg.

ThomasWaldmann:
Just checked out vorta master branch and installed it to a venv:

That one finds the homebrew borg:

```
2024-10-18 21:48:17,192 - vorta.borg.borg_job - INFO - Running command /opt/homebrew/bin/borg --version
```

ThomasWaldmann:
```
    @classmethod
    def prepare_bin(cls):
        """Find packaged borg binary. Prefer globally installed."""

        borg_in_path = shutil.which('borg')

        if borg_in_path:
            return borg_in_path
        elif sys.platform == 'darwin':
            # macOS: Look in pyinstaller bundle
            from Foundation import NSBundle

            mainBundle = NSBundle.mainBundle()

            bundled_borg = os.path.join(mainBundle.bundlePath(), 'Contents', 'Resources', 'borg-dir', 'borg.exe')
            if os.path.isfile(bundled_borg):
                return bundled_borg
        return None
```
Guess we need some additional logging in the else-branch for `os.environ["PATH"]` because the `shutil.which` obviously could not find my homebrew borg (and it is found if I use `which` in a terminal)

m3nu:
`vorta.spec` has the path that's being searched. This is an Apple app setting. Currently it's

`/usr/local/bin:/usr/local/sbin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/homebrew/bin`

ThomasWaldmann:
That might be, but something is broken there.

My borg is in `/opt/homebrew/bin` and does not get found by `shutil.which`.

[...]

PR #2166, opened 2024-12-10
Override $PATH for MacOS to find native homebrew borg binaries (#2100)

greigdp:
### Description
Override the $PATH environment variable on the MacOS (Darwin) platform, so that homebrew-provided native binaries for Borg will be used, rather than falling back to the built-in Borg binary.

### Related Issue
https://github.com/borgbase/vorta/issues/2100

### Motivation and Context
For an unknown reason (see linked issue above), the PyInstaller approach to setting the PATH variable does not appear to function any more - Vorta on ARM with a homebrew-installed native borg binary will not find that binary. Debugging shows the PATH variable is set wrongly, and will never find a binary in /opt/homebrew/bin. This patch will change that.

### How Has This Been Tested?
1. Debugging statements added around existing areas of code (see linked issue) to establish the PATH being used
2. Tested with this change and observed the homebrew based binary was found and shown in About.
3. Not tested end-to-end yet, as one of the apparent issues in testing was reproducibility against a production release - while I have built it according to the Github actions, this still needs tested against a development build that is created using the normal release method.

N/A

### Types of changes

### Checklist:

- [ ] I have read the [CONTRIBUTING](https://vorta.borgbase.com/contributing/) guide.
- [ ] My code follows the code style of this project.

### DIFF
--- a/src/vorta/borg/borg_job.py
+++ b/src/vorta/borg/borg_job.py
@@ -203,7 +203,12 @@ class BorgJob(JobInterface, BackupProfileMixin):
     @classmethod
     def prepare_bin(cls):
         """Find packaged borg binary. Prefer globally installed."""
-
+        # On MacOS, the PATH environment variable does not seem to be set when run as a pyinstaller binary.
+        # More info at https://github.com/borgbase/vorta/issues/2100
+        # Set the path to also find homebrew installs of Borg, and avoid falling back to the embedded binary.
+        if sys.platform == 'darwin':
+            os.environ["PATH"]="/usr/local/bin:/usr/local/sbin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/homebrew/bin"
+        # Now continue looking for the borg binary to use
         borg_in_path = shutil.which('borg')
 
         if borg_in_path:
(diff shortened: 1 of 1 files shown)


src/vorta/borg/borg_job.py
```diff
@@ -203,7 +203,13 @@ def prepare(cls, profile):
     @classmethod
     def prepare_bin(cls):
         """Find packaged borg binary. Prefer globally installed."""
-
+        # On MacOS, the PATH environment variable does not seem to be set when run as a pyinstaller binary.
+        # More info at https://github.com/borgbase/vorta/issues/2100
+        # Set the path to also find homebrew installs of Borg, and avoid falling back to the embedded binary.
+        if sys.platform == 'darwin':
+            current_path = os.environ.get("PATH", "")
+            os.environ["PATH"] = f"{current_path}:/opt/homebrew/bin:/usr/local/bin"
```
ThomasWaldmann, line 211: I think this would malfunction if there is not PATH set, resulting in a PATH with a leading column. An empty path element might get interpreted as CWD (iirc) and that is dangerous as it might execute binaries from any path you happen to be in.

m3nu: Added `/usr/bin:/bin` as minimal fallback. The old spec file also had `sbin`, which I think wouldn't be needed since Borg wouldn't be installed as admin tool on macOS.

m3nu (approved):

Issue #2189, opened 2025-01-13
logs: providing info on borg arguments and an approximate shell command


goebbe:
This issue is _not_ about code execution - it is only about the information in the  **Vorta logs**. 

In Vorta 0.10.3 an _approximate_ borg command is provided via the logs, e.g. whenever a backup is created.

There are two issues related to this: 
1. Sometimes the shell command in the logs is wrong - and makes debugging of the underlying issue difficult. e.g. see #2164
2. The printed shell command is sometimes not in line with the formatting recommendations of borg - e.g. quotation marks are stripped away internally. 

I would like to keep a working shell command in the logs. This proved to be useful in the past for copy/modify/past.

**Proposal**: 
1. new: Print the actual list of command line arguments (`cmd`) to the logs. (in addition to the shell command) 
2. change: Print an approximate shell command (build from cmd) that is in line with the borg formatting recommendations.  

TODO: Provide some examples to illustrate the issues.

I would like to get some feedback, on how to proceed. 
- Would it be o.k. to print the list with the command line arguments to the logs?
- I have some code pending, which would lead to better formatting - should I provide a pull request?

m3nu:
Sounds good to me to log this command closer to where it's run.

goebbe:
The associated pull request has been approved, but not merged yet.

VandalByte:
This issue has been resolved, thanks to a PR from @goebbe.

Issue #2164, opened 2024-12-06
Extra borg arguments with white space raise an error


goebbe:
A Vorta user had a hard time when pattern matching a folder-name with white space: https://github.com/borgbase/vorta/discussions/2156#discussioncomment-11470549
Since matching names that contain a single space is a common use case, I decided to open this issue. 

I am able to reproduce the issues, using Vorta 0.10.1 with borg 1.4 on LinuxMint22 (based on Ubuntu 24.4) 

The issue appears e.g. when trying to pattern match an item name (Folder/ File) that contains _white space_,
via Vortas "Extra borg arguments for borg create"  found at Vorta>Schedule>Shell Commands>Extra arguments for borg create. 

To reproduce, I created a folder named `test folder`, with a single space in the middle of the folder-name.
I used the following `--pattern` option matcher and tried to exclude the folder from my backup by passing: 
`--pattern=-home/User/test\ folder`

This led to an **error** message: `borg create:error: argument ARCHIVE: "folder": No archive specified`. 

This is the _borg command_ from the Vorta log files: 
`/usr/bin/borg create --remote-path=/usr/local/bin/borg --list --progress --info --log-json --json --filter=AM -C lz4 --pattern=-home/User/test\ folder ssh://backupUser@192.168.0.5:22/~/backup/vorta-repo-EB845::User-EliteBook-845-2024-12-06-110323 /home/User`

Since this _borg command_ in the log-files seems to be correct, I presumed, that his must be a borg issue and opened an issue on borg:  https://github.com/borgbackup/borg/issues/8578

However, it appears that the borg command (from the Vorta logs) actually work as expected, when copied and executed directly on the command line. 

## To summarize the issue:
The following syntax **works** for matching `test folder` with a white space, **when using borg directly** on the command line:
```
$ borg create  --pattern=-subdir/test\ folder  repo::space subdir
$ borg create  '--pattern=-subdir/test folder'  repo::space subdir
$ borg create  --pattern='-subdir/test folder'  repo::space subdir
```
However, when passing these valid options **using Vorta** (Extra borg arguments) an **error** is raised: 
```
--pattern=-subdir/test\ folder
'--pattern=-subdir/test folder'
--pattern='-subdir/test folder'
```

The error messages differ slightly, depending on which of the three options are used with Vorta. 
But while each of these option works on the command line, none of these options could be used successfully when passed/ applied via Vorta. 

## Extract from the Vorta log files when escaping the white space by `\`: 
```
[...]

PR #2190, opened 2025-01-13
logs: Add borg command arguments and a shell command approximation

goebbe:
### Description
This pull request has two purposes:
1. new: logs: print the list of borg command arguments, that are used in Vorta
2. modify: logs: print a shell approximation for the borg command that respects the borg recommendations.

The pull request only concerns what is printed to the logs.

### Related Issue
https://github.com/borgbase/vorta/issues/2189

Currently, when executing borg command, an "approximate" of the (shell) borg command is printed to the logs.
This command is an _approximation_ in the sense that it is artificially constructed from a list of command line arguments `cmd`, that is used internally by Vorta.  

Note that the list of command line arguments `cmd` does neither require nor contain quotation marks and can be passed "as is" to Popen()

For shell usage, however, borg recommends putting the following into quotation marks: 
- item names that contain whitespace " ",  since escaping with "\" is shell specific.
- arguments with characters that are used for pattern matching "*" "?"
- regular expressions

This pull request does two thing: 
1. Print the unmodified list of borg arguments `cmd` to the logs
2. Changes to the shell command approximation:
    - change the description, and make clear that this is a shell command approximation
    - add quotation marks, where recommendet 
    -  the aim is to get a shell command that could be copy/pasted to any shell and be directly executed.

### Motivation and Context
- currently there can be differences between the borg command in the logs and the actual list of borg arguments, this makes debugging cumbersome. See #2164
- the borg command (approximation) in the logs currently does not fit to the borg recommendations, i.e. use quotation marks for white space or regular expressions.

### How Has This Been Tested?
The following borg arguments have been added to the "extra borg arguments" via the Vorta UI - to check if the resulting shell command approximation provides the intended format / quotation marks:

```
--remote-path=/usr/local/bin/borg                   #no changes required
--pattern='+home/user/test folder/testfile'     #white space in folder name
--pattern='-re:home/user/test\sfolder2'          #regular expression
--pattern='-re:crazy?=cra'                         #regualr expression that contains a "="
-e 're:crazy?=cra'                                #regular expression but without pattern=
-e '/home/user/*.tmp'                            #wildcard should be quoted
-e /home/user/test/cache                          #no changes required
--patterns-from /home/user/patterns.lst        #no changes required
```

These are the **resulting etries in the logs**: 

The list of _borg command line arguments_ `cmd`: 
`2025-01-13 12:35:57,302 - vorta.borg.borg_job - INFO - borg command arguments: ['/usr/bin/borg', 'create', '--remote-path=/usr/local/bin/borg', '--list', '--progress', '--info', '--log-json', '--json', '--filter=AM', '-C', 'lz4', '--pattern=+home/user/test folder/testfile', '--pattern=-home/user/test folder', '--pattern=-re:home/user/test\\sfolder2', '--pattern=-re:crazy?=cra', '-e', 're:crazy?=cra', '-e', '/home/user/*.tmp', '-e', '/home/user/test/cache', '--patterns-from', '/home/user/patterns.lst', 'ssh://user@192.168.0.5:22/~/backup/vorta-repo::user-2025-01-13-123557', '/home/user']`

The _shell command approximation_: 
`2025-01-13 12:35:57,302 - vorta.borg.borg_job - INFO - shell command approximation: /usr/bin/borg create --remote-path=/usr/local/bin/borg --list --progress --info --log-json --json --filter=AM -C lz4 --pattern='+home/user/test folder/testfile' --pattern='-home/user/test folder' --pattern='-re:home/user/test\sfolder2' --pattern='-re:crazy?=cra' -e 're:crazy?=cra' -e '/home/user/*.tmp' -e /home/user/test/cache --patterns-from /home/user/patterns.lst ssh://user@192.168.0.5:22/~/backup/vorta-repo::user-2025-01-13-123557 /home/user`

Ths shell command approximation can be copied to a shell "as is".

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/borg/borg_job.py
+++ b/src/vorta/borg/borg_job.py
@@ -234,9 +234,22 @@ class BorgJob(JobInterface, BackupProfileMixin):
                 profile=self.params.get('profile_id', None),
             )
             log_entry.save()
-            cmd_log_tmp = [s.replace(" ", "\\ ") for s in self.cmd]  # escape whitespace - for logs
-            logger.info('Running command %s', ' '.join(cmd_log_tmp))
-            del cmd_log_tmp
+            logger.info('borg command arguments: %s', self.cmd)
+
+            # logs: Provide a shell command approximation.
+            cmd_tmp = self.cmd[:]
+            special_strings = [' ', '*', '?', 're:']
+            for i, arg in enumerate(cmd_tmp):
+                if any(specialstr in arg for specialstr in special_strings):
+                    if ('re:' in arg) and ('=' in arg) and (arg.find('=') < arg.find('re:')):
+                        cmd_tmp[i] = arg.replace("=", "='", 1)  # add quote after first "="
+                    elif ('re:' not in arg) and ('=' in arg):
+                        cmd_tmp[i] = arg.replace("=", "='", 1)
+                    else:
+                        cmd_tmp[i] = arg.replace(arg, "'" + arg)  # add quote at start
+                    cmd_tmp[i] += "'"  # add quote at the end
+            logger.info('shell command approximation: %s', ' '.join(cmd_tmp))
+            del cmd_tmp
 
         p = Popen(
             self.cmd,


goebbe:
@m3nu This is ready for review.

m3nu:
Thanks for the PR, as always. So this aims to log a command for copy & pasting and reversing some of the escaping done when logging this? I wonder if there is a simpler way to achieve something similar? E.g. here some different ways to log stuff.  

```
# This WILL escape the newline because %r uses repr(msg).
logging.debug("Repr style: %r", msg)

# This will NOT escape the newline; you'll actually see the newline rendered.
logging.debug("String style: %s", msg)
```

The issue could also come from our `join()` function.

For this PR I worry that:

- It's pretty complex and runs for each Borg command, even though it's rarely needed.
- If there is any issue in this function, the Borg command will fail.
- No tests to ensure the examples work.
- There may be more cases that need be added now or in the future. I think building a function that reverses some other escaping function is pretty complex.
- I wouldn't be able to debug it in the future and would likely just remove it, if there is any issue.

goebbe:
@m3nu: Thanks for the feedback. 

Printing the list of borg command arguments is a simple one-liner:
`            logger.info('borg command arguments: %s', self.cmd)`
This prints the exact list of borg arguments as passed to Popen() - this is everything that is required in the logs for debugging. 

So a simple (and correct) solution would be to just put this single line instead of (re-)constructing a valid shell command. 
Disadvantage: This list of "borg command arguments" _cannot_ be copied/ pasted to a _shell_ (not sure if this is even a goal) 

The list of borg command arguments does _not_ contain any escape characters or quotation marks. These are not required if no shell is involved. 
So whatever we do when building a valid _shell_ command for the logs - we have to manually take care about how we treat whitespace and special characters: 

- Thomas Waldman pointed out that whitespace should not be escaped by "\\" since this is not supported by all shells. 
- Moreover, there may be wildcards and regular expression which also should be put into quotation marks -  for shell commands (this is written in the borg docs).  


**Computational overhead:** I wondered about this, too. :-) To get a rough estimate, I run the code to format the pretty long and complicated list of borg arguments from my commit message, on my mid-range laptop inside a Jupyter notebook.  
It took 0.13 milliseconds - not sure if this is acceptable, or not. 

**Complexity**  I agree that the code in the pull request is not self-explanatory, since I tried to put the first quotation mark after the first "=" if there are any. 
Thinking about it - one way to reduce the complexity is to just always put the whole argument inside quotation marks, whenever it contains one of the special strings - which would result in a pretty self-explanatory four-liner, since I would get rid of the "if elseif else craziness":

```
special_strings = [' ','*', '?', 're:']
for i, arg in enumerate(cmd):
     if any(specialstr in arg for specialstr in special_strings):
        cmd[i]=arg.replace(arg,"\'"+arg+"\'")    #add quotation marks 
```

The only difference for the shell command approximation would be that now I get
`'--pattern=-home/user/test folder'`
instead of: 
`--pattern='-home/user/test folder'`
Which both run just fine in the shell, despite the whitespace.  :-) 

Currently, in 0.10.3, we write a borg command (approximation) to the logs that escape whitespace using "\\" and without any quotation marks (not even for regular expressions "re:") - I would like to fix this. 

I believe there are several valid options, here: 
1. Just print the plain _list_ of borg arguments
2. Simplify the code and just put the shell command (approximation)
3. Put both
4. ?

Any preferences? 

I would go with option 3, followed by option 2

goebbe:
Thinking about this again, I believe option 2 would be a good one: 
- The only change would be that in the logs, the borg arguments are put into quotation marks if they contain 
special strings (i.e. whitespace, wildcards or regular expressions)
- as a result the arguments are preserved (not changed), which is good for debugging
- the resulting borg command, in the logs, could be used as a shell command

goebbe:
@m3nu  I simpified the code and updated my pull request: 

**Changes**: 
- simpified logic: just put the whole argument within quotation marks if it contains whitspace, wildcards or regular expressions
    - good for **debugging**, since argument splitting and argments themselfe are not touched 
    - no complicated if/elseif/else construct  :-) 
    - much shorter
    - code is pretty much self explanatory (imo) - it would be easy to e.g. add a new "quotation_string".
- the resulting borg command from the logs can be copied/ pasted to a shell as is
- computational time is down by half, to 0,0731 milliseconds for complicated casees

This solution is more robust than what is in 0.10.3: The use of escape characters "\\"  for whitspace (as used curently) is shell dependent.

**Example log output** for a complicated case: 
`/usr/bin/borg create --remote-path=/usr/local/bin/borg --list --progress --info --log-json --json --filter=AM -C lz4 '--pattern=+home/user/test folder/testfile' '--pattern=-home/user/test folder' '--pattern=-home/user/test folder' '--pattern=-re:home/user/test\sfolder2' '--pattern=-re:crazy?=cra' -e 're:crazy?=cra' -e '/home/user/*.tmp' -e /home/user/test/cache --patterns-from /home/user/patterns.lst ssh://user@192.168.0.5:22/~/backup/vorta-repo::user-845-2025-01-15-083513 /home/user`

This command can be copied pasted from the logs to the shell - and runs just fine (only archive name has to be updated, obviously)

As before: The execution path is not concerned here - this is purely about a pretty and useful (for debuging and copy)  borg command in the logs.

m3nu:
The new version is much clearer! No objections to add this.

m3nu (approved):

goebbe:
The corresponding issue has now automatically been "marked as stale":
https://github.com/borgbase/vorta/issues/2189#issuecomment-2803572198

This pull request has been marked as approved on Jan 16 - does that mean it will be merged automatically, later? 
Or is there something missing, from my side?

m3nu:
Can you review this @VandalByte? I don't want any new errors just because we're formatting logs differently. After that I'm happy to merge it. Thanks!

VandalByte:
When it comes to errors, I ran tests with multiple flags and various borg commands, and I didn’t encounter any issues.

Some things I want to point out, as goebbe mentioned:
> The only difference for the shell command approximation would be that with the simplified code I would get
'--pattern=-home/user/test folder'
instead of:
--pattern='-home/user/test folder'
Which both run just fine in the shell, despite the whitespace. :-)

Flags that have a similar pattern like `--pattern=`, `--remote-path=` are getting fully quoted, if the code encounters any [char](https://github.com/borgbase/vorta/pull/2190/files#diff-5aac1bfe6bba70f21fbb7569b769d9a92ff1ecc74e4cf1322bd4894f3f068c1fR239) from `quote_strings`.  

**Actual** -> `'--pattern=- /home/vandal/bb/file3.txt'`  
**Expected** -> `--pattern='- /home/vandal/bb/file3.txt'`

I saw that code to handle that was removed from previous commit. Even though `argparse` might throw an error on something like this, borg seems to handle it fine internally.

Example, I tested it out by running:
```
/usr/bin/borg create --list --progress --info --log-json --json --filter=AM -C lz4 '--pattern=- /home/vandal/bb/file3.txt'   /home/vandal/Music/secret::litc1   /home/vandal/bb
```
And this is the expected command:
```
/usr/bin/borg create --list --progress --info --log-json --json --filter=AM -C lz4 --pattern='- /home/vandal/bb/file3.txt'   /home/vandal/Music/secret::litc2   /home/vandal/bb
```
Both worked as expected and ignored `file3.txt`, which was confirmed from `nfiles=2` (the directory had 3 files in total). And when the flag was left out all 3 files were added as expected.

As of right now borg seems to handle the fully quoted flags properly on their end based on my testing. Unless something changes in future versions, this borg command being shown in logs appears to work fine as expected.

src/vorta/borg/borg_job.py
```diff
@@ -234,9 +234,16 @@ def run(self):
...
-            cmd_log_tmp = [s.replace(" ", "\\ ") for s in self.cmd]  # escape whitespace - for logs
-            logger.info('Running command %s', ' '.join(cmd_log_tmp))
-            del cmd_log_tmp
+
+            # logs: put arguments with special strings in quotation marks
+            quote_strings = [' ', '*', '?', 're:']
+            cmd_tmp = self.cmd[:]
+            for i, arg in enumerate(cmd_tmp):
+                if any(quotestr in arg for quotestr in quote_strings):
+                    cmd_tmp[i] = arg.replace(arg, "'" + arg + "'")  # add quote
```
VandalByte, line 243: Correct me if I'm wrong but I don't see the need to use `replace()` here, since you are replacing the whole string. So this line could be something like:
```python
cmd_tmp[i] = "'" + arg + "'"
```
I don't prefer using f-strings, cause I have seen a case where it caused issues with older python versions in borgmatic.

goebbe:
@VandalByte Thanks for reviewing the code!

> I saw that code to handle that was removed from previous commit. Even though `argparse` might throw an error on something like this, borg seems to handle it fine internally.
... 
> As of right now borg seems to handle the fully quoted flags properly on their end based on my testing. Unless something changes in future versions, this borg command being shown in logs appears to work fine as expected.

I simplified the code of this pull request by quoting the whole argument, which works just fine with borg.
The idea to use this solution (simplification) came from a related bug report on borg: https://github.com/borgbackup/borg/issues/8578#issuecomment-2521084550  
cdchristensen pointed out that this is one supported way for quoting special strings in borg, Thomas Waldman (the main dev of borgbackup) gave thumps up.  Therefore, I guess this typo of quoting should be fine. 
I also ran a couple of tests to check if I can copy/ paste / run the borg command from the logs directly in a terminal - and it seems to work just fine.

src/vorta/borg/borg_job.py
goebbe: Thanks a lot! Using pythons string manipulation directly is a good idea, here. 
I changed this line, as suggested. 

With respect to the use of `shlex.quote(arg) ` here - this should not be necessary: 
- cmd is generated by using  `shlex.split( ) ` anyway 
- this pull request is only about the output in the logs -  my aim was to use the args "as used" by Vorta  - and just add quotation marks to allow copy/ paste/ run the borg command from the logs - for easier debugging.

VandalByte (approved):

VandalByte:
@goebbe I tested it out as well and everything looks good! 👍 

Sorry to be a stickler, but the linting test failed, try running `pre-commit` after staging the changes. It's probably due to the double space before the comment.

src/vorta/borg/borg_job.py
```diff
@@ -234,9 +234,16 @@ def run(self):
...
-            del cmd_log_tmp
+
+            # logs: put arguments with special strings in quotation marks
+            quote_strings = [' ', '*', '?', 're:']
+            cmd_tmp = self.cmd[:]
+            for i, arg in enumerate(cmd_tmp):
+                if any(quotestr in arg for quotestr in quote_strings):
+                    cmd_tmp[i] = "'" + arg + "'"  # add quotes
+
+            logger.info('Running command: %s', ' '.join(cmd_tmp))
```
m3nu, line 245: Much better than before. One thing I'd change: We can make `cmd_tmp` a little more descriptive. How about something like `cmd_args_to_log`?

goebbe: Done ! 
Using speaking names!  Thanks for the suggestions.

m3nu (approved):

m3nu:
Thanks for the addition @goebbe and for reviewing @VandalByte!

PR #2207, opened 2025-02-16
Fix for Validation Interval can be set more frequent than Backup Interval

### LINKED ISSUE #1971: Validation Interval should not be more frequent than Backup Interval
### Description

The Validate Interval QSpinBox in Schedule Tab allows me to set validation scheduling to be more frequent than the Backup Scheduling. For example I can set Backups to run every 6 weeks but set Validation to run every week.
Unless I'm reading the code wrong, doesn't validation using BorgCheckJob run only post-backup? Thus vorta shouldn't allow me to set Validation interval less than Backup interval.

### Reproduction

### OS

fedora 39

### Version of Vorta

0.9.1

Pip

### Version of Borg

_No response_

### Logs

_No response_

VandalByte:
### Description
This PR resolves the issue in the Schedule Tab where the Validate Interval QSpinBox allowed users to set the validation schedule to run more frequently than the backup schedule. This behavior was problematic since validation (using BorgCheckJob) only runs after a backup, meaning that it shouldn't be scheduled more frequently than the backup itself.

With this change, Vorta now ensures that the validation interval cannot be set to a value smaller than the backup interval, enforcing a more logical and consistent scheduling behavior.

### Related Issue
Resolves #1971

### Motivation and Context
- This change is required to enforce logical consistency between the backup and validation intervals in the Vorta scheduling system. 
- Previously, users were able to set the validation interval to run more frequently than the backup interval, which could lead to issues, as validation (via BorgCheckJob) only occurs after a backup.
### How Has This Been Tested?
- Tested the UI in the Schedule Tab to ensure that the Validate Interval QSpinBox no longer allows validation intervals smaller than the backup interval.
- Attempted to set a validation interval shorter than the backup interval, and confirmed that the UI prevents this configuration.

### Discussion:
- I've added a new function `on_validation_change()` to `/src/vorta/views/schedule_page.py`.
    - The function checks if Validation is enabled and if the Backup schedule is set to a periodic value with the unit `'week'`.
    - Since validation is measured in weeks, other units should not interfere with this logic.
    - The function ensures that the validation interval cannot be set to a value lower than the backup interval.
- On starting Vorta, the function is executed to set the validation interval value based on the backup interval.
- I also modified the `on_scheduler_change()` function to ensure this validation runs when the scheduler count or unit is changed.
- These changes prevent the user from setting the validation interval lower than the backup interval in the UI.

### Types of changes

- [ ]  Bug fix (non-breaking change which fixes an issue)
- [x]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

### DIFF
--- a/src/vorta/views/schedule_page.py
+++ b/src/vorta/views/schedule_page.py
@@ -89,10 +89,27 @@ class SchedulePage(SchedulePageBase, SchedulePageUI, BackupProfileMixin):
         self.draw_next_scheduled_backup()
 
     def on_validation_change(self):
-        """Updates the minimum value of the validation week counter based on the schedule interval"""
-        if self.validationCheckBox.isChecked() and self.scheduleIntervalUnit.currentData() == 'weeks':
-            self.validationWeeksCount.setValue(self.scheduleIntervalCount.value())
-            self.validationWeeksCount.setMinimum(self.scheduleIntervalCount.value())
+        """Ensures validation interval is not smaller than the backup interval."""
+        if self.validationCheckBox.isChecked():
+            backup_unit = self.scheduleIntervalUnit.currentData()
+            backup_value = self.scheduleIntervalCount.value()
+
+            # Backup interval to weeks conversion
+            if backup_unit == 'weeks':
+                min_validation_weeks = backup_value
+            elif backup_unit == 'days':
+                min_validation_weeks = backup_value / 7
+            elif backup_unit == 'hours':
+                min_validation_weeks = backup_value / (7 * 24)
+            elif backup_unit == 'minutes':
+                min_validation_weeks = backup_value / (7 * 24 * 60)
+            else:
+                min_validation_weeks = 1  # Default case, should not happen
+
+            # Set the validation weeks count
+            min_validation_weeks = max(1, int(min_validation_weeks))  # Ensures minimum 1 week
+            self.validationWeeksCount.setValue(min_validation_weeks)
+            self.validationWeeksCount.setMinimum(min_validation_weeks)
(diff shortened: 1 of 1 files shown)


VandalByte:
I realized this issue affects other units as well.
For example, in periodic backup, if the interval is set to `12 days` and validation is `1 week`, the same issue occurs.

So, I added checks for all units and ensured that the minimum value restriction applies only to 'Backup periodically'

m3nu:
> Previously, users were able to set the validation interval to run more frequently than the backup interval, which could lead to issues, as validation (via BorgCheckJob) only occurs after a backup.

What are the issues you talk about here?

VandalByte:
@m3nu 
So I haven't covered the full repo yet, I just looked at the UI part, but I did some backtracking, and from what I understand if the time period is set like mentioned above, when `BorgCheckJob.prepare()` runs, it would likely fail as it can't find any backups and return `msg['ok'] = False` and this would be logged -> `'Conditions for backup not met. Aborting.'`.

The issue is actually handled, and there shouldn't be any issues with normal functioning of Vorta, my PR adds the logical consistency at UI level. Please do correct me if I'm wrong somewhere.

m3nu:
I see 2 options here:

1. Add the code you suggested to automatically adjust validation interval. We need to maintain this new code in the future and possibly adjust it if scheduling changes.
2. Leave it as it is now. There may be confusion because validation doesn't run as often.

Not sure which one is better.

m3nu:
Thanks for testing! Good to know.

PR #2213, opened 2025-03-06
Fix bug with incorrect renaming of archives after aborting edit

### LINKED ISSUE #2041: Vorta displays the same name for all archives
### Description

1. Go to "Archives" tab
2. Double-click an archive name
3. Press escape to abort editing the name
4. Click "extract" for an archive
5. :warning: All archives get given name of the double-clicked archive :warning: 
6. The extract window appears

This also affects any archives created after this point too.

Screenshot_20240719_150220-1

The archives aren't _actually_ renamed, as quitting and restarting Vorta will have them return to their correct names.
However, in the bugged state, the incorrect names will be used for other functions, such as diffing (which results in an exception because it expects the names to be unique keys).

### Reproduction

### OS

Debian Testing, KDE Plasma 5.27.11

### Version of Vorta

0.9.1

Distribution package

### Version of Borg

1.4.0

### Logs

```pytb
2024-07-19 15:14:28,563 - vorta.keyring.abc - DEBUG - No module named 'objc'
2024-07-19 15:14:28,585 - vorta.keyring.abc - DEBUG - Using VortaKWallet5Keyring
2024-07-19 15:14:28,585 - vorta.borg.borg_job - DEBUG - Using VortaKWallet5Keyring keyring to store passwords.
2024-07-19 15:14:28,586 - vorta.keyring.kwallet - DEBUG - Retrieved password for repo ssh://blah@blah.repo.borgbase.com/./repo
2024-07-19 15:14:28,594 - vorta.borg.jobs_manager - DEBUG - Add job for site 2
2024-07-19 15:14:28,594 - vorta.borg.jobs_manager - DEBUG - Start job on site: 2
2024-07-19 15:14:28,603 - vorta.borg.borg_job - INFO - Running command /usr/bin/borg list --info --log-json --json-lines --format {mode}{user}{group}{size}{isomtime}{path}{source}{health}{NL} ssh://blah@blah.repo.borgbase.com/./repo::elyvilon-2024-07-17-182346
2024-07-19 15:14:30,105 - vorta.borg.jobs_manager - DEBUG - Finish job for site: 2
2024-07-19 15:14:30,105 - vorta.borg.jobs_manager - DEBUG - No more jobs for site: 2
2024-07-19 15:14:30,106 - vorta.scheduler - INFO - Setting timer for profile 1
2024-07-19 15:14:30,106 - vorta.scheduler - DEBUG - Scheduling next run for 2024-07-19 21:39:55.077533
```

VandalByte:
### Description
This PR fixes a bug where renaming an archive and then canceling the operation by pressing the ESC key caused subsequent actions (Extract, Diff, etc.) to rename all archives in the list to the attempted name in UI level.

**Changes made:**
-  ~~Created a delegate `EscKeyDelegate` to catch the ESC key press event when an item in table is being edited.~~
>Attaching event filter directly to archive table did not work because while editing events are then handled by the editor widget, not the viewport.

- Introduced a flag in `cell_changed` and `cell_double_clicked` to track whether the edit was completed or canceled.

### Related Issue
Fixes #2041

### Motivation and Context
This issue was originally reported by [romlok](https://github.com/borgbase/vorta/issues/2041#issue-2418901982), where canceling the renaming of an archive by pressing the ESC key caused all archives to be incorrectly renamed. The bug affects operations such as extracting and diffing, resulting in incorrect archive names and exceptions due to non-unique keys.

### How Has This Been Tested?
- Tested the UI by running the build and verifying the fix.
- Tested all the buttons that previously triggered the behavior, ensuring the issue was resolved.

### Types of changes

### Checklist:


VandalByte:
I just encountered another bug, renaming the archive now doesn't rename it in the backend and logs that archive not found (new archive name) and refreshes the list to previous state, I'll update the PR once I fix it.

VandalByte:
**Newly added changes:**
- Fixed the rename borg job not being run after archive name change (cell change).
- Added a `edit_delegate` to check the editor close to avoid race condition.
- Added a method `on_editing_finished` to facilitate the transition to `cell_changed` by avoiding race condition, where `cell_changed` was executed before the `cell_double_clicked`.
- Fixed the archive name not returning to original name when faced with a blank value.
- Removed ESC delegate as it was not doing anything.

m3nu:
It's ready in your opinion? Then I'll start testing it locally.

VandalByte:
@m3nu yeah it's ready for review. I was looking at ways to avoid using the global variable, but doing so would require significant changes to existing methods and might lead to potential issues down the line. So, I don't think that going to be the best option here.

m3nu:
Works as expected locally.

If you want, I saw one related thing that could be improved in the same feature: It says "Task started" when renaming. If this code is specific to renaming, why not change this to "Renaming archive..." (the 3 dots should be used for stuff that's ongoing)

m3nu (dismissed):

VandalByte:
@m3nu I have set the status accordingly but it's visible only for a split second then it shows "Task started" then the following statuses. One way I found is to suppress the text in the `_set_status` like this:
```python
def _set_status(self, text):
    if text == "Task started":
        return
    self.mountErrors.setText(text)
    self.mountErrors.repaint()
```
From the tests I did, this method doesn't seem to affect any other status messages. I haven't included this change in the latest commit.

m3nu:
Never mind then. Was just an idea.

VandalByte:
@m3nu adding this change I mentioned doesn't seem to affect any existing functionalities as of now. If we are not adding that change then the rest of the PR is ready to be merged.

m3nu:
No need to tag me each time.

VandalByte:
I tried adding the status update for rename operation in the bottom one as well (since it was missing), which apparently fixed the issue with the last one (above mentioned). So it's all good now.

https://github.com/user-attachments/assets/919d83d0-9517-4ce4-b308-d1bbe8b5c218

PR #2252, opened 2025-06-24
Add file dialog to exclude patterns

### LINKED ISSUE #2144: add a button to open a file chooser for excluding files/folders
#### The problem

There is no way (afaik) to exclude a folder by using a file dialog

#### Requested Solution

Sources View -> Manage excluded items -> Custom Tab -> Click "+" Button.
The very similar little Files/Folders/Paste Context Menu pops up for just excluding stuff.
This little context menu should of course preserve what the "+" Button  does right now.

Or add the new exclude file dialog button globally in the "patterns to exclude" view window.
There should be plenty of space in the bottom left corner. Although you'd either have 2 confusing "+ Buttons" or change the current functionality of the "+" Button 
#### Alternatives

Copy and pasting multiple Paths like a caveman gets old really quick.

### LINKED ISSUE #2260: When I am editing Exclude Items, it crashes
### Description

When adding new Exclude Items, after closing the dialog it crashes. App stays open but this dialog pops up

Image

### Reproduction

- [x] I tried to reproduce the issue.
- [x] I was able to reproduce the issue.

### OS

15.5 (24F74) MacOS

### Version of Vorta

0.10.3

Homebrew

### Version of Borg

1.4.0

### Logs

```pytb
OS: Darwin
Release: 24.5.0
Version: Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:43 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T8132
Date and Time: 2025-07-28 17:02:15
Borg Version: 1.4.0
Vorta Version: 0.10.3
  File "views/exclude_dialog.py", line 309, in custom_pattern_editing_finished
  File "views/exclude_dialog.py", line 305, in text_editing_finished
AttributeError: 'NoneType' object has no attribute 'text'
```

VandalByte:
### Description
This PR intends to add an option to use the file dialog to add the patterns directly into the "Custom" tab of exclusion dialog instead of adding then one click at a time.

### Related Issue

Resolves #2144 
Fixes #2260 (Unrelated bug fix)

### Motivation and Context
This option makes it easy for users to add the patterns by directly selecting files/folders from the file dialog and a pattern to exclude the same will be auto generated and added to the Custom list, which you can edit later on if needed.

### How Has This Been Tested?
Hasn't been tested yet, still a WIP.

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/views/exclude_dialog.py
+++ b/src/vorta/views/exclude_dialog.py
@@ -456,7 +456,7 @@ class ExcludeDialog(ExcludeDialogBase, ExcludeDialogUi):
         # Both files and folders use the same pattern format (fm:*/name),
         # confirmed through testing to correctly exclude them during backup creation.
         for path in paths:
-            patterns.append(f"fm:*/{Path(path).name}")
+            patterns.append(f"pf:{path}")
         return patterns
 
     def open_exclude_filedialog(self):


VandalByte:
As of right now I have two UI design concepts, it would be great to have some feedback

First one is this, adding two separate buttons, `Exclude File` to open the file dialog and `Exclude Pattern` to add one pattern at a time (functions same as the current plus button). 

Screenshot 2025-06-24 190429

Second one has a single plus button having a submenu with two choices as mentioned above, a lot cleaner on UI front but adds an extra click (for every pattern you manually add).
Screenshot 2025-06-24 192416

m3nu:
First option seems better since we have enough space and it's more explicit.

VandalByte:
By keeping the `+` button as it is, this is the current design:

Screenshot 2025-07-11 194747

Now for the button text other options I can think of is
- `Browse Exclusions`
- `Browse Exclusion Files`
- `Select Files and Folders to Exclude` (seems too long)
- `Add Exclusions`
- `Choose Exclusion Files`
- `Select Exclusion Files`

Making that `+` button into named button as shown in my previous [comment](https://github.com/borgbase/vorta/pull/2252#issuecomment-3000692840) left the `-` button looking out of place and changing the `-` button to a named button will further increase the overall button width in the dialog. Would love to have some feedback on this.

I'm starting on the implementation since the layout design has already been fixed (only the button label mentioned above can change, if needed).

m3nu:
Looks good to me.

VandalByte:
I have added the last few changes, works fine during testing, both files and folders given are being excluded as expected.

m3nu:
(rebased)

andrewcrook:
I would test for blank strings as well [`s.strip() == ‘’`](https://github.com/borgbase/vorta/issues/2260#issuecomment-3148485998) under `text_editing_finished()` just to be complete  #2260

VandalByte:
Also fixes #2260

m3nu:
Screenshot 2025-08-28 at 12 33 05

Excluding this pattern and a specific folder isn't the same. When a user selects a specific folder they would expect the exclusion to only apply to this one folder, not all with the same name.

m3nu:
Here a minimal reproducible example of the issue:

Given this source folder
`~`

And this excluded folder line
`fm:*/Downloads`

which was selected in the UI as ~/Downloads

I have these folders and files:
`~/Downloads`
`~/Desktop/Downloads/important vorta file.png
`
And only want to exclude ~/Downloads, but never had the intention to exclude ~/Desktop/Downloads and the important file in it.

Unfortunately after testing the feature I can confirm that **both folders are excluded**, which isn't the user's intention and very unexpected.

Screenshot 2025-08-29 at 06 40 35

VandalByte:
I have changes the pattern from `fm:` to `pf:`
> This pattern style is (only) useful to match full paths. This is kind of a pseudo pattern as it can not have any variable or unspecified parts - the full path must be given. pf:root/file.ext matches root/file.ext only. A leading path separator is always removed.

image
image


I have tested out the same and now it should work as expected. Also the wildcard path have been replaced with complete path of file/folder selected, as required by the `pf:` pattern.

Issue #2316, opened 2025-12-16
`on_scheduler_change` callback causes race conditions in updating BackupProfile, causing spurious catch-up runs


cincodenada:
### Description

In short: due to the controls loading their values sequentially, calling `on_scheduler_change` for every control change event results in BackupProfile's schedule fields being briefly reset to `0` at startup. This causes a race condition with the startup check for a "missed run": the missed-run check can happen before the controls fully load, which means it is run with e.g. a FixedTime of 0h0m, that is, midnight.

When this happens, Vorta incorrectly runs a catch-up run, because given a run the previous day, it calculates the "next run" as happening at midnight of the current day, which is always in the past, and thus "missed".

### Details

`on_scheduler_change` gets called three times at startup, once for each field (`intervalUnit`, `intervalCount`, `fixedTime`). Because `on_scheduler_change` unconditionally updates _all_ fields of `profile`, the following sequence occurred during my testing, with time since first log line on the left:
|time|update source|interval_unit|interval_count|fixed_hour|fixed_minute|
|---|---|---|---|---|---|
|350ms|intervalUnit|(set)|0|0|0|
|440ms|intervalCount|(set)|(set)|0|0|
|480ms|fixedTime|(set)|(set)|(set)|(set)|(set)|

In my tests, the missed-run check calculated its `next_time` at 385ms, thus getting the incorrect midnight `fixedTime` set.

### Proposed fixes

The simplest reasonable fix would probably be to use separate callbacks for each update that only updates the relevant fields. This would result in more database updating than is strictly necessary, but that's probably fine.

Debouncing the callback would fix this instance, but leaves it open to slower machines running into the same issue.

Perhaps a better fix would be to keep a flag for each control that is set the first time it fires an update callback, so that we could wait to update the profile until every control has been updated. Or, alternately, until the control is updated a second time - since the first time is just from loading the data from `profile` anyway (and thus we don't need to save those values _back_ to `profile`). I doubt it is possible for a user to e.g. update `intervalUnit` before `fixedTime` has loaded, so updating on the second call from any given control should be safe.

I'd be glad to open PRs for any of the above approaches, let me know if you have preferences (or a different approach)

### Background

I was investigating #2312 and noticed some strange log lines in there - it's clear that the user had set the time to 18:00, and yet these lines were present at the beginning of the log:
```
2025-12-10 09:48:06,138 - vorta.scheduler - INFO - Setting timer for profile 1
[...]

PR #2317, opened 2025-12-17
fix: Delay updating schedule until after fields have been initialized

cincodenada:
### Description

This is a fix for #2316. It changes the schedule field handling so that updates to the schedule fields don't propagate to the profile until after the fields are initialized (that is, loaded _from_ the profile).

This resolves a bug where we briefly reset schedule fields to 0 because we copied all field values to the profile after only updating the first (or second) field _from_ the profile. This in turn got in a race with the initial schedule check, and caused spurious "catchup" runs where Vorta calculated the next run to be at midnight (00:00) regardless of the user's settings, and thus always generating a "catchup" run, since midnight is the very beginning of the day.

I also modified the `init_db` fixture in tests slightly to support deferring the window load to the test itself. Unless a test uses the newly-introduced `window_load` fixture, `init_db` functions as it did before.

### Related Issue
Issue #2316 

### Motivation and Context
Solves a bug where fixed-time schedules would always generate a catch-up run if the application was started before the scheduled time, which also caused the scheduled runs to not take place.

### How Has This Been Tested?
I added a unit test specifically to test this bug, initially failing and then passing after the changes. I also manually tested as  follows:
 - Set a fixed-time schedule to run at 6pm
 - Ran a backup to start the schedule
 - Used the Linux utility `faketime` to simulate re-opening the app the next morning
 - Used `faketime` to simulate re-opening the app again at 10pm that same day

As expected, there was no catch-up run when opening the app in the morning, and it scheduled the 6pm run as anticipated. Opening the app at (simulated) 10pm _did_ do a catch-up run, as it was past the scheduled run time.

I also ran all other tests as well, which remain passing, other than the `test_scheduler_create_backup` test that fails in `main` on my machine too.

### Types of changes

### Checklist:


m3nu (commented):
## 🔍 Code Review

**PR:** #2317 - fix: Delay updating schedule until after fields have been initialized
**Branch:** `debug-scheduler-race` → `master`
**Changes:** +106 / -12 lines across 6 files

### 📋 Summary

This PR correctly fixes a race condition where schedule fields were being reset to 0 during UI initialization, causing spurious catchup runs. The fix uses a flag to defer writing to the profile until after all fields have been populated from the profile.

### ✅ What's Good
- Root cause correctly identified and addressed
- Clean implementation of `loaded` signal in `MainWindow` for coordinating initialization
- Good test coverage with parameterized test cases covering multiple scenarios
- Test fixture design (`window_load`) allows fine-grained control over initialization order

### 🔍 Review Details
7 inline comment(s) added.

| Severity | Count |
|----------|-------|
| 🔴 Error | 0 |
| 🟡 Warning | 2 |
| 🔵 Info | 2 |
| 💡 Suggestion | 2 |
| ✅ Good | 1 |

### 📊 Verdict: COMMENT 💬

Overall this is a solid fix for a tricky race condition. The suggestions below are minor and mostly about code style and cleanup.

---

src/vorta/views/schedule_page.py
```diff
@@ -63,23 +67,30 @@ def __init__(self, parent=None):
 
         self.app.scheduler.schedule_changed.connect(lambda pid: self.draw_next_scheduled_backup())
         self.populate_from_profile()
+        self.hasPopulatedScheduleFields = True
 
         # Listen for events
         self.app.profile_changed_event.connect(self.populate_from_profile)
 
     def on_scheduler_change(self, _):
```
m3nu: 💡 **[code-clarity]** Consider adding a brief comment

Good fix! The logic correctly moves the profile updates outside the for-loop so all fields are saved regardless of which radio button is checked. A brief comment explaining why the flag check is needed could help future maintainers understand the race condition being prevented.

**Example:**
```python
def on_scheduler_change(self, _):
    # Prevent writing default values back to profile during initial field population
    if not self._has_populated_schedule_fields:
        return
```

tests/unit/test_scheduler.py
```diff
@@ -191,3 +191,51 @@ def test_fixed(clockmock, passed_time, scheduled, now, hour, minute):
...
+        start_time=last_time,
+        end_time=last_time,
+    )
+    event.save()
+
+    # We have to replace the scheduler because of shared state (namely, pauses)
+    # We also reload because app init does that (via `set_borg_details_result`)
+    qapp.scheduler = VortaScheduler()
+    qapp.scheduler.reload_all_timers()
+    window_load()
```
m3nu: 💡 **[resource-management]** Consider cleaning up original scheduler

Replacing `qapp.scheduler` creates a new `VortaScheduler` instance, but the original scheduler may still have active timers or connections. Consider explicitly cleaning up the original scheduler first, or verify that the garbage collector handles this correctly.

**Potential improvement:**
```python
# Clean up existing scheduler resources
old_scheduler = qapp.scheduler
# ... any cleanup if needed ...
qapp.scheduler = VortaScheduler()
```

This might already be handled by the test teardown, just flagging for consideration.

Issue #2318, opened 2025-12-17
Tests don't run on non-Darwin platforms


cincodenada:
### Description

When I try to run `pytest` on non-Dariwn platforms, it immediately fails out with the following error:
```
________________________________ ERROR collecting tests/network_manager/test_darwin.py ________________________________
ImportError while importing test module '/home/ell/clone/vorta/tests/network_manager/test_darwin.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.13/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/network_manager/test_darwin.py:5: in <module>
    from vorta.network_status import darwin
src/vorta/network_status/darwin.py:5: in <module>
    from CoreWLAN import CWInterface, CWNetwork, CWWiFiClient
E   ModuleNotFoundError: No module named 'CoreWLAN'
```

### Reproduction

- [x] I tried to reproduce the issue.
- [x] I was able to reproduce the issue.

### OS

Ubuntu 24.04, xfce

### Version of Vorta

0.11

### What did you install Vorta with?

Other

### Version of Borg

1.4.0

### Logs

```pytb

```

aman-coder03:
Can i work on this?

m3nu:
What's left to implement, @aman-coder03 after https://github.com/borgbase/vorta/pull/2319? Those test will likely not work on Github anyways since they are VMs. There should be better issues to work on.

PR #2319, opened 2025-12-17
Skip darwin tests on non-darwin platforms

cincodenada:
### Description
Adds a simple check to skip a darwin-only test on non-darwin platforms

### Related Issue
#2318 

### Motivation and Context
I want to be able to run the tests on my Linux machine :)

### How Has This Been Tested?
Ran tests, they all ran except for the darwin one which was skipped as expected

### Types of changes

### Checklist:

### DIFF
--- a/tests/network_manager/test_darwin.py
+++ b/tests/network_manager/test_darwin.py
@@ -3,10 +3,9 @@ from unittest.mock import MagicMock
 
 import pytest
 
-if sys.platform != 'darwin':
-    pytest.skip("Skipping darwin-only network tests", allow_module_level=True)
+pytestmark = pytest.mark.skipif(sys.platform != 'darwin', reason="macOS only")
 
-from vorta.network_status import darwin
+from vorta.network_status import darwin  # noqa: E402
 
 
 def test_get_current_wifi_when_wifi_is_on(mocker):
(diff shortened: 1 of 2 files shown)


m3nu:
Thanks for the fix. The right thing to skip this conditionally. I do wonder why tests aren't failing on Github. Much of the output is already mocked. Maybe due to different versions used somewhere.

cincodenada:
Ya know I was curious about why it wasn't failing in the Github tests too, and so I just looked into it a bit, and pretty sure it's becuase we're just not running this test at all in the Github actions. Looks like the Actions explicitly run `make test-unit` and `make test-integration`, but this test (and one other for NetworkManager) are a secret _third_ category of tests, `network_manager`, which doesn't have a corresponding Makefile command.

I wonder if the Github runners even have enough of a desktop environment to have the `network_manager` tests even do anything, come to think of it. May be worth a shot though, I guess.

m3nu:
That's likely the reason. The network tests didn't work on GH so they put them separately. You can still run them locally with `uv run pytest tests/...` Can you add the skipping to both. Like this:

For test_darwin.py (macOS only):
```
import sys
pytestmark = pytest.mark.skipif(sys.platform != 'darwin', reason="macOS only")
```

For test_network_manager.py (Linux only):
```
import sys
pytestmark = pytest.mark.skipif(sys.platform != 'linux', reason="Linux only")
```                                                                                                                                  
This applies the skip condition to all tests in the file automatically. When you run tests on the wrong platform, pytest will skip the entire file with the reason displayed. [Source](https://docs.pytest.org/en/stable/example/markers.html#marking-whole-classes-or-modules)

Issue #2362, opened 2026-01-26
Add comprehensive type annotations and enforce type checking in CI


m3nu:
## Summary

Add comprehensive type annotations throughout the Vorta codebase and enforce type checking via CI. This will improve code quality, catch bugs earlier, enable better IDE support, and make the codebase more maintainable.

## Current State

- **Type coverage**: ~22% of files import `typing`, ~32% have actual type hints
- **No type checker**: Neither mypy nor pyright is configured
- **No py.typed marker**: Package cannot be used as a typed library
- **Inconsistent annotations**: Some files have excellent type hints (e.g., `network_status/abc.py`, `store/settings.py`), while others have none (e.g., `borg/borg_job.py`, `views/repo_tab.py`)
- **Minimum Python version**: Currently 3.8

### Examples of current inconsistency

**Well-typed file** (`src/vorta/network_status/abc.py`):
```python
def is_network_metered(self) -> bool:
    """Is the currently connected network a metered connection?"""

def get_current_wifi(self) -> Optional[str]:
    """Get current SSID or None if Wifi is off."""

def get_known_wifis(self) -> List['SystemWifiInfo']:
    """Get WiFi networks known to system."""
```

**Untyped file** (`src/vorta/borg/create.py`):
```python
def process_result(self, result):
    if result['returncode'] in [0, 1] and 'archive' in result['data']:
```

### 1. Bump minimum Python version to 3.10

**Rationale:**
- Python 3.8 reached end-of-life in October 2024
- Python 3.9 reaches end-of-life in October 2025
- Python 3.10+ provides better typing ergonomics:
  - `X | Y` union syntax instead of `Union[X, Y]`
  - `list[str]` instead of `List[str]` (builtin generics)
  - `ParamSpec` and `TypeAlias` for advanced patterns
  - Better error messages in type checkers
- PyQt6 (which Vorta uses) requires Python 3.6.1+ minimum, so this is compatible
- Most Linux distributions shipping Vorta have Python 3.10+ available

**Alternative:** Stay at 3.9 minimum if 3.10 is too aggressive, but strongly recommend 3.10+.

### 2. Choose and configure a type checker

**Recommended: mypy**

Reasons:
- Industry standard, well-documented
- Good PyQt/Qt stubs available (`PyQt6-stubs`)
- Mature, stable tool
- Integrates well with pre-commit and CI

**Configuration to add to `pyproject.toml`:**
```toml
[tool.mypy]
python_version = "3.10"
[...]

PR #2384, opened 2026-02-04
 Update requires-python from >=3.8 to >=3.10

mr-raj12:
PyQt6 requires Python >=3.10 and CI only tests 3.10+. Also update                                                     tox env_list to match CI coverage and add Python version classifiers.

  Closes #2380
### Description                                                                                                        
  Update Python version metadata in `pyproject.toml` and clean up stale references to Python 3.8/3.9 across project   
  configuration.

  **Changes across 2 files:**

  | File | Change |
  |------|--------|
  | `pyproject.toml` | `requires-python`: `>=3.8` → `>=3.10` |
  | `pyproject.toml` | Add explicit Python version classifiers (3.10, 3.11, 3.12, 3.13) |
  | `pyproject.toml` | tox `env_list`: remove `py38`/`py39`, add `py313` |
  | `.github/scripts/generate-matrix.sh` | Remove stale exclude rule referencing Python 3.8 + Borg 2.0.0b12 |
  | `uv.lock` | Regenerated to reflect new `requires-python` constraint |

  ### Related Issue

  Closes #2380

  ### Motivation and Context

  The declared `requires-python = ">=3.8"` is misleading — PyQt6 (a core dependency) requires Python >=3.10, CI only  
  tests 3.10+, and Python 3.8 is EOL. Users on Python 3.8/3.9 who install Vorta hit confusing PyQt6 errors instead of 
  a clear version mismatch from pip. The tox env_list also listed `py38`/`py39` but was missing `py313`, which is     
  already tested in CI.

  ### How Has This Been Tested?

  - `uv sync` — validates `pyproject.toml` parses correctly and dependencies resolve
  - `pre-commit run --all-files` — all 9 checks passed (ruff, ruff-format, whitespace, etc.)
  - CI matrix JSON validated — `generate-matrix.sh` output is valid JSON
  - These are metadata-only changes with no functional code impact

  ### Types of changes

  - [x] Bug fix (non-breaking change which fixes an issue)
  - [ ] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to change)

  ### Checklist:

  - [x] I have read the [CONTRIBUTING](https://vorta.borgbase.com/contributing/) guide.
  - [x] My code follows the code style of this project.
  - [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.
  - [ ] I have added tests to cover my changes.
  - [x] All new and existing tests passed.

[...]

PR #2402, opened 2026-02-09
Add mypy type checking infrastructure and core module annotations (#2362)

mr-raj12:
### Description

Phases 1+2 of #2362 — sets up mypy infrastructure and adds type annotations to core modules.

**Phase 1 (infrastructure):**
- Bump `requires-python` from `>=3.8` to `>=3.10` (3.8/3.9 are EOL)
- Configure mypy with permissive defaults in `pyproject.toml`
- Add `mypy`, `PyQt6-stubs`, `types-peewee` to dev dependencies
- Add informational mypy step to CI (`continue-on-error: true`)
- Create `py.typed` marker file (PEP 561)
- Remove stale py3.8 exclude from CI matrix

**Phase 2 (core module annotations):**
- Add `from __future__ import annotations` to all core modules (PEP 563)
- Type all functions in `store/models.py`, `connection.py`, `settings.py`, `migrations.py`
- Type all functions in `network_status/abc.py`, `darwin.py`, `network_manager.py`
- Type all methods in `scheduler.py`
- Add per-module strict override (`disallow_untyped_defs = true`) for fully clean modules
- Fix incorrect return type in `settings.py` (`List[Dict[str,str]]` → `list[dict[str,Any]]`)
- **Fix control-flow bug in `network_manager.py`**: previously, if the active connection wasn't wireless, `get_current_wifi()` fell through the `try` block without returning — the `return None` only existed inside the `except`. Now all non-returning paths explicitly hit `return None`.

**Review feedback addressed:**
- Replaced `assert con is not None` in `init_db()` with `elif con is not None:` guard — the assert would crash when called without arguments from `profile_export.py`
- Removed unused `formatted_time()` stub from `ArchiveModel`
- Replaced old-style typing imports (`Dict`, `Tuple`, `Union`, `Optional`) with modern syntax (`dict`, `tuple`, `X | Y`) in `scheduler.py`

**Notes:**
- `warn_return_any` and `check_untyped_defs` set to `false` initially — enabling them surfaces 100+ errors in unannotated modules. Will be tightened in Phase 5.
- `migrations.py` excluded from strict override — Peewee's `_meta` table name access generates ~30 false positives from incomplete type stubs.
- `scheduler.py` annotated but not in strict override — ScheduleStatus construction needs refactoring before strict enforcement.
- mypy not added to pre-commit — PyQt6-stubs is incompatible with mirrors-mypy isolated environments. Deferred to Phase 5.
- Full error count increases from ~137 to ~179 because annotated function bodies are now checked. Strict-enforced modules have 0 errors.

### Related Issue

Partial implementation of #2362 (Phases 1+2 of 5)

### Motivation and Context

The codebase has inconsistent type annotations with no type checker enforcing correctness. This PR adds infrastructure and annotates the core modules as a foundation for gradually typing the rest of the codebase.

### How Has This Been Tested?

- `uv run mypy src/vorta/store/ src/vorta/network_status/` — 0 errors on strict modules
- `uv run pre-commit run --all-files` — all hooks pass
- `uv run mypy src/vorta` — runs without crashes (179 errors, all in future-phase modules)

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,10 +31,9 @@ jobs:
         shell: bash
         run: uv run pre-commit run --all-files --show-diff-on-failure
 
-      - name: Type check with mypy (informational)
-        continue-on-error: true
+      - name: Type check with mypy
         shell: bash
-        run: uv run mypy src/vorta
+        run: uv run mypy src/vorta/store/ src/vorta/network_status/
 
   prepare-matrix:
     runs-on: ubuntu-latest
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,6 +141,10 @@ ignore_missing_imports = true
 module = ["vorta.store.models", "vorta.store.connection", "vorta.store.settings", "vorta.network_status.*"]
 disallow_untyped_defs = true
 
+[[tool.mypy.overrides]]
+module = ["vorta.store.migrations"]
+ignore_errors = true
+
 [tool.pylint.master]
 extension-pkg-whitelist = "PyQt6"
 load-plugins = ""


m3nu:
Did you evaluate other type‑checking tools, like the upcoming ty?

mr-raj12:
mypy was specifically recommended in #2362 for its mature ecosystem, PyQt6-stubs support, and stable CI
  integration — so I went with that. Also, ty looks promising but is still in early alpha and not production-ready yet.   Since the annotations are tool-agnostic, switching later would be straightforward if ty matures. 
  Open for changes as you say to switch if needed!

m3nu:
If mypy already support pyqt then that's a big plus for us naturally.

I think there are one or two others that are also widely used. From FB or MS maybe? Don't recall the name.

mr-raj12:
i guess pyright (Microsoft) and pyre (Meta). pyright is pretty popular but its PyQt6 stub support isn't as mature as mypy's yet, and pyre has a smaller community around it. Either way the annotations are just standard Python, so switching checkers later is no big deal if something better comes along. Btw rebased this onto current master — the requires-python bump from #2384 and the network changes from #2176 are now incorporated. So this PR just adds the mypy config + annotations on top.

m3nu:
I see you did the research. Will try to review this today.

m3nu (commented):
Thanks for working on this! The infrastructure setup (Phase 1) is solid — permissive mypy defaults with strict per-module overrides and an informational CI step is exactly the right gradual approach.

Phase 2 annotations are mostly good, with one must-fix and a few minor items. See inline comments.

### Summary

| Issue | Severity |
|-------|----------|
| `assert con is not None` in `connection.py` crashes during profile import | **Must fix** |
| `formatted_time -> None` cements a broken stub | Low |
| Mixed old/new typing style (`List` vs `list`) despite `__future__` import | Nit |
| Unused typing imports (`Dict`, `Tuple`, `Union`) in scheduler.py | Nit |

PR #2411, opened 2026-02-22
Add Clear button to log display area

### LINKED ISSUE #1618: FR: A clear button for the log messages in vorta's application window.

thesakshidiggikar:
This PR adds a "Clear" button to the log display area.

- The button is enabled only when log text is present.
- Clicking it clears the log text.
- The button is disabled again after clearing.

Closes #1618

### DIFF
--- a/src/vorta/views/main_window.py
+++ b/src/vorta/views/main_window.py
@@ -12,6 +12,7 @@ from PyQt6.QtWidgets import (
     QFileDialog,
     QListWidgetItem,
     QMessageBox,
+    QPushButton,
     QToolTip,
 )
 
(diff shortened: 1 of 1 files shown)


m3nu:
Wasn't this solved by truncating the output instead?

thesakshidiggikar:
@m3nu  Thanks for the question!

From what I observed, truncating the output limits the log size automatically,
but it doesn’t allow users to manually clear the log when needed.

This PR introduces an explicit "Clear" button so users can reset the log
display instantly without waiting for truncation.

If truncation was intended to fully replace this functionality,
I’m happy to adjust or close this PR accordingly.

Issue #2361, opened 2026-01-26
Refactor views: Add BaseTab class, ViewModel layer, and reorganize dialogs


m3nu:
### Summary

This issue proposes a comprehensive refactoring of the views layer to reduce code duplication, improve maintainability, and establish clearer architectural patterns. The current implementation has significant boilerplate duplication across all 10 tab/page components.

### Motivation

The views layer currently has:
- ~200-250 lines of duplicated boilerplate code across tabs/pages
- Signal connection and cleanup patterns repeated in 9 out of 10 components
- A thin `BackupProfileMixin` that could be replaced with proper dependency injection
- Large monolithic tab files (ArchiveTab is 900+ lines)
- All dialogs flat in the views directory with no organization

#### 1. Add BaseTab class to reduce duplicate code

**Current Problem:**
All tabs/pages repeat this pattern:
```python
def __init__(self, parent=None):
    self._palette_connection = QApplication.instance().paletteChanged.connect(...)
    self._profile_changed_connection = QApplication.instance().profile_changed_event.connect(...)
    self._backup_finished_connection = QApplication.instance().backup_finished_event.connect(...)
    self.destroyed.connect(self._on_destroyed)

def _on_destroyed(self):
    try:
        QApplication.instance().paletteChanged.disconnect(self._palette_connection)
    except (TypeError, RuntimeError):
        pass
    # ... repeat for each connection ...
```

**Proposed Solution:**
Create `BaseTab` (or `BaseProfileTab`) class that:
- Manages signal connections with automatic cleanup
- Provides template methods: `populate_from_profile()`, `set_icons()`
- Handles common `save_profile_attr()` / `save_repo_attr()` helpers
- Implements signal tracking to avoid manual disconnect boilerplate

**Affected Files:**
- `repo_tab.py` (~398 lines)
- `source_tab.py` (~373 lines)
- `archive_tab.py` (~900 lines)
- `schedule_tab.py` (~67 lines)
- `misc_tab.py` (~141 lines)
- `schedule_page.py`, `log_page.py`, `shell_commands_page.py`, `networks_page.py`

#### 2. Replace BackupProfileMixin with injection

**Current Problem:**
`BackupProfileMixin` in `src/vorta/store/models.py` is just:
```python
class BackupProfileMixin:
    def profile(self):
        return BackupProfileModel.get(id=self.window().current_profile.id)
```

This tightly couples views to the window hierarchy and makes testing difficult.

**Proposed Solution:**
[...]

PR #2422, opened 2026-03-13
Refactor views with BaseTab and introduce RepoTabViewModel

false200:
### Description

This PR starts the views-layer refactor from #2361 by introducing a shared `BaseTab`
abstraction, moving several views to injected profile access, and adding the first
ViewModel for `RepoTab`.

Changes included in this PR:

- added `BaseTab` to centralize:
  - active profile access via injection
  - tracked signal connections
  - automatic signal cleanup
  - shared `save_profile_attr` / `save_repo_attr` helpers
- updated several view classes to use `BaseTab` instead of repeating signal boilerplate
- wired profile injection from `MainWindow`
- added `RepoTabViewModel` and moved repo-related persistence logic out of `RepoTab`
- added unit tests for `BaseTab` and `RepoTabViewModel`

This is an incremental first slice of the refactor and does not yet include dialog
reorganization, splitting `ArchiveTab`, or adding ViewModels for the remaining tabs.

### Related Issue

Part of #2361

### Motivation and Context

The current views layer contains repeated signal connection and cleanup logic and relies
on `BackupProfileMixin`, which couples views to the window hierarchy and makes them
harder to test.

This change reduces duplicated boilerplate, introduces a cleaner injected profile-access
pattern, and establishes the first ViewModel-based separation of UI and persistence
logic. It is intended as groundwork for the larger refactor proposed in #2361.

### How Has This Been Tested?

Test environment:
- Windows
- `uv` project environment
- PyQt6 test setup already present in the repository

Passed:
- `uv run pre-commit run --files src/vorta/views/base_tab.py src/vorta/views/view_models/repo_tab_view_model.py src/vorta/views/repo_tab.py tests/unit/test_view_models.py`
- `uv run py -B -m pytest tests/unit/test_base_tab.py tests/unit/test_view_models.py tests/unit/test_misc.py tests/unit/test_schedule.py`

Notes:
- a broader repo/SSH-focused test run on this Windows setup showed some environment-sensitive failures:
  - `test_repo_unlink`
  - `test_repo_add_success`
  - `test_ssh_dialog_success`
  - `test_ssh_copy_to_clipboard_action`
  - `test_create`

Not applicable.

### Types of changes

### Checklist:

### DIFF
--- a/src/vorta/store/models.py
+++ b/src/vorta/store/models.py
@@ -56,6 +56,13 @@ class RepoModel(BaseModel):
     def is_remote_repo(self):
         return not self.url.startswith('/')
 
+    def is_shared_with_other_profiles(self, excluding_profile_id: int | None = None) -> bool:
+        """Return whether another profile still references this repository."""
+        query = BackupProfileModel.select().where(BackupProfileModel.repo == self)
+        if excluding_profile_id is not None:
+            query = query.where(BackupProfileModel.id != excluding_profile_id)
+        return query.exists()
+
     class Meta:
         database = DB
 
--- a/src/vorta/views/about_tab.py
+++ b/src/vorta/views/about_tab.py
@@ -5,8 +5,8 @@ from PyQt6 import QtCore, uic
 
 from vorta import config
 from vorta._version import __version__
-from vorta.store.models import BackupProfileMixin
 from vorta.utils import get_asset
+from vorta.views.base_tab import BaseTab
 from vorta.views.utils import get_colored_icon
 
 uifile = get_asset('UI/about_tab.ui')
@@ -15,12 +15,12 @@ AboutTabUI, AboutTabBase = uic.loadUiType(uifile)
 logger = logging.getLogger(__name__)
 
 
-class AboutTab(AboutTabBase, AboutTabUI, BackupProfileMixin):
+class AboutTab(BaseTab, AboutTabBase, AboutTabUI):
     refresh_archive = QtCore.pyqtSignal()
 
-    def __init__(self, parent=None):
+    def __init__(self, parent=None, profile_provider=None):
         """Init."""
-        super().__init__(parent)
+        super().__init__(parent=parent, profile_provider=profile_provider)
         self.setupUi(parent)
         self.versionLabel.setText(__version__)
         self.logLink.setText(
--- a/src/vorta/views/archive_tab.py
+++ b/src/vorta/views/archive_tab.py
@@ -32,7 +32,7 @@ from vorta.borg.prune import BorgPruneJob
 from vorta.borg.rename import BorgRenameJob
 from vorta.borg.umount import BorgUmountJob
 from vorta.i18n import translate
-from vorta.store.models import ArchiveModel, BackupProfileMixin, SettingsModel
+from vorta.store.models import ArchiveModel, SettingsModel
 from vorta.utils import (
     borg_compat,
     choose_file_dialog,
@@ -43,6 +43,7 @@ from vorta.utils import (
     pretty_bytes,
 )
 from vorta.views import diff_result, extract_dialog
+from vorta.views.base_tab import BaseTab
 from vorta.views.diff_result import DiffResultDialog, DiffTree
 from vorta.views.extract_dialog import ExtractDialog, ExtractTree
 from vorta.views.source_tab import SizeItem
@@ -65,18 +66,18 @@ class IconDelegate(QStyledItemDelegate):
         option.decorationSize = option.rect.size() - QtCore.QSize(0, 10)
 
 
-class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
+class ArchiveTab(BaseTab, ArchiveTabBase, ArchiveTabUI):
     prune_intervals = ['hour', 'day', 'week', 'month', 'year']
 
-    def __init__(self, parent=None, app=None):
+    def __init__(self, parent=None, app=None, profile_provider=None):
         """Init."""
-        super().__init__(parent)
+        super().__init__(parent=parent, profile_provider=profile_provider)
         self.setupUi(parent)
         self.is_editing = False  # track if the cell edit was completed or canceled
         self.mount_points = {}  # mapping of archive name to mount point
         self.repo_mount_point: Optional[str] = None  # mount point of whole repo
         self.menu = None
-        self.app = app
+        self.app = app or self.app
         self.toolBox.setCurrentIndex(0)
         self.repoactions_enabled = True
         self.renamed_archive_original_name = None
@@ -158,13 +159,12 @@ class ArchiveTab(ArchiveTabBase, ArchiveTabUI, BackupProfileMixin):
         self.set_icons()
 
         # Connect to events
-        self._palette_conn_1 = self.app.paletteChanged.connect(self.set_icons)
-        self._palette_conn_2 = self.app.paletteChanged.connect(self.populate_from_profile)
-        self.destroyed.connect(self._on_destroyed)
-        self.app.backup_finished_event.connect(self.populate_from_profile)
-        self.app.profile_changed_event.connect(self.populate_from_profile)
-        self.app.profile_changed_event.connect(self.toggle_compact_button_visibility)
-        self.app.backup_cancelled_event.connect(self.cancel_action)
+        self.track_palette_change()
+        self.track_palette_change(self.populate_from_profile)
+        self.track_backup_finished()
+        self.track_profile_change()
+        self.track_profile_change(self.toggle_compact_button_visibility)
+        self.track_signal(self.app.backup_cancelled_event, self.cancel_action)
 
     def set_icons(self):
         """Used when changing between light- and dark mode"""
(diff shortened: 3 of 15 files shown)


m3nu:
## Review: BaseTab + RepoTabViewModel Refactor

**Overall: The core idea is sound — request changes on execution details.**

The PR correctly identifies ~150 lines of real boilerplate duplication (signal connect/disconnect across 10 tab/page classes) and `BaseTab` is a clean solution. The cooperative inheritance, callable profile injection, and template methods are all well done. A few issues to address before this pattern gets replicated to remaining tabs.

---

### Issues

#### 1. `_without_signal_args` wrapping is inconsistent and sometimes unnecessary

`track_profile_change` and `track_backup_finished` both wrap callbacks with `_without_signal_args`. However:

- `profile_changed_event = pyqtSignal()` has **no arguments** — wrapping is pointless here. Qt already discards extra signal args that a slot doesn't accept.
- `backup_finished_event = pyqtSignal(dict)` does emit a `dict`, so wrapping is defensible (though Qt handles this too).
- In ArchiveTab, two connections to the *same* signal use different patterns:
  ```python
  self.track_profile_change(self.populate_from_profile)  # wrapped via _without_signal_args
  self.track_signal(self.app.profile_changed_event, self.toggle_compact_button_visibility)  # not wrapped
  ```

**Suggestion:** Only use `_without_signal_args` for `paletteChanged` (which emits a `QPalette` that slots genuinely don't want). For no-arg signals, connect directly.

#### 2. ArchiveTab signal cleanup is a silent behavior change

The existing `_on_destroyed` only disconnected 2 of 6 signal connections (the two palette ones). `backup_finished_event`, `profile_changed_event`, and `backup_cancelled_event` were never disconnected — a pre-existing leak. The new `track_signal` approach properly disconnects all 6. This is a good fix but should be called out in the PR description since it changes behavior.

#### 3. `unlink_repository` has redundant archive deletion

```python
ArchiveModel.delete().where(ArchiveModel.repo_id == repo.id).execute()  # 1st delete
# ...
repo.delete_instance(recursive=True)  # 2nd delete — recursive=True also deletes archives
```

This was carried over from the original RepoTab code (which had a comment acknowledging it). Since the ViewModel is meant to be clean business logic, worth fixing rather than carrying the redundancy forward.

#### 4. Dual persistence paths

Both `BaseTab` and `RepoTabViewModel` expose `save_profile_attr()`. The PR uses the ViewModel consistently for RepoTab, while other tabs use `BaseTab.save_profile_attr` directly. Fine as a transitional state, but should be documented — when ViewModels are added for other tabs, `BaseTab.save_profile_attr`/`save_repo_attr` should eventually be removed or made private.

#### 5. `BackupProfileMixin` not removed from models.py

Acceptable for an incremental PR, but worth noting in the description that it's still present and used by untouched files (`diff_result.py`, `extract_dialog.py`, `main_window.py`, etc.).

#### 6. ViewModel adds indirection without proportional benefit (yet)

`RepoTabViewModel` methods are 1:1 mapped to UI actions doing `profile.X = val; profile.save()`. No validation, error handling, caching, or domain logic — it's code extraction, not concern separation. 77 new lines replacing ~20 inline lines in RepoTab.

The boilerplate reduction from `BaseTab` is the real win of this PR. If the ViewModel pattern is being established now for future use, that's valid but should be stated explicitly so reviewers understand the intent.

#### 7. Test gaps

- **test_base_tab.py**: Only tests `track_profile_change` cleanup. Missing coverage for `track_palette_change`, `track_backup_finished`, `_default_profile_provider` fallback, and `current_profile()` returning None.
- **test_view_models.py**: `unlink_repository` test doesn't verify behavior when a repo is shared by multiple profiles — the most dangerous operation in the ViewModel.
- No integration test with a real tab + ViewModel wired together.

#### 8. Minor: `_without_signal_args` should be `@staticmethod`

It doesn't reference `self`, only uses the `callback` parameter.

---

### What works well

- Clean cooperative inheritance with `BaseTab` first in MRO
- Profile injection via callable provider — testable and decoupled from widget hierarchy
- Template methods (`populate_from_profile`, `set_icons`) with no-op defaults
- Consistent application across all 10 tab/page classes
- `TypedDict` for `RepoListItem`
- Incidentally fixes a pre-existing signal leak in ArchiveTab

false200:
@m3nu Thanks for the detailed review,  I’ve applied the requested follow up changes. One question on the transitional architecture: would you prefer BaseTab.save_profile_attr()/save_repo_attr() to remain public for now until more ViewModels are introduced, or should I already make that direction more explicit in code in this PR?

m3nu:
## Follow-up after force-push

The updated revision addresses most of the earlier feedback — `@staticmethod`, the `_without_signal_args` wrapping consistency, the shared-repo check in `unlink_repository`, and significantly better tests. Two remaining items:

### 1. `repo_unlink_action` UI doesn't match the new ViewModel behavior

The old code always deleted the repo, so the UI could unconditionally remove it from the dropdown and show "Repository was Unlinked." Now that `unlink_repository` preserves shared repos, the UI path is wrong for that case:

```python
# repo_tab.py lines 300-301
self.repoSelector.removeItem(selected_repo_index)
self.repoSelector.setCurrentIndex(0)
```

When another profile still references the repo, `removeItem` removes it from the dropdown, then `populate_from_profile()` (called a few lines later) re-populates from DB and the repo reappears — a visible flicker. The message "Repository was Unlinked" is also misleading when the repo was only detached from the current profile, not deleted.

Suggestion: have `unlink_repository` return whether the repo was deleted, and branch the UI accordingly (skip `removeItem` when preserved, adjust the message text). Add a UI-level test for `repo_unlink_action` with a shared repo to cover the dropdown and signal behavior.

### 2. Document the intended persistence path

`BaseTab.save_profile_attr` and `RepoTabViewModel` both write to the same models. For RepoTab the answer is clear (use the ViewModel), but for future tabs there's no guidance. A brief comment above `save_profile_attr` would prevent drift — something like:

```python
# Convenience helper for simple pages. Tabs with a ViewModel should
# route writes through it instead.
```

m3nu:
One more architectural note — consider dropping `RepoTabViewModel` from this PR and shipping BaseTab alone.

BaseTab is the clear win: it solves the stated problem (signal boilerplate) with a net reduction in code. The ViewModel adds 150 lines (78 source + 72 tests) to replace ~20 lines of inline `profile.x = val; profile.save()` in RepoTab. Every method is a trivial proxy — no validation, no error handling, no multi-model coordination.

Looking at what else RepoTab does (`init_repo_stats`, `repo_change_passphrase_action`, `copy_URL_action`, `populate_from_profile`), it's all view-layer orchestration or read-only formatting. There isn't persistence logic left for the ViewModel to absorb. RepoTab is a settings form, not a complex workflow like ArchiveTab's job orchestration.

Removing it would also eliminate the two issues from my earlier comments — the dual persistence path question and the unlink UI/ViewModel mismatch — since both only exist because of the ViewModel layer. The shared-repo protection in `unlink_repository` is good logic, but it could live as a method on `RepoModel` rather than requiring a ViewModel class.

Without the ViewModel, this PR becomes a clean -30 net lines instead of +47, and you can introduce ViewModels in a later slice when a tab (likely ArchiveTab) has enough domain logic to justify the abstraction.

false200:
I've amended this PR accordingly and removed RepoTabViewModel so this change focuses on introducing BaseTab and reducing the signal boilerplate.

m3nu (approved):
Looks good after the latest revision.

BaseTab cleanly centralizes signal tracking and profile access with a net -22 lines of source code. The shared-repo unlink logic landed in the right place (`RepoModel.is_shared_with_other_profiles`), the UI branches correctly on delete vs detach, and the dropdown/message behavior is covered by a proper UI-level test. Single persistence path, no unnecessary abstractions.

Minor items that could be picked up in later slices but don't block this one:

- The `populate_from_profile(); track_profile_change()` two-liner repeats across several tabs — a `call_now=True` parameter on the track methods would trim it.
- The lambda glue for widget-to-model binding (`lambda new_val, attr='x': self.save_profile_attr(attr, new_val)`) repeats 9 times across schedule_page, shell_commands_page, and networks_page. A `bind_profile_attr(signal, attr)` helper on BaseTab would absorb most of it.
- `BackupProfileMixin` is still defined in models.py and used by untouched files (`diff_result.py`, `extract_dialog.py`, `main_window.py`) — worth noting in #2361 for a future slice.

CodeLoopdroid:
> Looks good after the latest revision.
> 
> BaseTab cleanly centralizes signal tracking and profile access with a net -22 lines of source code. The shared-repo unlink logic landed in the right place (`RepoModel.is_shared_with_other_profiles`), the UI branches correctly on delete vs detach, and the dropdown/message behavior is covered by a proper UI-level test. Single persistence path, no unnecessary abstractions.
> 
> Minor items that could be picked up in later slices but don't block this one:
> 
> * The `populate_from_profile(); track_profile_change()` two-liner repeats across several tabs — a `call_now=True` parameter on the track methods would trim it.
> * The lambda glue for widget-to-model binding (`lambda new_val, attr='x': self.save_profile_attr(attr, new_val)`) repeats 9 times across schedule_page, shell_commands_page, and networks_page. A `bind_profile_attr(signal, attr)` helper on BaseTab would absorb most of it.
> * `BackupProfileMixin` is still defined in models.py and used by untouched files (`diff_result.py`, `extract_dialog.py`, `main_window.py`) — worth noting in [Refactor views: Add BaseTab class, ViewModel layer, and reorganize dialogs #2361](https://github.com/borgbase/vorta/issues/2361) for a future slice.

Hi @m3nu,

Thanks again for the review and the suggestions in the PR, I really appreciate the feedback.

Since I’m contributing here while preparing for GSoC, I’d be happy to pick up some of the follow up items you mentioned (like the `call_now=True` improvement for the track methods, the `bind_profile_attr` helper or eventually removing `BackupProfileMixin`) as the next slices of #2361, if that sounds good.

Also, since this is listed as a GSoC idea, am wondering how you’d recommend preparing a good proposal for it.

Thanks!

PR #2466, opened 2026-03-21
Fix WiFi detection when wired connection is primary

### LINKED ISSUE #1775: Vorta ignores "Allowed Networks" value
### Description

I've set up a backup job that should only run if the laptop is on a certain network. Now it seems that when I tick the box of which network is allowed this setting is simply ignored and the backup runs anyway. 

One thing which might be the reason is that the laptop is **also** connected to a wired network connection via docking station. So something with the check of the **active** network does not quite work as it should.  

### Reproduction

- [ ] I was able to reproduce the issue.

### OS

Linux Ubuntu Mate 22.04

### Version of Vorta

0.8.12

Distribution package

### Version of Borg

1.2.4

### Logs

_No response_

CrazyTodd-one:
## Summary
Fixes #1775 — When a laptop is connected via both WiFi and a wired connection (e.g. docking station), the primary connection is wired. `get_current_wifi()` only checked the primary connection, so the WiFi SSID was never detected and network-based backup restrictions were silently bypassed.

**Changes:**
- `get_current_wifi()` now iterates all active NetworkManager connections instead of only the primary one
- Added `get_active_connections_paths()` to the DBus adapter, reading the `ActiveConnections` property
- Per-connection DBus failures are handled gracefully (logged and skipped, other connections still checked)

## Test plan
- [x] Added `test_get_current_wifi_with_wired_primary_and_wifi_active` — simulates the docking station scenario (wired primary + WiFi active), verifies WiFi SSID is found
- [x] Added `test_get_current_wifi_skips_failed_connection` — verifies that if one connection throws DBusException, remaining connections are still checked
- [x] Updated existing tests to use `get_active_connections_paths` mock
- [x] Fixed incorrect `802-11-ethernet` type string to `802-3-ethernet` in parametrized test


m3nu (approved):
Clean, focused fix for a long-standing issue (#1775). The approach of iterating all active connections instead of only the primary is correct and well-implemented.

Strengths:
- Reuses existing `read_dbus_property()` helper -- consistent with the adapter pattern
- Per-connection `DBusException` handling so one failure doesn't block checking others
- Two good new tests covering the wired+wifi scenario and per-connection failure resilience
- Fixes pre-existing `802-11-ethernet` → `802-3-ethernet` typo in test data
- API contract preserved (`get_current_wifi() -> str | None`)

Note: competing PR #2473 addresses the same issue but has significant problems (debug prints in production, deletes metered-network check, catches bare `Exception`, no tests, breaks ABC contract). Recommend closing it in favor of this one.

m3nu (approved):

PR #2469, opened 2026-03-22
feat: Delete key removes items in source and profile lists

### LINKED ISSUE #2447: Support using the Delete key to remove items in Profile and Source lists
#### The problem

Currently, removing items from the profile list and source lists requires using the mouse (either via the right-click context menu or '-' UI button). There is no support for the Delete keyboard shortcut. I'd love to use Del keyboard shortcut instead of using mouse.

#### Requested Solution
It would be great to add support for the `Delete` key so we can quickly remove selected items in those lists without touching the mouse.

Fikri-20:
Closes #2447

Added a Delete key shortcut to both the source files table and the profile selector list, wired to the existing remove/delete actions.

### DIFF
--- a/src/vorta/views/main_window.py
+++ b/src/vorta/views/main_window.py
@@ -110,7 +110,7 @@ class MainWindow(MainWindowBase, MainWindowUI):
         self.profileRenameButton.clicked.connect(self.profile_rename_action)
         self.profileExportButton.clicked.connect(self.profile_export_action)
         self.profileDeleteButton.clicked.connect(self.profile_delete_action)
-        QShortcut(QKeySequence(Qt.Key.Key_Delete), self.profileSelector).activated.connect(self.profile_delete_action)
+        QShortcut(QKeySequence.StandardKey.Delete, self.profileSelector).activated.connect(self.profile_delete_action)
         self.profileAddButton.addAction(self.tr("Create new profile"), self.profile_add_action)
         self.profileAddButton.addAction(self.tr("Import from file…"), self.profile_import_action)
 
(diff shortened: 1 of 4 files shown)


Fikri-20:
hey @m3nu , could you please review the PR and tell me if I need to tweak anything else?

Fikri-20:
Hi @m3nu, I'm applying for GSoC 2026 and have these PR ready for review. Could you please take a look?

m3nu:
## Review of PR #2469

Clean, focused PR. Reuses existing `profile_delete_action` and `source_remove` methods, so all validation (confirmation dialog, last-profile guard) is preserved. Tests cover both shortcuts. No new imports needed in production code.

### Issues

1. **macOS keyboard mismatch**: `Qt.Key.Key_Delete` maps to the **forward-delete** key, which doesn't exist on most Mac keyboards (only full-size ones). On macOS, users expect Backspace (Qt's `Key_Backspace`) for deletion. The shortcut will be effectively unusable on most Macs. Consider adding `Key_Backspace` as an additional shortcut, or using `QKeySequence.StandardKey.Delete` which Qt maps platform-appropriately.

2. **Inconsistent style between files**:
   - `main_window.py` creates the shortcut inline: `QShortcut(...).activated.connect(...)`
   - `source_tab.py` uses a two-line pattern with a local variable, matching the existing `shortcut_copy` pattern in that file.

   The one-liner works (Qt parent-child ownership keeps it alive), but the two-line pattern is clearer and consistent with the rest of source_tab.

3. **Exclusion tab not covered**: The exclusion list also has a remove action. Not blocking, but worth considering for consistency.

Fikri-20:
Hi @m3nu, thanks for the review! I've updated the PR to use `QKeySequence.StandardKey.Delete` for cross-platform compatibility as suggested. This should now work correctly on macOS laptops as well.

Let me know if any other tweaks are needed.

Fikri-20:
@m3nu would like to know if anything I should add/tweak for this PR.

Fikri-20:
Hi @m3nu , 
I appreciate your effort to review the PR, but should I know what’s the matter with the PR to be closed? What should I consider to make this PR mergable?

PR #2470, opened 2026-03-23
fix: guard darwin network test import behind platform check

Denay2468:
## Problem
Running `uv run pytest tests/` on Windows fails at collection time:

    ModuleNotFoundError: No module named 'CoreWLAN'

`pytestmark` skips test execution but cannot prevent the top-level
import from running on Windows where CoreWLAN does not exist.
This crashes the entire test suite before any test runs.

## Fix
Wrapped the import in `if sys.platform == 'darwin'` so it never
executes on Windows or Linux.

## Before / After
Before: ERROR collecting tests/network_manager/test_darwin.py
After:  11 darwin tests skip cleanly on Windows

## Tested on
- Windows 11 ✅

### DIFF
--- a/tests/network_manager/test_darwin.py
+++ b/tests/network_manager/test_darwin.py
@@ -1,76 +1,76 @@
 import sys
 from unittest.mock import MagicMock
- 
+
 import pytest
- 
+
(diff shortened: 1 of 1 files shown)


Denay2468:
All checks passing. Fixed CRLF -> LF line endings. Ready for review!

PR #2471, opened 2026-03-23
fix: include ExclusionModel in profile export/import

### LINKED ISSUE #2414: excludes list is not added to the exported settings
### Description

I encountered a bug.

I wanted to transfer a Vorta Profile to another device. I used the "Aktiviertes Profil exportieren" (export activated profile) button to create a json file. I also selected to include the password.

Looking into the json file I found that my list of excludes (configured in "Ausschlüsse verwalten" = manage excludes, first tab  "Benutzerdefiniert" (Custom))  list (about 10 entries) can not be found in the exported profile.

I think this should be included in the profile config file.

Thanks a lot.
Matthias

### Reproduction

- [x] I tried to reproduce the issue.
- [x] I was able to reproduce the issue.

### OS

xubuntu 24.04

### Version of Vorta

0.11.3

Flatpak

### Version of Borg

1.4.3

### Logs

```pytb

```

Denay2468:
## Problem
Closes #2414

When exporting a profile to JSON, the excludes list (custom patterns
and presets configured in "Manage Exclusions") was silently missing
from the exported file. On import, exclusions were never restored,
causing users to lose all their exclude rules when transferring
profiles between devices.

## Root Cause
ProfileExport.from_db() serializes SourceFileModel and
WifiSettingModel but never serializes ExclusionModel.
Similarly, to_db() restores sources and wifi settings but
skips exclusions entirely.

## Fix
- Added ExclusionModel to imports in profile_export.py
- Serialize exclusions in from_db() following the same pattern
  as SourceFileModel
- Restore exclusions in to_db() with profile id reassignment
- Clean up ExclusionModel key at end of to_db()

## Tested
- All 9 existing export/import unit tests pass ✅
- Windows 11 ✅

### DIFF
--- a/src/vorta/profile_export.py
+++ b/src/vorta/profile_export.py
@@ -1,9 +1,9 @@
 import datetime
 import json
 from json import JSONDecodeError
- 
+
 from playhouse.shortcuts import dict_to_model, model_to_dict
- 
+
 from vorta.keyring.abc import VortaKeyring
 from vorta.store.connection import DB, SCHEMA_VERSION, init_db
 from vorta.store.models import (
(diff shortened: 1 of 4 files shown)


m3nu:
(PR was merged after fixing major white space issues, rest of the content was sound)

Issue #1631, opened 2023-03-08
Set supported schemes when opening file dialog.


real-yfprojects:
You can open `smb://` and other URLs when adding a repository although borg only supports normal paths (file scheme) and the ssh scheme. This is because vorta is missing to set the supported schemes attribute.

See https://doc.qt.io/qt-5/qfiledialog.html#supportedSchemes-prop

_Originally posted by @ThomasWaldmann in https://github.com/borgbase/vorta/issues/1628#issuecomment-1459041597_

i1sm3ky:
@real-yfprojects @ThomasWaldmann Would `dialog.setSupportedSchemes(["file", "ssh"])` work?

real-yfprojects:
I think so.

ThomasWaldmann:
Not sure whether the file dialog will be able to let you "select" a ssh: url somehow, but we can try.

What it for sure won't do, is the special ssh: path styles borg uses to indicate current directory and home directory on the server, usually ssh:// urls always give absolute paths only.

real-yfprojects:
> Not sure whether the file dialog will be able to let you "select" a ssh: url somehow, but we can try.

I was thinking that there might be a file selector somewhere out there that allows just that, so why not tell it that vorta supports `ssh://`.

> What it for sure won't do, is the special ssh: path styles borg uses to indicate current directory and home directory on the server, usually ssh:// urls always give absolute paths only.

Are normal ssh: paths supported by borg?

ThomasWaldmann:
`ssh://user@host:port/some/abs/path` is the usual and supported form.

`ssh://user@host:port/./some/rel/path` and `ssh://user@host:port/~/some/home/path` is supported by borg, but likely nothing else.

i1sm3ky:
So should we just allow the local path to be added using the dialog box and leave the ssh to be filled manually?

i1sm3ky:
@real-yfprojects so I just add `dialog.setSupportedSchemes(["file", "ssh"])` and merge it right?

ThomasWaldmann:
@i1sm3ky i guess this needs trying interactively. try it before any change, check how the dialogue behaves and what it offers (esp. considering non-file: stuff), then change the schemes and try again how it behaves.

this is platform dependent, so try on all platforms you have access to.

i1sm3ky:
[...]

PR #2476, opened 2026-03-24
fix: restrict folder dialog to file:// URLs (#1631)

OsamaGharibElwaly:
## Summary

Sets \QFileDialog.setSupportedSchemes(['file'])\ in \choose_file_dialog()\ so the add-repo, mount, and extract folder pickers do not offer unsupported schemes (e.g. \smb://\, \sftp://\) that can freeze or misbehave on some platforms.

## Testing

- \pytest tests/unit/test_utils.py::test_choose_file_dialog_sets_file_scheme_only\

Related to #1631


m3nu (approved):
Minimal, correct fix for #1631. `QFileDialog.setSupportedSchemes(['file'])` is the documented Qt mechanism and all Vorta file dialogs only need local filesystem access.

- Placement after `setOption(ShowDirsOnly)` is fine -- order doesn't matter for these independent properties
- Applied unconditionally to both folder and file pickers, which is correct
- Test verifies `setSupportedSchemes` is called with the right argument

No issues found. Ready to merge once CI passes.

PR #2517, opened 2026-06-06
add ArchiveTableModel (QAbstractTableModel) spike

ebuzerdrmz44:
A **small spike** of the Archive table as a `QAbstractTableModel`, opened as a **draft** to confirm the shape before going deep on the view rewiring (as offered in the Phase 5 #2361 thread).

  Adds, and **nothing else** (the live Archive tab is untouched):
  - `src/vorta/views/partials/archive_table_model.py` ,`ArchiveTableModel`, mirroring the existing `EventLogTableModel`. Owns the per-row presentation currently inline in `ArchiveTab.populate_from_profile`: time/size/duration formatting, trigger icon + tooltip, raw **sort-role** keys, an editable **name** column, plus `set_rows(...)` and `archive_at(row)`.
  - `tests/unit/test_archive_table_model.py`  15 isolated unit tests (content + ordering, sort role, fixed/dynamic units, editable flag, setData/dataChanged, icons/tooltips,headers).

  Deliberately deferred to the wiring PR:
  - Swap `QTableWidget` → `QTableView` and call `model.set_rows(...)` from `populate_from_profile`.
  - **Sorting:** an `ArchiveSortProxyModel(QSortFilterProxyModel)` overriding `lessThan()` (matching `treemodel.py`), reading the model's `SortRole`.
  - Re-routing the ~10 `archiveTable.item(row, col)` reads to `model.archive_at(row)`.
  - Migrating in-place rename + selection/double-click/delete handling.

  Conventions: no DB reads in the model (the `enable_fixed_units` setting is injected by the view via `set_rows(..., use_fixed_units=...)`); translation contexts reuse existing catalog entries (`Form` headers, `ArchiveTab` tooltips).

  ### Related Issue
  
  Phase 5 of the views refactor (#2361)  the redirect to `QAbstractTableModel` + `views/partials/`.

  ### Motivation and Context

  Per the Phase 5 redirect, tabular views (Archive, Log) should become `QAbstractTableModel` + `QTableView` rather than ViewModels. **_This spike extracts the Archive table's presentation logic into a self-contained, idiomatic model so the shape can be reviewed before the larger `archive_tab.py` migration_**.

  ### How Has This Been Tested?
  
  `uv run nox -r -- tests/unit/test_archive_table_model.py` . The model is exercised directly in isolation (no widget). No behavior change to the live app, since the model isn't wired in yet.
  
  ### Types of changes

  ### Checklist:

### DIFF
--- a/src/vorta/views/partials/archive_table_model.py
+++ b/src/vorta/views/partials/archive_table_model.py
@@ -8,11 +8,11 @@ from __future__ import annotations
 from datetime import timedelta
 from typing import Any, Dict, List, Optional
 
-from PyQt6.QtCore import QAbstractTableModel, QModelIndex, Qt
+from PyQt6.QtCore import QAbstractTableModel, QModelIndex, QObject, Qt
 
 from vorta.i18n import trans_late, translate
 from vorta.store.models import ArchiveModel
-from vorta.utils import find_best_unit_for_sizes, pretty_bytes
+from vorta.utils import find_best_unit_for_sizes, pretty_bytes, uses_dark_mode
 from vorta.views.utils import get_colored_icon
 
 #: Decimal digits shown in the size column.
@@ -32,6 +32,8 @@ class ArchiveTableModel(QAbstractTableModel):
 
     #: Role returning native, comparable values for sorting via a proxy model.
     SortRole = Qt.ItemDataRole.UserRole
+    #: Role returning the backing ArchiveModel; auto-mapped through the sort proxy.
+    ArchiveRole = Qt.ItemDataRole.UserRole + 1
 
     _HEADERS = (
         trans_late('Form', 'Date'),
@@ -39,15 +41,19 @@ class ArchiveTableModel(QAbstractTableModel):
         trans_late('Form', 'Duration'),
         trans_late('Form', 'Mount Point'),
         trans_late('Form', 'Name'),
-        '',  # trigger icon column has no header label
+        trans_late('Form', 'Trigger'),
     )
 
-    def __init__(self, parent: Optional[Any] = None):
+    _TRIGGER_ICONS = {'scheduled': 'clock-o', 'user': 'user'}
+
+    def __init__(self, parent: Optional[QObject] = None):
         """Init."""
         super().__init__(parent)
         self._rows: List[ArchiveModel] = []
         self._mount_points: Dict[str, str] = {}
         self._fixed_unit: Optional[int] = None
+        self._icon_cache: Dict[str, Any] = {}  # themed icons; invalidated on dark-mode switch
+        self._icon_cache_dark: Optional[bool] = None
 
     def set_rows(
         self,
@@ -98,6 +104,8 @@ class ArchiveTableModel(QAbstractTableModel):
             return row.name  # pre-fill the rename editor with the current name
         if role == self.SortRole:
             return self._sort_data(row, column)
+        if role == self.ArchiveRole:
+            return row  # proxy-safe accessor: the proxy forwards data() through mapToSource
         if role == Qt.ItemDataRole.DecorationRole and column == self.COL_TRIGGER:
             return self._trigger_icon(row)
         if role == Qt.ItemDataRole.ToolTipRole and column == self.COL_TRIGGER:
@@ -136,11 +144,16 @@ class ArchiveTableModel(QAbstractTableModel):
         return None
 
     def _trigger_icon(self, row: ArchiveModel) -> Any:
-        if row.trigger == 'scheduled':
-            return get_colored_icon('clock-o')
-        if row.trigger == 'user':
-            return get_colored_icon('user')
-        return None
+        icon_name = self._TRIGGER_ICONS.get(row.trigger)
+        if icon_name is None:
+            return None
+        dark = uses_dark_mode()
+        if dark != self._icon_cache_dark:  # theme switched: themed icons are stale
+            self._icon_cache.clear()
+            self._icon_cache_dark = dark
+        if icon_name not in self._icon_cache:
+            self._icon_cache[icon_name] = get_colored_icon(icon_name)
+        return self._icon_cache[icon_name]
 
     def _trigger_tooltip(self, row: ArchiveModel) -> Any:
         if row.trigger == 'scheduled':
--- a/src/vorta/views/partials/archive_table_model.py
+++ b/src/vorta/views/partials/archive_table_model.py
@@ -1,0 +1,176 @@
+"""
+Qt table model exposing `ArchiveModel` rows to the ArchiveTab `QTableView`.
+
+"""
+
+from __future__ import annotations
+
+from datetime import timedelta
+from typing import Any, Dict, List, Optional
+
+from PyQt6.QtCore import QAbstractTableModel, QModelIndex, Qt
+
+from vorta.i18n import trans_late, translate
+from vorta.store.models import ArchiveModel
+from vorta.utils import find_best_unit_for_sizes, pretty_bytes
+from vorta.views.utils import get_colored_icon
+
+#: Decimal digits shown in the size column.
+SIZE_DECIMAL_DIGITS = 1
+
+
+class ArchiveTableModel(QAbstractTableModel):
+    """Table model for backup archive rows; the name column is editable for in-place rename."""
+
+    # Column indices in render order (matching the legacy QTableWidget layout).
+    COL_TIME = 0
+    COL_SIZE = 1
+    COL_DURATION = 2
+    COL_MOUNT = 3
+    COL_NAME = 4
+    COL_TRIGGER = 5
+
+    #: Role returning native, comparable values for sorting via a proxy model.
+    SortRole = Qt.ItemDataRole.UserRole
+
+    _HEADERS = (
+        trans_late('Form', 'Date'),
+        trans_late('Form', 'Size'),
+        trans_late('Form', 'Duration'),
+        trans_late('Form', 'Mount Point'),
+        trans_late('Form', 'Name'),
+        '',  # trigger icon column has no header label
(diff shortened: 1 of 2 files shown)


ebuzerdrmz44:
Quick question on sorting before I wire this in: custom `SortRole` (current approach) or override `lessThan()` in a proxy like `treemodel.py`? Same question as #2515 . whatever you prefer I'll apply across Archive/Source.

ebuzerdrmz44:
m3nu small heads-up: I've got finals the next few days, back at **_full pace_** **June 11**.After that I will work through the Archive and Source tabs ( and any other tabs that need it) per your reviews here and on the others. #2514 and #2515 are reshaped and ready for re-review whenever you get a chance. No rush , thanks!

src/vorta/views/partials/archive_table_model.py
```diff
@@ -0,0 +1,176 @@
...
+        self.beginResetModel()
+        self._rows = list(rows)
+        self._mount_points = dict(mount_points or {})
+        if use_fixed_units:
+            self._fixed_unit = find_best_unit_for_sizes((r.size for r in self._rows), precision=SIZE_DECIMAL_DIGITS)
+        else:
+            self._fixed_unit = None
+        self.endResetModel()
+
+    def archive_at(self, row: int) -> Optional[ArchiveModel]:
```
m3nu, line 78: 🔴 **[P1 — data-safety] `archive_at(row)` takes a *source* row, but the wiring will feed it *proxy* rows.**

This method is correct in isolation, but the PR's deferred recipe ("re-route the ~10 `archiveTable.item(row, col)` reads to `model.archive_at(row)`") will be wrong once `ArchiveSortProxyModel` is in place: `selectionModel().selectedRows()` returns **proxy** indices, and `archive_at(proxy_row)` indexes `self._rows` in **source** order. After any sort, that resolves to the wrong archive for rename (`archive_tab.py:466,496,591`), delete (`701-702`), and diff (`757-758`) — i.e. acting on the wrong backup.

The repo already establishes the rule that proxy indices must be mapped first — see `diff_result.py:180` and `extract.py:174` (`index = self.sortproxy.mapToSource(index)`).

**Recommended:** expose the archive through a custom data role rather than a raw-row accessor, e.g. `ArchiveRole = Qt.ItemDataRole.UserRole + 1` returned from `data()`. Callers then use `index.data(ArchiveRole)` on the proxy index, and `QSortFilterProxyModel` auto-forwards through `mapToSource` — there's no raw row to get wrong, and it's symmetric with the existing `SortRole`. (If you keep `archive_at`, every one of the ~10 call sites must `proxy.mapToSource()` first, and rename it / document it as source-indexed.)

src/vorta/views/partials/archive_table_model.py
```diff
@@ -0,0 +1,176 @@
...
+            return row.duration or 0
+        if column == self.COL_MOUNT:
+            return self._mount_points.get(row.name, '')
+        if column == self.COL_NAME:
+            return row.name
+        if column == self.COL_TRIGGER:
+            return row.trigger or ''
+        return None
+
+    def _trigger_icon(self, row: ArchiveModel) -> Any:
```
m3nu, line 146: 🟡 **[P2 — perf regression] Trigger icon is re-rendered from disk on every paint.**

`get_colored_icon` (`views/utils.py:10`) does `open()` + read + `QImage.fromData` (SVG parse) + `scaledToHeight(128)` + wrap, with **no caching**. `data()` calls this on every `DecorationRole` request, which Qt fires per repaint/scroll for each visible trigger cell — and it renders a 128px image for a cell-sized icon. This is a regression from the legacy path, which built the icon **once per populate** and stored it in the `QTableWidgetItem`.

**Recommended:** cache the two icons (`clock-o`, `user`) on the model, keyed on dark-mode state, and rebuild them on `QEvent.PaletteChange` (the icons are theme-colored, so a plain permanent cache would go stale on a light/dark switch).

src/vorta/views/partials/archive_table_model.py
```diff
@@ -0,0 +1,176 @@
...
+    #: Role returning native, comparable values for sorting via a proxy model.
+    SortRole = Qt.ItemDataRole.UserRole
+
+    _HEADERS = (
+        trans_late('Form', 'Date'),
+        trans_late('Form', 'Size'),
+        trans_late('Form', 'Duration'),
+        trans_late('Form', 'Mount Point'),
+        trans_late('Form', 'Name'),
+        '',  # trigger icon column has no header label
```
m3nu: 🟡 **[P3 — behavior] The "Trigger" column header is silently dropped.** The live `.ui` defines a `Trigger` header for column 5 (`archive_tab.ui:175`); `_HEADERS[5] = ''` removes it. That's a real UX change, not a no-op, and it diverges from the PR's "reuse existing catalog entries" claim. Either keep `trans_late('Form', 'Trigger')`, or make the icon-only-no-header decision explicit in the description.

src/vorta/views/partials/archive_table_model.py
```diff
@@ -0,0 +1,176 @@
...
+
+from PyQt6.QtCore import QAbstractTableModel, QModelIndex, Qt
+
+from vorta.i18n import trans_late, translate
+from vorta.store.models import ArchiveModel
+from vorta.utils import find_best_unit_for_sizes, pretty_bytes
+from vorta.views.utils import get_colored_icon
+
+#: Decimal digits shown in the size column.
+SIZE_DECIMAL_DIGITS = 1
```
m3nu, line 19: 🔵 **[P4 — maintainability]** `SIZE_DECIMAL_DIGITS` is now defined here *and* at `archive_tab.py:56`. On wiring, have one own it and the other import it so they can't drift.

src/vorta/views/partials/archive_table_model.py
```diff
@@ -0,0 +1,176 @@
...
+        row = self._rows[index.row()]
+        column = index.column()
+
+        if role == Qt.ItemDataRole.DisplayRole:
+            return self._display_data(row, column)
+        if role == Qt.ItemDataRole.EditRole and column == self.COL_NAME:
+            return row.name  # pre-fill the rename editor with the current name
+        if role == self.SortRole:
+            return self._sort_data(row, column)
+        if role == Qt.ItemDataRole.DecorationRole and column == self.COL_TRIGGER:
```
m3nu, line 109: 🔵 **[P5 — parity]** The legacy `user`-trigger item sets `AlignRight` (`archive_tab.py:339`); the model exposes no `TextAlignmentRole`. Negligible since the cell is icon-only, but worth a line on the wiring parity checklist so it's a conscious drop rather than an accidental one.

src/vorta/views/partials/archive_table_model.py
```diff
@@ -0,0 +1,176 @@
...
+    _HEADERS = (
+        trans_late('Form', 'Date'),
+        trans_late('Form', 'Size'),
+        trans_late('Form', 'Duration'),
+        trans_late('Form', 'Mount Point'),
+        trans_late('Form', 'Name'),
+        '',  # trigger icon column has no header label
+    )
+
+    def __init__(self, parent: Optional[Any] = None):
```
m3nu: 🔵 **[P6 — nit]** `parent: Optional[Any]` → prefer `Optional[QObject]`, matching the typed style in `treemodel.py` (same nit I left on #2515).

m3nu (commented):
## 🔍 Review — ArchiveTableModel spike

The shape is right and this is the correct foundation for the wiring PR. I diffed the model against the live `archive_tab.py:295-340` formatting and fidelity is good; the `SortRole` design is exactly what we settled on in #2515 and even fixes a latent bug (durations >24h render as `'1 day, …'` and sort wrong as text today — raw-seconds sort keys order them correctly). Tests (15, widget-free) are the bar I'd want for the other tables.

Six findings inline, two of which are important to resolve **before** the wiring is written against this API:

- 🔴 **P1 (data-safety):** `archive_at(row)` is source-indexed; once the planned sort proxy is added it will target the **wrong archive** for rename/mount/delete. Use a custom data role instead.
- 🟡 **P2 (perf regression):** the trigger icon is re-rendered from disk on every paint.
- 🟡 P3: the "Trigger" column header is silently dropped.
- 🔵 P4–P6: a duplicated constant, a missed alignment parity, and a typing nit.

Not approving since this is a draft — leaving as a comment. Shape-wise: 👍, address P1/P2 in the wiring PR and the recipe is solid.

ebuzerdrmz44:
Thanks for the review m3nu , all six findings are addressed . 

  - P1 : added `ArchiveRole = Qt.ItemDataRole.UserRole + 1` returned from `data()`. Selection is resolved via `index.data(ArchiveRole)` on proxy indices, so `QSortFilterProxyModel` auto-forwards through `mapToSource` 
  - P2 : trigger icons cached on the model, keyed on dark-mode, invalidated on theme switch.
  - P3: restored the Trigger header.
  - P6: parent: `Optional[QObject]`.

P4/P5 are handled in the wiring, as you suggested: it single-sources SIZE_DECIMAL_DIGITS from the model (P4), and the AlignRight drop is intentional , the Trigger cell is icon-only and IconDelegate centers it (P5).

The wiring is implemented and green (Phase 4 part 1: QTableView + ArchiveTableModel fully wired into archive_tab.py. _**I'll open it as a follow-up PR once this spike lands**_. **Phase 4 part 2** (peeling rename/diff/context-menu off the orchestrator) will be a separate slice.

ebuzerdrmz44:
Hi @m3nu , I've addressed all six findings here , so this is ready for re-review whenever you have a moment. #2519 (the wiring) is stacked on top, so it unblocks once this lands. and I've also got the next piece ready to go,  happy to open it whenever works for you, Thanks!

m3nu:
Sounds good! On it.

PR #2435, opened 2026-03-15
Fix #1080: prevent crash when removing source during size recalculation

ebuzerdrmz44:
### Description
When a source is removed while its size is being recalculated, disconnect the
running thread's signal and add a `DoesNotExist` guard in `set_path_info` for cases
where the signal was already queued before disconnection.

### Related Issue
#1080

### Motivation and Context
Removing a source while its size is being recalculated causes the app to crash
with `QThread: Destroyed while thread is still running`. This happens because
the thread completes and tries to call `set_path_info` after the source has
already been deleted.

### How Has This Been Tested?
Tested manually:
1. Added a large folder in Source tab
2. Clicked Recalculate
3. Removed the source while calculating
4. Without fix: app crashes with "QThread: Destroyed while thread is still running"
5. With fix: app continues normally

Unit tests: 189 passed, 6 skipped, 0 failed.

### Screenshots (if appropriate):
N/A

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist:
- [x] I have read the CONTRIBUTING guide.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #1080, opened 2021-10-18
Missing UI element may trigger crash when recalculating folder sizes


m3nu:
**Describe the bug**
Vorta may crash in rare situations, while calculating source folder sizes. First reported [here](https://github.com/borgbase/vorta/commit/9bad152a8006710f42e4bf95e7fca96a28b7ffa3#commitcomment-58205419)

**To Reproduce**
Steps to reproduce the behavior: (not verified)
1. Go to source tab
2. Add folder
3. Do "Recalculate"
4. Remove folder right away
5. Maybe see error

**Environment (please complete the following information):**
- OS: macOS
- Vorta version: master branch, 0.8, Oct 2021
- Installed from: Git

Vorta and Borg versions can be found in Main Window > Misc Tab.

**Additional context**
Updated on my system and got this when trying to add a new folder and recalculating. I think it might be related to this
```
021-10-18 14:35:37,480 - vorta.borg.borg_job - INFO - Running command /usr/bin/borg --version
2021-10-18 14:36:15,675 - root - CRITICAL - Uncaught exception, file a report at https://github.com/borgbase/vorta/issues/new
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/vorta/views/source_tab.py", line 156, in sources_update
    self.update_path_info(row)  # Update data for each entry
  File "/usr/lib/python3.9/site-packages/vorta/views/source_tab.py", line 100, in update_path_info
    self.sourceFilesWidget.item(index_row, SourceColumn.Type).setText(self.tr("Calculating..."))
AttributeError: 'NoneType' object has no attribute 'setText'

```

real-yfprojects:
It shouldn't be possible that the error given relates to a source being deleted since the method is called immediately after clicking the button and as far as I know before any other window events are processed.
https://github.com/borgbase/vorta/blob/461ea0521092091bfb03b8293c6bb7e13a2c9ebd/src/vorta/views/source_tab.py#L95

Deleting a source while calculating shouldn't be a problem as well. The result are handled by `set_path_info` which searches for the source path in the source table. In case the source isn't found any more the for loop changing the entries won't process any lines.
https://github.com/borgbase/vorta/blob/461ea0521092091bfb03b8293c6bb7e13a2c9ebd/src/vorta/views/source_tab.py#L66-L72

delfino-nunez:
Can't reproduce it on Vorta 0.8.3.
It works for me on Vorta 0.8.3 from Github,

Steps

[...]

PR #2523, opened 2026-06-23
refactor(source): convert source files table to QTableView + QAbstractTableModel

ebuzerdrmz44:
### Description

Migrates the Source tab's file list from an item-based `QTableWidget` to a `QTableView` backed by a dedicated `QAbstractTableModel` (`SourceFilesModel` in `views/partials/`).

`source_tab.py` shrinks substantially: the imperative "disable sort → setItem loop → re-enable sort" choreography in `add_source_to_table`, `set_path_info`, `update_path_info`, and `populate_from_profile` is replaced by model calls. Per-row size/count recalculation now updates the backing `SourceFileModel` in place and emits `dataChanged`, instead of poking `QTableWidgetItem`s by row index.

### Related Issue

Part of #2361 (finish the views refactor / Phase 5 table-model migration).

### Motivation and Context

`QTableWidget` couples data and presentation and forces row-index bookkeeping, which is exactly what caused the historical mid-recalculation crash (#1080 / #2435). Moving to `QTableView` + `QAbstractTableModel` makes the model the single source of truth and lets sort/selection compose through a `QSortFilterProxyModel`.

Key design points:

- **Numeric-safe sorting via a shared `SortProxyModel`.** The model exposes raw comparable values (`dir_size` as int, file-count as int) under a `SortRole` (`Qt.UserRole`); `SortProxyModel.lessThan` compares those keys in Python so sizes ≥ 2 GiB don't hit Qt's C++ 32-bit `int` truncation. The old `SizeItem`/`FilesCount` `__lt__` hacks are gone.
- **Selection via a `SourceRole` (`UserRole + 1`)**, returned from `data()`. Callers resolve the backing object with `index.data(SourceRole)` on *proxy* indices, which `QSortFilterProxyModel` auto-forwards through `mapToSource`.
- **Themed path icons are cached** on the model, keyed on dark-mode state and invalidated on theme switch, rather than re-rendered from disk on every paint.
- **Path keying** for recalculation results: `set_path_info` matches on `dir` and returns `None` when the source was removed mid-calculation, so the caller skips persistence.
- **i18n preserved**: headers keep the `.ui`'s `Form` context (and "Calculating…" its `SourceTab` context), so existing catalog entries are reused.

### Rebased on #2519 — shared sort proxy + `SizeItem` removed

#2519 (Archive tab, Phase 4 pt 1) merged first, so this PR is rebased onto it and folds in the cleanup that merge triggered:

- **Extracted one shared sort proxy** into `views/partials/sort_proxy.py`: a single generic `SortProxyModel` keying off `Qt.UserRole`, now used by both the Archive and Source tables. This removes the near-duplicate per-model proxy classes (`ArchiveSortProxyModel` and this PR's original `SortProxyModel`) before the remaining tab migrations spawn a third copy. It also fixes the both-`None` strict-weak-ordering nit deferred from #2519 (`if lv is None: return rv is not None`).
- **Deleted the now-dead `SizeItem`** class and its orphaned imports (`QTableWidgetItem`, `sort_sizes`) from `source_tab.py` — `archive_tab.py` no longer references it after #2519.
- **Dropped a dead `SIZE_DECIMAL_DIGITS` import** left in `archive_tab.py` by #2519.
- Moved `updateThreads` from a class attribute to an instance attribute so worker lists aren't shared across `SourceTab` instances.

### How Has This Been Tested?

- New `tests/unit/test_source_files_table_model.py`, including a > 2 GiB size-sort regression test that runs through `SortProxyModel`.
- Existing `tests/unit/test_source.py` updated; `tests/unit/test_archive_table_model.py` switched to the shared proxy.

### Types of changes
- [x] Refactor (non-breaking change; no user-facing behavior change)

### Checklist:

### DIFF
--- a/src/vorta/views/partials/source_files_table_model.py
+++ b/src/vorta/views/partials/source_files_table_model.py
@@ -1,0 +1,172 @@
+"""Qt table model exposing `SourceFileModel` rows to the SourceTab `QTableView`."""
+
+from __future__ import annotations
+
+from typing import Any, Dict, List, Optional, Set
+
+from PyQt6.QtCore import QAbstractTableModel, QModelIndex, QObject, Qt
+
+from vorta.i18n import trans_late, translate
+from vorta.store.models import SourceFileModel
+from vorta.utils import pretty_bytes, uses_dark_mode
+from vorta.views.utils import get_colored_icon
+
+
+class SourceFilesModel(QAbstractTableModel):
+    """Table model for backup source rows; rows update in place during size recalculation."""
+
+    COL_PATH = 0
+    COL_SIZE = 1
+    COL_FILES = 2
+
+    #: Role returning native, comparable values for sorting via a proxy model.
+    SortRole = Qt.ItemDataRole.UserRole
+    #: Role returning the backing SourceFileModel; auto-mapped through the sort proxy.
+    SourceRole = Qt.ItemDataRole.UserRole + 1
+
+    _HEADERS = (
+        trans_late('Form', 'Path'),
+        trans_late('Form', 'Size'),
+        trans_late('Form', 'File Count'),
+    )
+
+    _CALCULATING = trans_late('SourceTab', 'Calculating…')
+
+    def __init__(self, parent: Optional[QObject] = None):
+        """Init."""
+        super().__init__(parent)
+        self._rows: List[SourceFileModel] = []
+        self._calculating: Set[str] = set()
+        self._icon_cache: Dict[str, Any] = {}  # themed icons; invalidated on dark-mode switch
+        self._icon_cache_dark: Optional[bool] = None
+
+    def set_rows(self, rows: List[SourceFileModel]) -> None:
+        """Replace the model contents and notify attached views."""
+        self.beginResetModel()
+        self._rows = list(rows)
+        self._calculating.clear()
+        self.endResetModel()
+
+    def add_source(self, source: SourceFileModel) -> int:
+        """Append ``source`` as a new row and return its row index."""
+        row = len(self._rows)
+        self.beginInsertRows(QModelIndex(), row, row)
+        self._rows.append(source)
+        self.endInsertRows()
+        return row
+
+    def source_at(self, row: int) -> Optional[SourceFileModel]:
+        """Return the `SourceFileModel` backing ``row``, or None if out of range."""
+        if 0 <= row < len(self._rows):
+            return self._rows[row]
+        return None
+
+    def mark_calculating(self, path: str) -> None:
+        """Show ``Calculating…`` for the row matching ``path`` until results arrive."""
+        self._calculating.add(path)
+        row = self._row_for_path(path)
+        if row is not None:
+            self.dataChanged.emit(self.index(row, self.COL_SIZE), self.index(row, self.COL_FILES))
+
+    def set_path_info(self, path: str, data_size: int, files_count: int, is_dir: bool) -> Optional[SourceFileModel]:
+        """Apply recalculated size/count to the row matching ``path`` and return it.
+
+        Keyed on ``path`` rather than a stored row index; returns None when no row matches
+        (e.g. the source was removed mid-calculation, #1080 / #2435) so the caller skips persistence.
+        """
+        self._calculating.discard(path)
+        row = self._row_for_path(path)
+        if row is None:
+            return None
+        source = self._rows[row]
+        source.dir_size = data_size
+        source.dir_files_count = files_count
+        source.path_isdir = is_dir
+        self.dataChanged.emit(self.index(row, self.COL_PATH), self.index(row, self.COL_FILES))
+        return source
+
+    def _row_for_path(self, path: str) -> Optional[int]:
+        for row, source in enumerate(self._rows):
+            if source.dir == path:
+                return row
+        return None
+
+    def rowCount(self, parent: QModelIndex = QModelIndex()) -> int:
+        if parent.isValid():
+            return 0
+        return len(self._rows)
+
+    def columnCount(self, parent: QModelIndex = QModelIndex()) -> int:
+        if parent.isValid():
+            return 0
+        return len(self._HEADERS)
+
+    def data(self, index: QModelIndex, role: int = Qt.ItemDataRole.DisplayRole) -> Any:
+        if not index.isValid():
+            return None
+
+        source = self._rows[index.row()]
+        column = index.column()
+
+        if role == Qt.ItemDataRole.DisplayRole:
+            return self._display_data(source, column)
+        if role == self.SortRole:
+            return self._sort_data(source, column)
+        if role == self.SourceRole:
+            return source  # proxy-safe accessor: the proxy forwards data() through mapToSource
+        if role == Qt.ItemDataRole.ToolTipRole and column == self.COL_PATH:
+            return source.dir
+        if role == Qt.ItemDataRole.DecorationRole and column == self.COL_PATH:
+            return self._path_icon(source)
+        if role == Qt.ItemDataRole.TextAlignmentRole and column == self.COL_SIZE:
+            return Qt.AlignmentFlag.AlignRight | Qt.AlignmentFlag.AlignVCenter
+        return None
+
+    def _display_data(self, source: SourceFileModel, column: int) -> Any:
+        if column == self.COL_PATH:
+            return source.dir
+        if column == self.COL_SIZE:
+            if source.dir in self._calculating:
+                return translate('SourceTab', self._CALCULATING)
+            if source.dir_size > -1:
+                return pretty_bytes(source.dir_size)
+            return ''
+        if column == self.COL_FILES:
+            if source.dir in self._calculating:
+                return translate('SourceTab', self._CALCULATING)
+            if source.path_isdir and source.dir_files_count > -1:
+                return str(source.dir_files_count)
+            return ''
+        return None
+
+    def _sort_data(self, source: SourceFileModel, column: int) -> Any:
+        """Native comparable value used by the sort proxy."""
+        if column == self.COL_PATH:
+            return source.dir
+        if column == self.COL_SIZE:
+            return source.dir_size
+        if column == self.COL_FILES:
+            return source.dir_files_count if source.path_isdir else -1
+        return None
+
+    def _path_icon(self, source: SourceFileModel) -> Any:
+        icon_name = 'folder' if source.path_isdir else 'file'
+        dark = uses_dark_mode()
+        if dark != self._icon_cache_dark:
+            self._icon_cache.clear()
+            self._icon_cache_dark = dark
+        if icon_name not in self._icon_cache:
+            self._icon_cache[icon_name] = get_colored_icon(icon_name)
+        return self._icon_cache[icon_name]
+
+    def headerData(
+        self,
+        section: int,
+        orientation: Qt.Orientation,
+        role: int = Qt.ItemDataRole.DisplayRole,
+    ) -> Any:
+        if role != Qt.ItemDataRole.DisplayRole:
+            return None
+        if orientation == Qt.Orientation.Horizontal and 0 <= section < len(self._HEADERS):
+            return translate('Form', self._HEADERS[section])
+        return None
--- a/src/vorta/views/partials/source_files_table_model.py
+++ b/src/vorta/views/partials/source_files_table_model.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from typing import Any, Dict, List, Optional, Set
 
-from PyQt6.QtCore import QAbstractTableModel, QModelIndex, QObject, Qt
+from PyQt6.QtCore import QAbstractTableModel, QModelIndex, QObject, QSortFilterProxyModel, Qt
 
 from vorta.i18n import trans_late, translate
 from vorta.store.models import SourceFileModel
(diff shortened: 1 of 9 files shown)


ebuzerdrmz44:
Added `SortProxyModel` to fix the same ≥2 GiB size-sort overflow  caught in #2519 .

  This is intentionally a near-duplicate of #2519's `ArchiveSortProxyModel`. I kept them per-PR so neither blocks the other . Not sure that's the right call long-term though: how do you think we should handle the duplication? Dedupe into one shared proxy now, or leave it until the first of the two merges and clean it up then?

src/vorta/views/source_tab.py
```diff
@@ -30,13 +31,9 @@
 logger = logging.getLogger(__name__)
 
 
-class SourceColumn:
-    Path = 0
-    Size = 1
-    FilesCount = 2
-
-
 class SizeItem(QTableWidgetItem):
```
m3nu: 🟡 **[dead-code]** `SizeItem` is now dead code — #2519 landed first

Your cross-PR note assigns `SizeItem` cleanup to "whichever of the two lands second". #2519 merged today and current master's `archive_tab.py` no longer imports `SizeItem` — so the cleanup now belongs to this PR.

**Fix:** Delete the `SizeItem` class, plus the imports that become orphaned with it: `QTableWidgetItem` and `sort_sizes`. Note the ruff config ignores `F401`, so lint won't flag the dead imports — needs the manual sweep.

src/vorta/views/partials/source_files_table_model.py
```diff
@@ -0,0 +1,185 @@
...
+        role: int = Qt.ItemDataRole.DisplayRole,
+    ) -> Any:
+        if role != Qt.ItemDataRole.DisplayRole:
+            return None
+        if orientation == Qt.Orientation.Horizontal and 0 <= section < len(self._HEADERS):
+            return translate('Form', self._HEADERS[section])
+        return None
+
+
+class SortProxyModel(QSortFilterProxyModel):
```
m3nu: 🟡 **[duplication]** Consolidate with `ArchiveSortProxyModel` (now on master)

This is a line-for-line duplicate of `ArchiveSortProxyModel` (`archive_table_model.py:199`), and per your PR comment the plan was to dedupe once the first of the two merged — #2519 has now merged.

Since both models define `SortRole = Qt.ItemDataRole.UserRole`, the proxy doesn't actually need a per-model class at all.

**Fix:** Extract one shared proxy, e.g. `views/partials/sort_proxy.py`:
```python
class SortProxyModel(QSortFilterProxyModel):
    """Sort proxy comparing `Qt.UserRole` keys in Python to avoid Qt's 32-bit int truncation."""

    def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
        lv = left.data(Qt.ItemDataRole.UserRole)
        rv = right.data(Qt.ItemDataRole.UserRole)
        if lv is None:
            return rv is not None
        if rv is None:
            return False
        return lv < rv
```
and use it here and in `archive_table_model.py` (drop `ArchiveSortProxyModel` or alias it). This prevents a third copy appearing in the upcoming smaller-tab migrations.

src/vorta/views/partials/source_files_table_model.py
```diff
@@ -0,0 +1,185 @@
...
+    """Sort proxy comparing `SortRole` keys in Python to avoid Qt's 32-bit int truncation."""
+
+    def lessThan(self, left: QModelIndex, right: QModelIndex) -> bool:
+        lv = left.data(SourceFilesModel.SortRole)
+        rv = right.data(SourceFilesModel.SortRole)
+        if lv is None:
```
m3nu: 🔵 **[strict-weak-ordering]** Both-None case violates strict weak ordering

When `lv` and `rv` are both `None`, `lessThan(a, b)` and `lessThan(b, a)` both return `True`, which violates the strict-weak-ordering contract Qt's sort relies on and can produce unstable orderings.

**Fix:**
```python
if lv is None:
    return rv is not None
```
This is the same nit deferred from #2519 — fixing it once in the shared proxy (see comment above) resolves both tables.

src/vorta/views/source_tab.py
```diff
@@ -320,29 +241,17 @@ def source_copy(self, index=None):
...
+            return
+        # Resolve the backing sources up front; row indices shift once rows are removed.
+        sources = [index.data(SourceFilesModel.SourceRole) for index in indexes]
+        for source in sources:
+            if source is None:
+                continue
+            source.delete_instance()
+            self._discard_update_thread(source.dir)
+            logger.debug(f"Removed source {source.dir}")
+        self.populate_from_profile()
```
m3nu, line 236: 🔵 **[behavior]** Full repopulate drops "Calculating…" placeholders on other rows

`populate_from_profile()` → `set_rows()` clears the model's `_calculating` set. So removing one source while others are mid-recalculation blanks their "Calculating…" cells (the data still lands correctly when each worker finishes — only the placeholder is lost). It also resets any ad-hoc sort back to the saved settings.

The old code removed rows without touching the rest of the table.

**Suggestion (fine as a follow-up):** add a targeted `remove_source(source)` to the model using `beginRemoveRows`/`endRemoveRows` instead of repopulating, leaving `_calculating` and the current sort untouched.

m3nu (commented):
## 🔍 Code Review

**PR:** #2523 — refactor(source): convert source files table to QTableView + QAbstractTableModel
**Branch:** `refactor/source-table-model` → `master`
**Changes:** +465 / −178 across 5 files

### 📋 Summary
Solid migration that honors every #2361 Phase 5 decision — but **#2519 merged earlier today**, which answers two open questions in this PR and calls for one cleanup round before merge.

### ✅ What's Good
- `SourceRole` resolves backing objects through the proxy on the destructive `source_remove` path — exactly the mapToSource footgun the #2361 decisions warn about.
- Path-keyed recalculation with a `None` return kills the #1080/#2435 row-index crash class, with a test for the removed-mid-calculation case.
- The `SortProxyModel` correctly follows the post-#2519 pattern, and the >2 GiB sort test covers the exact truncation regression that pattern exists for.
- i18n contexts check out (`Form` matches the `.ui` `<class>`; `Calculating…` keeps `SourceTab`), no DB reads in the model, thorough model tests.

### 🔍 Review Details
4 inline comment(s).

| Severity | Count |
|----------|-------|
| 🟡 Warning | 2 |
| 🔵 Info | 2 |

**Re: your dedup question in the comments** — merge order answered it: #2519 landed first, so I'd consolidate in this PR. Extract a shared generic proxy (both `SortRole`s are `Qt.UserRole`) into e.g. `views/partials/sort_proxy.py` and use it from both tables, before the smaller-tab migrations create a third copy. Details inline.

One non-inline nit: the PR description still says sorting is "not a `lessThan` proxy subclass" — stale relative to the code, worth updating so future archaeology isn't misled.

### 📊 Verdict: COMMENT 💬
Not merging as-is, but these are consequences of merge timing, not defects. One cleanup round and this is good to go.

---

ebuzerdrmz44:
on the `Calculating…`-placeholder-on-repopulate point . I will  add the targeted `remove_source()` (`beginRemoveRows`/`endRemoveRows`) in a separate small PR rather than widen this one.

m3nu (commented):
Reviewed in depth — the migration is correct and pattern-faithful: model in `partials/`, raw `SortRole` keys, every index access resolved via `index.data(SourceRole)` so the proxy footgun is handled, and per-path (rather than per-row-index) async updates kill the #1080/#2435 crash class outright. Nice catches along the way: `updateThreads` was a shared class attribute on master, and the shared `SortProxyModel` extraction includes the both-`None` strict-weak-ordering fix. The 18 new model tests are solid.

Two asks before merge:

1. **Guard against overlapping recalculations of the same path.** `_discard_update_thread` disconnects and drops *all* threads whose `objectName()` matches the path. If the same path is recalculated twice concurrently (double-click on Update), the first result to arrive drops the only reference to the still-running second thread — "QThread destroyed while running" territory. Simplest fix: skip spawning in `update_path_info` when the path is already pending.
2. **Add a remove-under-sort test**: sort by size descending, select the first visual row, call `source_remove()`, assert the correct DB record was deleted. That's the exact proxy/mapToSource scenario from #2361 and the one convention-critical path currently untested.

Known and fine to defer to the `beginRemoveRows` follow-up you already mentioned: the full model reset in `source_remove` wipes other rows' "Calculating…" placeholders and the selection.

Merge order: this lands after #2525/#2526 and will need a rebase around the `archive_tab.py` import block (this PR deletes `ArchiveSortProxyModel`).

Take-or-leave nits: pass a roles list on the `dataChanged` emits (the archive model does), type the icon cache `Dict[str, QIcon]`, `header.setVisible(True)` is redundant now that the `.ui` attribute is gone, and there's a stray blank line in the `sort_proxy.py` docstring.

ebuzerdrmz44:
addressed , waiting for #2526 to merge so that we can marge this too after rebase

m3nu (commented):
Content review passed — both asks from my last round are addressed in `f47320bc`. Holding at COMMENT rather than approving only because of two mechanical blockers, not defects: the branch is **CONFLICTING** against master, and **CI has never run on it**, so the "all tests passed" checklist item is currently unverified by the pipeline.

**Verified addressed:**

- **Duplicate-recalc guard** — `update_path_info` now early-returns when a thread with the same `objectName()` is already in `self.updateThreads`. That closes the "QThread destroyed while running" window in `_discard_update_thread`.
- **`test_source_remove_under_sort`** — sorts by size descending, selects visual row 0, asserts the *large* record is the one deleted. Exactly the proxy/`mapToSource` scenario from the #2361 decisions.
- All the take-or-leave nits landed: roles lists on the `dataChanged` emits, typed `Dict[str, QIcon]` icon cache, redundant `header.setVisible(True)` dropped, docstring cleaned.

**Things I checked independently that hold up:**

- i18n context is right — `source_tab.ui` has `<class>Form</class>`, so `trans_late('Form', 'Path'/'Size'/'File Count')` reuses the existing catalog entries, and `Calculating…` keeps the `SourceTab` context that `self.tr()` produced before.
- `QtCore` is still used at `source_tab.py:367` (`WindowType.Sheet`), so dropping the `findItems` call doesn't orphan the import.
- Sort persistence is intact: `header.sortIndicatorChanged` → `update_sort_order` is still wired, and `sortItems` → `sortByColumn` is the correct QTableView equivalent.
- `set_icons()` swapping the per-row `SourceFileModel.get()` loop for a `viewport().update()` is a strict improvement — it drops N DB queries per theme switch, and the dark-mode-keyed icon cache handles invalidation.
- Deleting `SizeItem.__lt__` / `FilesCount.__lt__` is a real correctness win beyond the sorting fix: `FilesCount.__lt__` returned `1`/`0` for non-numeric text, which is not a valid strict weak ordering.

**To merge:**

1. Rebase onto master once #2526 lands (the conflict is the `archive_tab.py` import block — this PR removes `ArchiveSortProxyModel` and `SIZE_DECIMAL_DIGITS`).
2. Get CI green.

I'll approve on a clean rebase with green checks — no further content review needed.

**One new nit worth folding into the rebase:** `SortProxyModel.lessThan` hardcodes `Qt.ItemDataRole.UserRole` instead of reading `self.sortRole()`. Since this is now the *shared* proxy the remaining tab migrations will reuse, honouring `setSortRole()` costs one line and avoids someone subclassing it later. No-op today, as both models define `SortRole = Qt.UserRole`.

Still fine to defer, as agreed: the full model reset in `source_remove()` wiping other rows' "Calculating…" placeholders and the selection, pending your targeted `beginRemoveRows`/`endRemoveRows` follow-up.

m3nu:
Content is good, but needs rebase.

ebuzerdrmz44:
Rebased, this last one closes #2361

m3nu (approved):
Reviewed the branch code directly at `cb9eb8d`. The consolidation round landed cleanly — `ArchiveSortProxyModel` is gone, both tables share `views/partials/sort_proxy.py`, and the both-`None` strict-weak-ordering nit from #2519 is fixed.

Checks that came back clean:

- No dangling references — `SourceColumn`, `SizeItem`, `FilesCount`, `SIZE_DECIMAL_DIGITS` are referenced nowhere else in `src/` or `tests/`; the only external consumer of `sourceFilesWidget` is `tests/unit/test_source.py`, which this PR updates.
- i18n holds. `pylupdate5 -translate-function trans_late` scans `.py` alongside `.ui`, so `trans_late('Form', ...)` extracts under the same context `source_tab.ui` declared (`<class>Form</class>`) and `Calculating…` keeps `SourceTab`. Existing catalog entries are reused; only `<location>` lines move.
- `sources_update` iterates the source model rather than the proxy, so dynamic re-sorting cannot shuffle rows out from under the loop.
- Icon caching is a real win: `get_colored_icon` does disk I/O plus an SVG re-render per call, and the old `set_icons` paid that once per row *plus* a DB `get()` per row.

Also worth calling out that this quietly fixes several real bugs beyond the migration itself: `updateThreads` was a **class** attribute shared across every `SourceTab` instance; `update_path_info` now refuses to spawn a duplicate worker for a path already calculating; and the thread-list removal now iterates a copy, where the old `set_path_info` mutated the list mid-iteration and silently skipped entries.

Merging as-is. Three follow-ups, none blocking:

1. The targeted `remove_source()` with `beginRemoveRows`/`endRemoveRows` you already offered — `source_remove` currently goes through `populate_from_profile()`, so a full model reset clears `_calculating` on unrelated rows and drops selection. Self-healing, since the pending workers still deliver through `set_path_info`, but worth tightening.
2. `sort_sizes()` in `src/vorta/utils.py` becomes dead production code after this merge — only `tests/unit/test_utils.py` still references it. Fine to remove it along with its test.
3. `proxy.sortRole` stays `DisplayRole` while `lessThan` reads `UserRole`. Dynamic re-sort still fires because the model's `dataChanged` emissions include `DisplayRole`, so this works today, but it is an implicit coupling in both proxies. Worth normalizing across Archive and Source together at some point rather than here.

One thing I looked at and decided is *not* a regression: `set_path_info` is no longer profile-scoped, so a worker outliving a profile switch could apply profile A's size to profile B's row with the same path. The old code had the same race via `SourceFileModel.get(dir=path, profile=self.profile())`, and workers were never cancelled on profile change either way.

PR #2525, opened 2026-06-29
refactor(archive): peel diff and rename into views/archive/ helpers (Phase 4 pt 2)

ebuzerdrmz44:
### Description

Phase 4 **part 2** of the Archive tab refactor. Part 1 (#2519) converted the table to `QAbstractTableModel` + `QTableView` and peeled the first two operations (`mount/unmount`, `extract`) into `views/archive/` collaborator classes. This PR continues that exact pattern, peeling **diff** and **rename** off the orchestrator into focused helpers:

- `views/archive/archive_diff.py` → `ArchiveDiff` (`diff_action`, `list_diff_result`, `show_diff_result`)
- `views/archive/archive_rename.py` → `ArchiveRename` (`cell_double_clicked`, `on_name_edited`, `_revert_name`, `rename_result`)

`ArchiveTab` stays the container (populate, selection, context-menu wiring, refresh, status, prune settings). Pure structural move — **no behavior change**.

Helpers follow the existing `ArchiveExtract` / `ArchiveMount` convention from #2519. The context menu stays in `ArchiveTab` since it just dispatches to the other helpers.

`delete` and `check/compact/prune` are held back for a follow-up PR so the helper shape can be reviewed here first. Happy to redirect the decomposition — this is the cheap point to do it.

> **Rebased onto master.** #2519 has merged, so this branch is now rebased onto it and the diff is exactly the part-2 work:
> - `src/vorta/views/archive/archive_diff.py` (new)
> - `src/vorta/views/archive/archive_rename.py` (new)
> - the `src/vorta/views/archive_tab.py` deltas (and the two test call-site updates)

### Related Issue

Part of #2361 (Phase 4 — split large tabs into parts). Builds on #2519.

### Motivation and Context

`archive_tab.py` was a large orchestrator mixing the table view with diff, rename, delete and repo-maintenance logic. #2361 calls for converting the table to a model (done in #2519) and then peeling the remaining operations into focused helpers, keeping `ArchiveTab` as the container. This PR does that for diff and rename, making each operation independently readable and shrinking the orchestrator, with no change to runtime behavior.

### How Has This Been Tested?

[...]

Issue #2493, opened 2026-04-06
`Keep all archives of the last` option does not persist


CPU-Blanc:
### Description

Setting the option `Keep all archives of the last` does not persist. I am unsure when exactly it is unset as sometimes it appears random, but it will _always_ reset on system reboot.

While the field is populated, pruning is completed as expected and archives are retained. On reboot, the field is empty and no longer works as expected, pruning archives that should be retained.

This was not an issue on older versions of Vorta and appears to be a regression, but I cannot specify exactly which version as I was unaware of how long this has been not working

### Reproduction

- [x] I tried to reproduce the issue.
- [x] I was able to reproduce the issue.

### OS

Arch 6.9.11-arch1-1, Hyprland 0.52.2

### Version of Vorta

0.11.3

### What did you install Vorta with?

Distribution package

### Version of Borg

1.4.4

### Logs

```pytb
N/A. No relevant lines in the log file
```

CPU-Blanc:
Further update and possible info.
I have 2 profiles for backing up different directories at different intervals. The "keep all archives" option was set in profile A, which runs multiple times a day. Profile B runs once a day.
On boot today I quickly opened Vorta to see if the option persisted. It did, but Profile B ran (as it is set to at a specified time), and once it completed, the option was now empty on both profiles

m3nu:
Fixed in 4838d9ad. `populate_from_profile` called `setValue()` on the prune-interval spinboxes, which fires `valueChanged` → `save_prune_setting`. At that point `prune_keep_within` still held stale text from the previous profile (or empty on startup), so the persisted value got overwritten with `""`. Now block signals on the prune widgets during population.

PR #2526, opened 2026-07-13
refactor(archive)  tab split  (phase 4 pt 3) : delete + maintenance helpers

ebuzerdrmz44:
### Description
 
Continues the `ArchiveTab` breakup started in #2519 (pt 1) and #2525 (pt 2, diff + rename). This peels the remaining destructive/maintenance actions out of the god-object into the same collaborator-class pattern under `views/archive/`:

- **`archive/archive_delete.py`**  `ArchiveDelete` (`delete_action` / `delete_result`)
- **`archive/archive_maintenance.py`**  `ArchiveMaintenance` (`check`, `compact`, `prune` actions + results, `save_prune_setting`, `toggle_compact_button_visibility`)

`archive_tab.py` shrinks by 130 lines and now just wires these helpers alongside the pt 2 `ArchiveDiff` / `ArchiveRename`. 

No behavior change .

### Notes
- **Stacks on #2525.** Until that merges, this PR's diff includes pt 2's commit; review only the top commit here. Kept as **draft** for that reason.
- The helpers call the tab's `_set_status` / `_toggle_all_buttons`. De-underscoring that tab/helper boundary is a deliberate later cleanup, kept consistent with pt 2 rather than widened here.
- Test call sites updated to the new module paths (`tests/unit/test_archives.py`, `tests/integration/test_archives.py`).

### DIFF
--- a/src/vorta/application.py
+++ b/src/vorta/application.py
@@ -300,7 +300,7 @@ class VortaApp(QtSingleApplication):
                         double_newline,
                         str(exception),
                         double_newline,
-                        self.tr('Consider removing or repairing this file to ' 'get rid of this message.'),
+                        self.tr('Consider removing or repairing this file to get rid of this message.'),
                     ),
                 )
                 return
--- a/src/vorta/views/archive/archive_delete.py
+++ b/src/vorta/views/archive/archive_delete.py
@@ -1,0 +1,51 @@
+from vorta.borg.delete import BorgDeleteJob
+from vorta.store.models import ArchiveModel
+
+
+class ArchiveDelete:
+    def __init__(self, tab):
+        self.tab = tab
+
+    def delete_action(self):
+        # Since this function modifies the UI, we can't put the whole function in a JobQueue.
+        archives = [archive.name for archive in self.tab.selected_archives()]
+
+        if not archives:
+            self.tab._set_status(self.tab.tr("No archive selected"))
+            return
+
+        params = BorgDeleteJob.prepare(self.tab.profile(), archives)
+        if not params['ok']:
+            self.tab._set_status(params['message'])
+            return
+
+        if len(archives) > 1:
+            body = self.tab.tr("Are you sure you want to delete all the selected archives?")
+        else:
+            body = self.tab.tr("Are you sure you want to delete the selected archive?")
+        if not self.tab.confirm_dialog(self.tab.tr("Confirm deletion"), body):
+            return
+
+        job = BorgDeleteJob(params['cmd'], params, self.tab.profile().repo.id)
+        job.updated.connect(self.tab._set_status)
+        job.result.connect(self.delete_result)
+        self.tab._toggle_all_buttons(False)
+        self.tab.app.jobs_manager.add_job(job)
+
+    def delete_result(self, result):
+        archives = result['params']['archives']
+        if result['returncode'] == 0:
+            if len(archives) > 1:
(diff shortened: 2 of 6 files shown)


m3nu (dismissed):
Reviewed the pt-3 commit (93a4c93) in depth — it's a faithful, line-by-line move of delete/check/compact/prune into `ArchiveDelete`/`ArchiveMaintenance`, matching the extract/mount collaborator pattern. Selection goes through `selected_archives()` / `ArchiveRole`, so it's proxy-safe; helpers are strongly referenced on the tab, so the signal lifecycle is sound; all test call sites are updated with no stale patches.

Two nits before merge:

1. The description says `toggle_compact_button_visibility` moved into `ArchiveMaintenance` — it actually stays on the tab, which is correct (`application.py` calls `archiveTab.toggle_compact_button_visibility()`), but please fix the description.
2. The `application.py` string-literal cleanup hunk is unrelated to this PR. Harmless, but out of scope.

Once #2525 is squash-merged, rebase onto master (the diff collapses to the pt-3 commit) and un-draft — then this is good to go. Optional but cheap: a smoke test each for `check_action`/`compact_action`/`prune_action`, which currently have no direct coverage (pre-existing gap, not something this PR introduced).

ebuzerdrmz44:
@m3nu rebased onto master now that #2525 is in and addressed the review nits. The old approval got auto-dismissed when I pushed, so this should just need a quick re-confirm, not a full re-review.

Also heads up: #2523 is waiting on this one. It deletes ArchiveSortProxyModel and touches the same import block in archive_tab.py, so we can rebase it once this lands instead of fighting the same conflict twice.

m3nu (approved):
Re-reviewed the current head after the rebase onto master. The pt-3 move is faithful and correct — approving.

**What I verified:**

- `ArchiveDelete` / `ArchiveMaintenance` follow the stateless-collaborator pattern from `archive_extract.py` / `archive_mount.py`, and both are held as strong attributes on the tab, so the `job.result.connect(...)` lifetimes are sound.
- Selection goes through `self.tab.selected_archives()` → `ArchiveRole`, so the proxy `mapToSource` footgun is handled.
- Every call site is updated: the `bDelete` / `compactButton` / `bPrune` / `bCheck` connections, the `archiveitem_contextmenu` `button_connection_pairs` list, the prune-interval `valueChanged` connections, and both test files. No stale references left.
- `test_populate_does_not_overwrite_prune_keep_within` (the #2493 regression) exercises the real signal path rather than patching the method, so it still guards after the move.
- `toggle_compact_button_visibility` correctly stayed on the tab — `application.py` calls `archiveTab.toggle_compact_button_visibility()` directly.

**Two things from my earlier review are still open on this head — please fix before you hit merge:**

1. The PR description still lists `toggle_compact_button_visibility` as moving into `ArchiveMaintenance`. It didn't, and that's the right call. Worth correcting so future archaeology isn't misled.
2. The `application.py` string-literal hunk (line 303, implicit concat → single literal) is still in the diff. Runtime-identical and harmless to the `.ts` catalogs, but unrelated to the tab split. Drop it or split it out.

Neither blocks, hence the approval rather than another round.

**Optional, pre-existing gap:** `check_action` / `compact_action` / `prune_action` still have no direct coverage. A smoke test each, mirroring `test_archive_delete`, would be cheap insurance now that they live in a new module.

Merge order note: #2523 conflicts on the `archive_tab.py` import block, so it rebases after this lands.

Issue #2360, opened 2026-01-26
Feature: Scheduler Refactor - Jobs Store, Separation of Concerns, and Jobs View


m3nu:
## Summary

Refactor Vorta's scheduler to improve reliability, maintainability, and user visibility into scheduled and past jobs.

### 1. Add Persistent Jobs Store

**Current state:** `EventLogModel` only tracks *executed* jobs. Pause states and pending schedules are stored in-memory and lost on restart.

**Goal:** Add a database model to track:
- Pending/scheduled jobs (survive restarts)
- Skipped jobs with reasons (audit trail)
- Link between scheduled intent and execution results

### 2. Separate Scheduling, Execution, and State Management

**Current state:** `VortaScheduler` (532 lines) handles scheduling, execution triggering, and state management in one class with manual lock management and interleaved post-backup tasks.

**Goal:** Split into focused components:
- **Scheduling**: Calculate next run times, manage timers
- **Execution**: Create jobs, queue them, handle results, coordinate post-backup tasks
- **State**: Pause/resume persistence, job status tracking, crash recovery

### 3. Add Jobs View UI

**Current state:** Log page only shows past jobs. No view of pending jobs or filtering options.

**Goal:** New Jobs page showing:
- Pending (scheduled) and completed jobs in one view
- Filtering by profile, repository, job type, and status
- Skip reasons for jobs that didn't run
- Ability to cancel pending or re-queue failed jobs

### 4. Improve Reliability

**Current issues:**
- QTimer 32-bit limitation (~24.8 days) with imprecise workaround
- Race conditions between job checks and submission
- Post-backup tasks not guaranteed sequential
- DBus sleep/resume detection fails silently

**Goal:** More reliable scheduling that handles edge cases gracefully.

## Out of Scope

- Cron expression support (e.g., "weekdays at 9am")
- Job dependencies/chaining
- Retry policies for failed jobs

## Open Questions

1. Should the Jobs view replace the existing Log page or be a separate tab?
2. What filtering options are most important?
3. How should crash recovery handle jobs that were running when the app crashed?

## Related Files

Key files that will be affected:
- `src/vorta/scheduler.py` - Current scheduler (to be refactored)
- `src/vorta/store/models.py` - Add new job model
- `src/vorta/borg/jobs_manager.py` - Job queue management
- `src/vorta/views/schedule_tab.py` - Add Jobs page
- `src/vorta/views/log_page.py` - Existing log view

khadejaa23:
[...]

PR #2530, opened 2026-07-19
add JobModel: persistent scheduler job store (#2360, phase 1)

ebuzerdrmz44:
### Description

First PR of the scheduler refactor (#2360), split into small stacked PRs. The full breakdown is on the issue.

This adds `JobModel` together with its first consumer, so the table arrives with a real write path rather than as a bare schema.

`JobModel` records intent: what the scheduler meant to do, including the runs that never reach borg and so never produce a log line. `EventLogModel` stays the record of what executed, linked through the `event_log` FK. `profile` and `repo_url` are plain strings mirroring `EventLogModel`, with `profile_name` stored next to the id so the history still names the profile after someone deletes it.

No migration. `connection.py` builds tables with `create_tables()`, which is safe by default, so the table shows up on fresh and existing DBs alike.

### What writes to it

Three call sites go through `VortaScheduler._record_skip`:

- borg's prepare step fails in `create_backup`
- the repo is busy with another job
- the network is down for a catch-up run in `set_timer_for_profile`

A real error records `status='failed'`. The expected cases (no WiFi, metered connection, busy repo) record `status='skipped'`.

The catch-up site passes the missed occurrence as `scheduled_at` and writes through `get_or_create`. A network that stays down overnight then leaves one row per missed run instead of one per timer tick.

`init_db` deletes job rows older than six months, next to the existing `EventLogModel` sweep.

`db_lock` moved from `vorta.borg.borg_job` to `vorta/store/models.py`, which is why `borg_job.py` turns up in the diff.

### Not written yet

`event_log` and the `scheduled`, `running` and `interrupted` statuses have no producer in this PR. They arrive with the run-lifecycle and crash-recovery phases in the #2360 breakdown, so there is nothing to go hunting for here.

### DIFF
--- a/src/vorta/borg/borg_job.py
+++ b/src/vorta/borg/borg_job.py
@@ -20,11 +20,10 @@ from vorta.borg.jobs_manager import JobInterface
 from vorta.i18n import trans_late, translate
 from vorta.keyring.abc import VortaKeyring
 from vorta.keyring.db import VortaDBKeyring
-from vorta.store.models import EventLogModel
+from vorta.store.models import EventLogModel, db_lock
 from vorta.utils import borg_compat, pretty_bytes
 
 keyring_lock = Lock()
-db_lock = Lock()
 logger = logging.getLogger(__name__)
 
 FakeRepo = namedtuple('Repo', ['url', 'name', 'id', 'extra_borg_arguments', 'encryption'])
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -331,6 +331,15 @@ class VortaScheduler(QtCore.QObject):
                     return  # create_backup will lead to a call to this method
                 elif profile.schedule_make_up_missed and not self._net_up and needs_network:
                     logger.debug('Skipping catchup %s (%s), the network is not available', profile.name, profile.id)
+                    with db_lock:
+                        JobModel.create(
+                            profile=profile_id,
+                            repo_url=profile.repo.url if profile.repo else None,
+                            job_type='backup',
+                            status='skipped',
+                            trigger='catchup',
+                            skip_reason='Network unavailable for catch-up.',
+                        )
 
                 # calculate next time from now
                 if profile.schedule_mode == 'interval':
@@ -433,6 +442,15 @@ class VortaScheduler(QtCore.QObject):
         # Skip if a job for this profile (repo) is already in progress
         if self.app.jobs_manager.is_worker_running(site=profile.repo.id):
             logger.debug('A job for repo %s is already active.', profile.repo.id)
+            with db_lock:
+                JobModel.create(
+                    profile=profile_id,
+                    repo_url=profile.repo.url if profile.repo else None,
+                    job_type='backup',
+                    status='skipped',
+                    trigger='scheduled',
+                    skip_reason='Repository is busy with another job.',
+                )
             self.pause(profile_id)
             return
 
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -7,13 +7,13 @@ from datetime import datetime as dt
 from datetime import timedelta
 from typing import Any, NamedTuple
 
+import peewee as pw
 from packaging import version
 from PyQt6 import QtCore, QtDBus
 from PyQt6.QtCore import QTimer
 from PyQt6.QtWidgets import QApplication
 
 from vorta import application
-from vorta.borg.borg_job import db_lock
 from vorta.borg.check import BorgCheckJob
 from vorta.borg.compact import BorgCompactJob
 from vorta.borg.create import BorgCreateJob
@@ -324,22 +324,19 @@ class VortaScheduler(QtCore.QObject):
                             profile.name,
                             profile_id,
                         )
-                        self.create_backup(profile_id)
+                        self.create_backup(profile_id, trigger=JobModel.Trigger.CATCHUP.value)
                     finally:
                         self.lock.acquire()  # with-statement will try to release
 
                     return  # create_backup will lead to a call to this method
                 elif profile.schedule_make_up_missed and not self._net_up and needs_network:
                     logger.debug('Skipping catchup %s (%s), the network is not available', profile.name, profile.id)
-                    with db_lock:
-                        JobModel.create(
-                            profile=profile_id,
-                            repo_url=profile.repo.url if profile.repo else None,
-                            job_type='backup',
-                            status='skipped',
-                            trigger='catchup',
-                            skip_reason='Network unavailable for catch-up.',
-                        )
+                    self._record_skip(
+                        profile,
+                        JobModel.Trigger.CATCHUP.value,
+                        'Network unavailable for catch-up.',
+                        scheduled_at=next_time,
+                    )
 
                 # calculate next time from now
                 if profile.schedule_mode == 'interval':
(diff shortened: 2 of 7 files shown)


ebuzerdrmz44:
Reworked this from the bare model into something with an actual use.

The three schema questions are decisions now. Profile and repo are plain strings like _`EventLogModel`_, so a job's history sticks around even after its profile or repo gets deleted. I kept scheduled and interrupted in the status enum but nothing writes them yet.

I also folded in the first consumer, skip-reason recording, so there's a real path through the table instead of an empty schema sitting there. The skip write goes through `db_lock` the same way the EventLogModel writes do.

Ready for another look whenever you get a chance.

ebuzerdrmz44:
@m3nu Since this is the first of a stack, here's the full shape of #2360  so it's clearer what this PR is (and isn't) trying to do. Roughly 4 phases,10-11 small PRs ( I broke my original roadmap in #2360 to make all of it easy to review :laughing:  ):

  Phase A : Persistence (the jobs store)
 - PR1  (this PR): the JobModel table + the first skip-reason consumer + record skip reasons at the remaining skip points (repo busy, network down, mount) 
  - PR2: persist pause/resume state so it survives a restart

  Phase B : Split VortaScheduler into State / Scheduling / Execution (PR4–6). The Execution one wires the job→result link and makes post-backup tasks run in order.

  Phase C : Reliability (PR7–8): the QTimer 24.8-day overflow, and sleep/resume detection on non-logind systems.

  Phase D : Jobs View UI (PR9–11): the Jobs page, filters, cancel/re-queue. _I'm holding this whole phase_ until you weigh in on three open questions on the issue . 

  This PR is deliberately the narrow shape-approval slice: get the table right, prove it works with one consumer, then stack the rest on top once you're happy with it. Everything downstream depends on this one merging first.

m3nu (changes requested):
Thanks for folding in the first consumer — a table with a real write path is much easier to judge than a bare schema, and the phased breakdown on #2360 is genuinely helpful.

The overall shape is defensible: `JobModel` records *intent* (including the things that never produce a log line), `EventLogModel` stays the record of what *executed*, linked through the `event_log` FK. Plain strings for `profile`/`repo_url` mirror `EventLogModel` exactly. Your no-migration claim checks out — `init_db` calls `DB.create_tables([...])` unconditionally with `safe=True`, so the table appears on fresh and existing DBs alike.

Requesting changes on three points, then this is good to build on.

### Required

**1. Failures are being recorded as skips.** In `create_backup`, the `JobModel.create(...)` sits in the outer block covering *both* branches of the `level == 'error'` check. So a genuine failure — "Conditions for backup not met", repo unreachable, source missing — gets written with `status='skipped'` and the error text in `skip_reason`. That defeats the table's stated purpose of capturing intent, and it'll poison the Phase D Jobs View from day one. Use `status='failed'` when `level == 'error'`; keep `skipped` for the expected WiFi/metered/busy cases.

**2. No retention policy → unbounded growth.** `init_db` purges `EventLogModel` rows older than six months. `JobModel` gets no equivalent, so on a machine with an aggressive schedule and a flaky network the table only ever grows. Please add the purge alongside the existing one *in this PR* — retrofitting it after users have accumulated rows is strictly worse.

**3. Extract the triplicated write.** Three near-identical `with db_lock: JobModel.create(...)` blocks. Collapse into a `_record_skip(self, profile, trigger, reason, status='skipped')` helper on `VortaScheduler`. With ~10 more PRs stacked on this, it will not stay at three.

### Cheap, should fix here

**4. Move `db_lock` out of `vorta.borg.borg_job`.** Importing a DB-serialization lock from the borg job module into the scheduler is a layering inversion — it belongs in `vorta/store/`. Every PR in the stack will want it, so move it now, while there's exactly one new importer.

**5. Constrain the string enums.** `status`, `job_type` and `trigger` are free-form `CharField`s written with bare literals. Module-level constants (or a small `JobStatus` class) in `models.py` will stop the eight downstream PRs drifting on `'skipped'` vs `'skip'`.

**6. Missing test for the third write site.** `test_scheduler.py` covers the prepare-failure and repo-busy skips, but not the catchup/network-down write in `set_timer_for_profile`.

### Worth settling before the stack builds on this schema

**7. `profile` stores the profile *id* as a string.** Consistent with `EventLogModel`, so I'm not objecting — but the rationale in your comment ("a job's history sticks around even after its profile gets deleted") isn't actually achieved: what survives is a dangling numeric id with no name. If the Phase D Jobs View is meant to render history for deleted profiles, store `profile_name` alongside it.

**8. Dead columns.** `scheduled_at`, `event_log` and the `scheduled`/`running`/`interrupted` statuses are never written by production code — only by the test. Fine as a phase-1 slice given the roadmap, but say so explicitly in the description so a reviewer doesn't go hunting for the writer.

**9. No index.** The Jobs View will filter on profile/status/created_at. Adding it now avoids a migration later.

On your original three questions: agreed on plain strings (2), and yes, a startup sweep marking leftover `running` jobs as `interrupted` is enough for crash recovery (3). For (1), your instinct is right — keep next-run derivation in `set_timer_for_profile` as the single source of truth, and persist only skipped/interrupted/completed. Don't materialise scheduled rows.

Minor: this PR tracks #2360, not #2361 — worth keeping the cross-links straight since the two refactors are running in parallel.

ebuzerdrmz44:
want to flag this before you look again: I moved `db_lock` into `store/models.py` like you asked, but ended up not using it in the scheduler, so `borg_job` is still its only importer. The lock is held across `process_result()`, which loops every archive on a `borg list` . I  think  grabbing it from the scheduler would be the first time we take it on the GUI thread, and a large repo sync _**would freeze**_ the UI. Recording is best-effort instead, so a lost row can't affect scheduling.

Narrowing the lock in `borg_job.run` to just the `save()` call would also fix it properly. Which would you rather?

ebuzerdrmz44:
@m3nu  All nine done , ready for another look.

I also deduped the catch-up skips, since a down network wrote a row per tick. And I rewrote the description.

ebuzerdrmz44:
@m3nu bumping this one since the whole #2360 stack sits behind it. All nine points are in as of 43662e8.
 
Still open from my side: the db_lock question. Recording is best-effort right now so a lost row can't affect scheduling. If you'd rather I narrow the lock in borg_job.run to just the save() calls, that's a small standalone PR and I'll open it.

#2532 (pause persistence) is the phase 1 sibling, also green and independent of this one, if it's easier to look at them together.

m3nu (approved):
All nine points check out in 43662e8 — the failed/skipped split, the six-month purge next to the `EventLogModel` sweep, `_record_skip`, the enums, `profile_name`, the index, and the catch-up test with the dedupe assertion. Merging.

**On `db_lock`:** your read is right — don't take it in the scheduler. `borg_job.py:342` holds the lock across `process_result`, and `BorgListRepoJob.process_result` issues a statement per archive with no `atomic()` wrapper, so acquiring that on the GUI thread would freeze the UI for the length of a repo sync. Best-effort recording is the correct choice for a GUI-thread write and stays as-is.

Yes to the follow-up PR narrowing the lock to just the `save()` calls — the current scope is wrong on its own merits, it's a DB-serialization lock being used to serialize archive processing. Frame it as a behaviour change rather than cleanup, though: today the lock guarantees two concurrent repo syncs never interleave their writes, and narrowing hands that to SQLite's busy timeout (WAL, no `timeout` pragma, so pysqlite's 5s default). Almost certainly fine, since each statement holds the write lock for milliseconds — but worth saying out loud in the PR description.

**Merge order for the rest of the stack:** #2537 next — it's independent and only overlaps #2532, in `VortaScheduler.__init__`. Then #2532 last, so it absorbs the rebase against both; that also makes it the natural home for the "a pause isn't recorded as a skip" gap you flagged. I'm reviewing #2532 now — it needs one fix before it goes in, I'll leave details there.

PR #2538, opened 2026-08-12
add a Jobs view: pending runs and recorded jobs in one table (#2360, phase 4)

ebuzerdrmz44:
## Description

Adds the Jobs view from #2360 goal 3. One table showing what the scheduler is holding a time for, and what it has already recorded.

Depends on nothing in flight. Branches off master, and does not touch `log_page.py`.

## Screenshot

Screenshot from 2026-08-12 14-40-26

## Where it lives

A `Jobs` page in the Schedule tab's toolbox, following the Related Files map in the issue (`views/schedule_tab.py - Add Jobs page`), which also leaves the existing Log page alone.

The table is **global, not scoped to the selected profile**, so it has a Profile column and lists every profile's jobs. Flagging it because it was one of the issue's open questions and was never settled in words.

## Where the rows come from

Recorded rows come from `JobModel`. Pending rows come from the scheduler's in-memory timers through a new `VortaScheduler.pending_jobs()`. So nothing writes `Status.SCHEDULED` to the table and the view has to merge the two halves itself.

`pending_jobs()` returns a `PendingJob` per profile whose timer is `SCHEDULED` or `TOO_FAR_AHEAD`, soonest first. It exists so the page never reads `self.timers` directly, which keeps the split in phase 2 an internal change.

## Why the refresh is split in two

`reload_records` re-reads `JobModel` and is wired to `backup_finished_event`. `reload_pending` only walks the timers and is wired to `schedule_changed`.

They are separate because `schedule_changed` is a hot signal: `reload_all_timers` emits once per profile every 15 minutes, and three of the four emit sites fire while `VortaScheduler.lock` is held. A single refresh doing `JobModel.select()` would put a full table read inside the scheduler's critical section on every schedule change. The record query is also capped at 200 rows, since the 6 month purge is the only other bound on that table.

One residual worth naming: `pending_jobs()` still does a primary key profile lookup per armed timer inside that lock. That is a handful of rows rather than the whole history, so I left it rather than caching profile names in the scheduler.

## What is deliberately not here

- **No `completed` or `running` rows.** Nothing writes those statuses yet, so the table renders pending plus skipped and failed. The lifecycle transitions and the `event_log` link belong to the Execution phase, and the view picks them up for free once they land.
- **No filters.** Next PR, as a combo box over the proxy. Status first, then profile. The Repository column is carried on the row for the same reason, so the repo filter has something to filter on.
- **No source role on the model, and no actions.** Cancel and re-queue arrive with the actions PR, and the role that maps a proxy row back to a job goes in with its first real consumer rather than as an unused accessor here.

## Notes

No schema change, no migration.

The Time column means `created_at` for a recorded job and the scheduled time for a pending one. That reads fine today because records only exist for skips, but once jobs carry real execution timestamps it is worth deciding whether the column is "when recorded" or "when due".

Tests cover the accessor (scheduled, too far ahead for a QTimer, profile deleted under its timer, never-run profile), the row rendering for both sources, and the page itself through a real `schedule_changed` emission.

Merging this allows me to go on Phase D.

### DIFF
--- a/src/vorta/assets/UI/jobs_page.ui
+++ b/src/vorta/assets/UI/jobs_page.ui
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>JobsPage</class>
+ <widget class="QWidget" name="JobsPage">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>687</width>
+    <height>384</height>
+   </rect>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <widget class="QTableView" name="jobsTable">
+     <property name="font">
+      <font>
+       <pointsize>11</pointsize>
+       <bold>false</bold>
+      </font>
+     </property>
+     <property name="showGrid">
+      <bool>false</bool>
+     </property>
+     <property name="sortingEnabled">
+      <bool>true</bool>
+     </property>
+     <attribute name="verticalHeaderVisible">
+      <bool>false</bool>
+     </attribute>
+    </widget>
+   </item>
+  </layout>
+ </widget>
+ <resources/>
+ <connections/>
+</ui>
--- a/src/vorta/assets/UI/schedule_tab.ui
+++ b/src/vorta/assets/UI/schedule_tab.ui
@@ -93,6 +93,21 @@
       <layout class="QGridLayout" name="shellCommandsLayout">
       </layout>
       </widget>
+     <widget class="QWidget" name="page_jobs">
+      <property name="geometry">
+       <rect>
+        <x>0</x>
+        <y>0</y>
+        <width>687</width>
+        <height>384</height>
+       </rect>
+      </property>
+      <attribute name="label">
+       <string>Jobs</string>
+      </attribute>
+      <layout class="QGridLayout" name="jobsLayout">
+      </layout>
+      </widget>
     </widget>
    </item>
   </layout>
--- a/src/vorta/scheduler.py
+++ b/src/vorta/scheduler.py
@@ -77,6 +77,13 @@ class ScheduleStatus(NamedTuple):
     time: dt | None = None
 
 
+#: Timer states that hold a time for a run, and the row status each one shows as.
+PENDING_STATUSES = {
+    ScheduleStatusType.SCHEDULED: JobModel.Status.SCHEDULED.value,
+    ScheduleStatusType.PAUSED: JobModel.Status.PAUSED.value,
+}
+
+
 class PendingJob(NamedTuple):
     """A run the scheduler is holding a time for, but that hasn't been recorded yet."""
 
@@ -84,11 +91,15 @@ class PendingJob(NamedTuple):
     profile_name: str
     repo_url: str | None
     scheduled_at: dt
+    status: str
 
 
 class VortaScheduler(QtCore.QObject):
     #: The schedule for a profile changed.
     schedule_changed = QtCore.pyqtSignal()
+
+    #: A job outcome was recorded.
+    jobs_changed = QtCore.pyqtSignal()
 
     def __init__(self) -> None:
         super().__init__()
@@ -533,7 +544,8 @@ class VortaScheduler(QtCore.QObject):
         pending = []
 
         for profile_id, timer in self.timers.items():
-            if timer['type'] not in (ScheduleStatusType.SCHEDULED, ScheduleStatusType.TOO_FAR_AHEAD):
+            status = PENDING_STATUSES.get(timer['type'])
+            if status is None:
                 continue
 
             profile = BackupProfileModel.get_or_none(id=profile_id)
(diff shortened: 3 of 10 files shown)


m3nu (changes requested):
Reviewed against a local rebase onto current master (`ebf75712`, i.e. with #2532 and #2547 in). The structure holds up — the model/page split follows the phase 5 table convention, `pending_jobs()` guards the deleted-profile race, and the page never touches `self.timers` directly. One blocker, one nit, plus the rebase itself.

## Rebase: the silent conflict we both flagged is real

Rebasing onto `ebf75712` conflicts in **`tests/unit/test_scheduler.py` only**, and only in the import hunk. `src/vorta/scheduler.py` auto-merges clean — which leaves `pending_jobs()` referencing `ScheduleStatusType.TOO_FAR_AHEAD`, deleted by #2547. So a routine conflict resolution ships an `AttributeError` on the Jobs page. Worth doing deliberately:

- `scheduler.py:536` → `if timer['type'] is not ScheduleStatusType.SCHEDULED:`
- Keep `test_pending_jobs_includes_a_run_too_far_ahead_for_a_timer` rather than dropping it. After #2547 a 5-week-out run is armed in chunks and is genuinely `SCHEDULED`, so the case still deserves coverage — it's the assertion that changes, not the test's reason for existing. I renamed mine to `test_pending_jobs_includes_a_run_beyond_a_single_timer_chunk` and flipped the assert to `ScheduleStatusType.SCHEDULED`.

With that: `git grep TOO_FAR_AHEAD` is empty, full unit suite is 271 passed / 7 skipped, `ruff check` clean.

## Blocker: recorded jobs never refresh in normal operation

`reload_records` is wired to `backup_finished_event`, and that signal is emitted only by `BorgJob` subclasses — 13 sites, all under `src/vorta/borg/`. Every `_record_skip` call site is a path where no Borg command runs. Since skips and failures are the only rows `JobModel` holds today, the recorded half of the table is stale by construction:

```
STATUSES AFTER SKIP + schedule_changed: ['scheduled']   # the skip isn't there
```

The test in this PR tips it off, actually — `test_jobs_page_merges_pending_runs_with_stored_records` has to call `page.reload_records()` by hand, because no signal covers it.

This compounds with the paused-profile question, and the combination is worse than either half alone. Two of the three `_record_skip` sites call `self.pause(profile_id)` right after, which flips the timer to `PAUSED` — filtered out of `pending_jobs()`:

```
BEFORE PAUSE: ['scheduled'] | AFTER PAUSE: []
Schedule page status: ScheduleStatus(type=PAUSED, time=05:00)
```

So at the exact moment a profile has something to report, the Jobs view goes **completely empty** for it — pending row gone, skip record not loaded — while the Schedule page next to it reads "Paused until 05:00".

Two things to fix:

1. Add a dedicated signal (`jobs_changed` or similar) emitted from `_record_skip`, wired to `reload_records`. Please don't just hang `reload_records` off `schedule_changed` — that puts `JobModel.select()` back inside the scheduler's critical section, which is exactly what your two-way split exists to avoid. Your reasoning there is right; it just needs the third signal to be complete.
2. On the paused question you raised: include `PAUSED` timers in `pending_jobs()`. A pause is a held time with a `dt`, the row can carry the status, and it keeps the Jobs view consistent with what the Schedule page is saying. Happy to hear the argument for excluding them, but silently dropping them isn't the right default.

## Nit: use the shared sort proxy

`jobs_page.py:22` uses a raw `QSortFilterProxyModel`; `archive_tab.py` and `source_tab.py` use `views/partials/sort_proxy.py::SortProxyModel`. Not a bug as written — no `UserRole` keys on the model, and `%Y-%m-%d %H:%M` happens to sort lexicographically the same as chronologically — but it diverges from the convention, and it breaks quietly the first time someone changes that format string. Cheap to align now.

## Checked, no action needed

`view-list-details` icon exists; toolbox index 4 matches the page order in `schedule_tab.ui`; rendering raw enum values in Type/Trigger/Status matches how the existing Log page does it. New translatable strings aren't in the `.ts` files yet, but that's the periodic resync, not something for this PR.

## Answering the open question in your description

The global, not-profile-scoped table is the right call. The scheduler is global, and a per-profile view would hide precisely the cross-profile contention ("Repository is busy with another job.") that the skip records exist to explain.

Nice work overall — the reasoning in the description about why the refresh is split is the part I'd have asked for if it weren't already there. Fix the record refresh and the paused rows and this is good to go.

ebuzerdrmz44:
Thanks for the review,all four addressed and pushed: rebased onto master.

I will start Phase B today , this closes Phase D.

m3nu:
Do you still get the display bug shown in your screenshot on the left side? The profile list should expand to push the buttons down.

ebuzerdrmz44:
> Do you still get the display bug shown in your screenshot on the left side? The profile list should expand to push the buttons down.

Yes, it happens when I make the window bigger.

ebuzerdrmz44:
> > Do you still get the display bug shown in your screenshot on the left side? The profile list should expand to push the buttons down.
> 
> Yes, it happens when I make the window bigger.

But it predates the PR.

m3nu:
Planned for today.

Issue #337, opened 2019-07-28
Why is vorta using 500MB+ memory when in tray?


JackDinn:
Yesterday i noticed vorta was using over half a Gig of memory, I just restarted it and continued with my day but this morning i notice that again it was creeping up and was at 300MB+
![](https://i.imgur.com/i9KAwRh.png)

Vorta 0.6.21
System:    Host: greg-pc Kernel: 5.1.19-1-MANJARO x86_64 bits: 64 compiler: gcc v: 9.1.0 Desktop: KDE Plasma 5.16.3 

As i write this its now up to 370MB, please advise.

m3nu:
PyQt isn't very efficient with RAM. Usage should be a bit lower when the settings window is closed.

You can also look into the upcoming Go implementation. It's a lot more efficient, but doesn't have all the features of the Python version yet.

https://github.com/borgbase/vorta-go

JackDinn:
> Usage should be a bit lower when the settings window is closed.
This memory usage is with the settings window closed.

I shall take a quick look at the go front-end, thx.

m3nu:
While I'd love to improve memory usage, this is mainly down to Qt. If there are still any specific steps we could do in the context of this project, please share here. Until then I'm closing it.

m3nu:
Also: RAM usage on macOS is just 129 MB with the main window closed. Not great, but not 500 MB either.

![Screen Shot 2020-05-28 at 12 02 58](https://user-images.githubusercontent.com/3916435/83097660-3ffc4200-a0db-11ea-8ee9-84f2bdb7efe6.jpg)

justauserid:
> Also: RAM usage on macOS is just 129 MB with the main window closed. Not great, but not 500 MB either.

<img width="723" height="136" alt="Image" src="https://github.com/user-attachments/assets/86f68d9f-3b22-463a-a41a-3edee240abb9" />

Looks like it has increased now. For main process the real memory us showing ~110MB as shown by you but total memory usage does contribute in causing memory pressure esp. in the age of local LLMs :-)

ThomasWaldmann:
Yay! #2551

m3nu:
Great memory gains, @ThomasWaldmann . Just did a build and it's down to 27 MB of compressed VM 🙌

<img width="702" height="113" alt="Image" src="https://github.com/user-attachments/assets/12743a77-f188-46ea-a128-5dc083473daa" />

PR #2551, opened 2026-09-05
Load Sparkle without PyObjC's class scan (~150 MB less memory on macOS)

ThomasWaldmann:
### Description

`get_updater()` loads the Sparkle framework with `objc.loadBundle('Sparkle', globals(), bundle_path)`. PyObjC's default `scan_classes=True` does not just register Sparkle: it calls `objc_getClassList()`, realizes **every** Objective-C class in the process (~67,000 on macOS 15 once Qt's Cocoa plugin and the system frameworks are loaded) and wraps each one as a Python class that is kept in this module's globals for the lifetime of the app. That costs ~150 MB of physical footprint and only happens in the frozen app bundle (the updater is only set up when `sys.frozen`), which is why the shipped `Vorta.app` idles at ~220–250 MB while a venv install of the same code idles at ~80–120 MB.

This PR loads only the framework (`scan_classes=False`) and looks up the one class Vorta uses explicitly with `objc.lookUpClass('SUUpdater')`. Nothing else changes: `SUUpdater` was the only bare name taken from the scan (hence the old `# noqa: F821`), other Sparkle classes remain reachable through `objc.lookUpClass`, and objects returned from Sparkle are wrapped lazily by PyObjC either way.

### Related Issue

- https://github.com/borgbase/vorta/issues/337 (Vorta using several hundred MB in the tray)
- https://github.com/borgbackup/borg/discussions/10328 (asks for a lighter Borg GUI; Vorta's ~300–400 MB on macOS is the concern)

### Motivation and Context

Vorta is a tray-resident app, so its idle footprint is what users see all day. Measured on the shipped 0.11.5 bundle right after launch (scratch `$HOME`, no backup run yet): 218 MB with the main window hidden, 250 MB with it shown; `heap` reported 66,870 realized ObjC classes and 71,919 Python objects (94 MB) allocated by the class scan. A pip install of the same version on the same machine: 91 MB / 120 MB. A 150-cycle `borg create` run showed no per-backup growth, so the class scan is the dominant cost.

### How Has This Been Tested?

- New unit test `tests/unit/test_updater.py` (macOS only): asserts `objc.loadBundle` is called with `scan_classes=False` and `SUUpdater` is obtained via `objc.lookUpClass`.
- `pytest tests/unit`: 265 passed, 7 skipped. `ruff check` / `ruff format --check` clean.
- Two app bundles built with PyInstaller from this tree (macOS 15.7.9, Apple M3, Python 3.11, PyQt6 6.6.1), launched under a scratch `$HOME` with the main window hidden and measured 35 s after launch:

  | | physical footprint | realized ObjC classes |
  |---|---|---|
  | `master` (ebf7571) | 228.8 MB | 66,928 |
  | this branch | **78.6 MB** | 2,322 |

- Isolated check with PyObjC 12 and the real `Sparkle.framework`: the default scan costs +62 MB and adds 15,559 names to the module, of which 57 come from Sparkle; `scan_classes=False` + `lookUpClass` costs +1 MB and returns the same class.

Side note for maintainers: building from `master` needed the spec's `datas` to collect `assets/UI` recursively (the dialogs' `.ui` files moved into subdirectories), otherwise the bundle fails at startup on `assets/UI/dialogs/exception.ui`. That is unrelated to this change and not included here.

### Types of changes

### Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/vorta/updater.py
+++ b/src/vorta/updater.py
@@ -25,8 +25,11 @@ def get_updater():
             'Frameworks',
             'Sparkle.framework',
         )
-        objc.loadBundle('Sparkle', globals(), bundle_path)
-        sparkle = SUUpdater.sharedUpdater()  # noqa: F821
+        # Only load the framework. The default class scan (`scan_classes=True`) wraps every
+        # Objective-C class in the process (tens of thousands) as a Python class and keeps
+        # them in this module's globals for the lifetime of the app, costing well over 100 MB.
+        objc.loadBundle('Sparkle', globals(), bundle_path, scan_classes=False)
+        sparkle = objc.lookUpClass('SUUpdater').sharedUpdater()
 
         # A default Appcast URL is set in vorta.spec, when setting it here it's saved to defaults,
         # so we need both cases.
--- a/tests/unit/test_updater.py
+++ b/tests/unit/test_updater.py
@@ -1,0 +1,30 @@
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+
+from vorta.updater import get_updater
+
+
+@pytest.mark.skipif(sys.platform != 'darwin', reason='The Sparkle updater is macOS only.')
+def test_get_updater_loads_sparkle_without_class_scan(qapp, monkeypatch):
+    """
+    Sparkle must be loaded without PyObjC's class scan.
+
+    The scan wraps every Objective-C class in the process as a Python class and keeps them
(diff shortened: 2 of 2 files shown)


ThomasWaldmann:
@m3nu Have a look!

Looks like Claude Fable 5.1 found a nice tweak to reduce memory usage significantly.

Also, it found some other issues while doing that, see the other PRs.

